PR #5, opened 2016-06-22
Create super packs to save disk space

matthewrsj:
- superpack-maker.sh

  Create superpacks for each bundle in an update version. The
  superpack replaces oldest pack in that directory (excluding
  zero packs) and is a superset of updates needed for all newer
  packs. Leaves a defined number of newest delta packs unchanged.
  The remaining delta packs are symlinked to the superpack.

- create-update.sh

  Define variables for superpack-maker.sh, call pack-maker.sh, and
  call superpack-maker.sh

- Makefile.am

  Install superpack-maker.sh

### DIFF
--- a/create-update.sh
+++ b/create-update.sh
@@ -5,9 +5,6 @@ PREFIX=
 LOG_DIR="$PWD/logs"
 NOPUBLISH=0
 ZEROPACKS=1
-BACK_COUNT=30       # Number of packs to create backward from current version
-DELTA_PACK_COUNT=5  # Number of those packs to keep as delta packs, the rest
-                    # will become symlinks to a superpack
 
 # Strip the trailing and leading whitespace on variables to sanitize them
 function strip_whitespace {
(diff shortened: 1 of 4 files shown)


create-update.sh
```diff
@@ -130,6 +133,10 @@ if [ $NOPUBLISH -eq 0 ]; then
 	sudo cp "$PWD/.mix-version" "$STATE_DIR/www/version/format$FORMAT/latest"
 fi
 
-# step 6: archive the swupd-server logs for this mix build
+# step 6: create delta packs and super packs
```
tmarcu: These two scripts (step 6 overall) should be left out of create-update.sh and run as a separate part of update, since delta packs are not required to create a fully working update.

create-update.sh
```diff
@@ -130,6 +133,10 @@ if [ $NOPUBLISH -eq 0 ]; then
...
+sudo mixer-pack-maker.sh $MIXVER $BACK_COUNT
```
tmarcu: This script should accept a $STATE_DIR arg as well, since it may be run from any location w/arbitrary state_dir.

tmarcu:
PR Merged with minor cleanups.

PR #6, opened 2016-08-30
pack maker rewrite: support zero packs, change delta pack creation logic

phmccarty:
The first aim of this patch is to centralize pack creation for SWUPD,
adding the ability to create zero packs in addition to delta packs.

Secondly, to make delta pack creation more flexible, instead of
considering the last N times a bundle changed, require the user to
specify the "from" version delta packs should be based on. More than one
--from option can be passed if several sets of delta packs need to be
built.

To facilitate these different modes of operation, I added option
parsing, also with --help output and some examples documented in the
script header.

Signed-off-by: Patrick McCarty <patrick.mccarty@intel.com>

### DIFF
--- a/pack-maker.sh
+++ b/pack-maker.sh
@@ -1,110 +1,206 @@
-#!/bin/bash -e
-# usage:
-#	mixer-pack-maker.sh <to version> <back_count>
-#	mixer-pack-maker.sh 2820 2
-#	mixer-pack-maker.sh 2730 3
+#!/bin/bash
 #
-# For each bundle changed as of the Manifest.MoM at the <to version>,
-# create version-pair packs going back <back_count>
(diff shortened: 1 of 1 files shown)


phmccarty:
I'm thinking of reworking the parallelization of pack creation to use the `parallel` command instead, since it is much more flexible than using noninteractive shell job control.

So, most of the core changes I made here remain unmodified, just with some slight rework to adapt to using `parallel`.

phmccarty:
Uploaded second patch, implementing the parallelization with GNU Parallel instead.

phmccarty:
Rebased to latest master.

tmarcu:
Merged into master 4da94d56f12ecdbe5ef78d59b06af3161fbdb2ce.

PR #9, opened 2016-09-20
Simplify the hardlink command invocation

phmccarty:
Because the intent is to create hardlinks where possible within the
image/MIXVER directory tree, only the toplevel directory name needs to
be given. This avoids arbitrarily long command lines when the glob is
expanded by the shell.

Signed-off-by: Patrick McCarty <patrick.mccarty@intel.com>


tmarcu:
Merged into master 200af6aa8dbb5e346c5671e90fe3cb551bc85572

PR #10, opened 2016-10-06
build-chroot: honor system's proxy

dorileo:
We should follow the system's proxy configuration instead of working
around it.

Signed-off-by: Leandro Dorileo <leandro.maciel.dorileo@intel.com>


phmccarty:
Looks good to me. +1

tmarcu:
+1, Merged as part of the mixer-tools rewrite.

PR #13, opened 2016-10-19
Script rewrites

tmarcu:


Makefile.am
```diff
@@ -3,11 +3,13 @@ dist_bin_SCRIPTS = \
 	create-update.sh \
 	build-chroots.sh \
 	init-mix.sh \
-	init-versions.sh \
 	update-bundles.sh \
 	pack-maker.sh \
 	superpack-maker.sh
 
+instdir = ${datadir}/mixer-tools
```
phmccarty: should be `$(datadir)` instead of `${datadir}`

add-rpms.sh
```diff
@@ -1,77 +1,73 @@
 #!/bin/bash
+source /usr/share/mixer-tools/helpers
```
phmccarty: Note that you are installing this file to `$(datadir)/helpers`, so if mixer-tools is built with a datadir different than `/usr/share`, the file sourcing here will fail.

One solution would be to make the scripts templates instead, and substitute the correct values at configure time. The line in the template would look like `source @datadir@/mixer-tools/helpers`.

add-rpms.sh
```diff
@@ -1,77 +1,73 @@
...
-  ret=$(file "$rpm")
-  echo "$ret" | grep --quiet ": RPM"
-  if [[ $? -ne 0 ]] ; then
-    echo "ERROR $rpm IS NOT VALID"
-  else
-    echo "Copying $rpm"
-    cp "$rpm" "$REPODIR"
-  fi
-done
+ret=$(file "$rpm")
```
phmccarty: indentation was removed for this line...

create-update.sh
```diff
@@ -89,58 +73,35 @@ fi
...
 
 # step 4: hardlink relevant dirs
 sudo -E "hardlink" -f "$STATE_DIR/image/$MIXVER"/
 
 # step 5: update latest version
 if [ $NOPUBLISH -eq 0 ]; then
-	sudo cp "$PWD/.mix-version" "$STATE_DIR/image/LAST_VER"
-	sudo cp "$PWD/.mix-version" "$STATE_DIR/www/version/format$FORMAT/latest"
+	sudo -E echo "$MIXVER" > "$STATE_DIR/image/LAST_VER"
+	sudo -E echo "$MIXVER" > "$STATE_DIR/www/version/format$FORMAT/latest"
```
phmccarty: `echo` might be a shell builtin, in which case I would expect these two commands to fail. I would suggest using `sh -c ...` instead.

init-mix.sh
```diff
@@ -1,13 +1,12 @@
-#!/bin/bash
-
+#!/bin/bash -x
```
phmccarty: shell tracing on by default?

update-bundles.sh
```diff
@@ -1,48 +1,48 @@
-#!/bin/bash
+#!/bin/bash -x
```
phmccarty: shell tracing on by default?

helpers
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
```
tmarcu: Yes ty for pointing that out... should have been clear in my editor, I'll fix.

init-mix.sh
tmarcu: Debug artifact. removed.

update-bundles.sh
tmarcu: Same as previous, removed.

Makefile.am
phmccarty: it's for style consistency

Makefile.am
```diff
@@ -3,11 +3,13 @@ dist_bin_SCRIPTS = \
 	create-update.sh \
 	build-chroots.sh \
 	init-mix.sh \
-	init-versions.sh \
```
tmarcu: No longer needed. The builder.conf hosts the values, and I will adjust the scripts to auto-increment or set the next mix value when starting to build the version. Imho we shouldn't need a whole file to edit a version #.

phmccarty: I'm okay with that approach. The removal was something that glaringly stood out when reviewing this patch, as there was no mention of it in the commit message.

tmarcu: Good point, I will amend the commit message, it should be stated.

PR #18, opened 2016-12-13
Improve clr-bundles branch updates

phmccarty:
With the current workflow, the "mix-bundles" repo is used for
mix-specific bundle changes, and the "clr-bundles" repo is treated as
read-only.

So in clr-bundles, the CLRVER_mix branches should always point to the
commit tagged with CLRVER. To make this handling more robust, first
verify that CLRVER exists and is a tag. Then create the CLRVER_mix
branch based on CLRVER, or if the branch already exists, reset it
appropriately.

Since this is the best place to detect invalid CLEAR_VERSION numbers
passed in the builder.conf, I've added that error handling too.

### DIFF
--- a/helpers
+++ b/helpers
@@ -54,15 +54,14 @@ update_repo() {
 		git pull origin master
 	fi
 	set +e
-	# checkout the tag relating to the clear version used to build against
-	git checkout tags/"$CLRVER"
-	local branch="${CLRVER}_mix"
-	git rev-parse --verify "$branch"
+	# Ensure that CLRVER exists and is a tag
+	git rev-parse --verify "${CLRVER}^{tag}" &> /dev/null
 	if [ $? -eq 0 ]; then
-		git checkout "$branch"
-		git pull
-		else
-		git checkout -b "$branch"
+		# Always reset the _mix branch to be based off the tag
+		git checkout -B "${CLRVER}_mix" "${CLRVER}"
+	else
+		echo "ERROR: invalid CLEAR_VERSION in builder.conf (${CLRVER})"
+		exit 1
 	fi
 	set -e
 	cd ..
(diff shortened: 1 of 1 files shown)


helpers
```diff
@@ -55,10 +55,10 @@ update_repo() {
 	fi
 	set +e
 	# Ensure that CLRVER exists and is a tag
-	git rev-parse --verify "${CLRVER}^{tag}" &> /dev/null
+	git rev-parse --verify "refs/tags/${CLRVER}^{tag}" &> /dev/null
 	if [ $? -eq 0 ]; then
 		# Always reset the _mix branch to be based off the tag
-		git checkout -B "${CLRVER}_mix" "${CLRVER}"
+		git checkout -B "${CLRVER}_mix" "refs/tags/${CLRVER}"
```
bryteise: Might want to clean after this too in order to avoid new bundle files not in git. Also this can fail if the branch already exists, should the return from the checkout command be tested?

helpers
```diff
@@ -54,15 +54,14 @@ update_repo() {
...
-	git rev-parse --verify "$branch"
+	# Ensure that CLRVER exists and is a tag
+	git rev-parse --verify "${CLRVER}^{tag}" &> /dev/null
 	if [ $? -eq 0 ]; then
-		git checkout "$branch"
-		git pull
-		else
-		git checkout -b "$branch"
+		# Always reset the _mix branch to be based off the tag
+		git checkout -B "${CLRVER}_mix" "${CLRVER}"
```
phmccarty: The -B option doesn't fail if the branch already exists. I agree that a `git clean` might be helpful.

bryteise:
Just commented as the second change was pushed =). I think my comments still apply though.

helpers
bryteise: Ah indeed, I wasn't checking my output correctly. It was failing on the ambiguous object name in checkout for my testing.

PR #21, opened 2017-05-25
Hardlink rpms when calling add-rpms

tmarcu:
To save time on copying fullfiles, just make a hardlink from the files in
RPMDIR to the REPODIR.

### DIFF
--- a/src/builder/builder.go
+++ b/src/builder/builder.go
@@ -498,8 +498,18 @@ func (b *Builder) AddRPMList(rpms []os.FileInfo) {
 			fmt.Println("ERROR: RPM is not valid! Please make sure it was built correctly.")
 			os.Exit(1)
 		} else {
-			fmt.Printf("Copying %s\n", rpm.Name())
-			helpers.CopyFile(b.Repodir+"/"+rpm.Name(), b.Rpmdir+"/"+rpm.Name())
+			if _, err = os.Stat(b.Repodir + "/" + rpm.Name()); err == nil {
+				continue
+			}
+		}
+		fmt.Printf("Hardlinking %s to repodir\n", rpm.Name())
+		err := os.Link(b.Rpmdir+"/"+rpm.Name(), b.Repodir+"/"+rpm.Name())
+		if err != nil {
+			err = helpers.CopyFile(b.Repodir+"/"+rpm.Name(), b.Rpmdir+"/"+rpm.Name())
+			if err != nil {
+				helpers.PrintError(err)
+				os.Exit(1)
+			}
 		}
 	}
 	// Save current dir so we can get back to it


src/builder/builder.go
```diff
@@ -498,8 +498,15 @@ func (b *Builder) AddRPMList(rpms []os.FileInfo) {
...
-			fmt.Printf("Copying %s\n", rpm.Name())
-			helpers.CopyFile(b.Repodir+"/"+rpm.Name(), b.Rpmdir+"/"+rpm.Name())
+			if _, err = os.Stat(b.Repodir + "/" + rpm.Name()); err == nil {
+				continue
+			}
+			fmt.Printf("Hardlinking %s to repodir\n", rpm.Name())
+			err := os.Link(b.Rpmdir+"/"+rpm.Name(), b.Repodir+"/"+rpm.Name())
+			if err != nil {
+				helpers.PrintError(err)
+				os.Exit(1)
```
bryteise: Seems harsh to fail and not just fall back to copyfile in case folks have these on different filesystem partitions.

src/builder/builder.go
```diff
@@ -498,8 +498,15 @@ func (b *Builder) AddRPMList(rpms []os.FileInfo) {
 			fmt.Println("ERROR: RPM is not valid! Please make sure it was built correctly.")
 			os.Exit(1)
 		} else {
-			fmt.Printf("Copying %s\n", rpm.Name())
-			helpers.CopyFile(b.Repodir+"/"+rpm.Name(), b.Rpmdir+"/"+rpm.Name())
+			if _, err = os.Stat(b.Repodir + "/" + rpm.Name()); err == nil {
+				continue
+			}
```
phmccarty: If err != nil, I suppose the code assumes that you hit an ENOENT condition and can safely try to hardlink. Can you recover from other errors that may occur?

src/builder/builder.go
tmarcu: @bryteise fair point.

src/builder/builder.go
tmarcu: What other errors do you think should be covered?

phmccarty: Well, stat(2) can return many different errors. The ones that are most interesting (aside from ENOENT) are related to path traversal failures: EACCES, ELOOP, ENAMETOOLONG, and ENOTDIR. You can probably rule out EACCES, because mixer runs as root. The others are possible, though unlikely.

phmccarty: On second thought, os.Link() will likely fail for any of the additional error conditions I mentioned for os.Stat(), so falling through seems pretty safe in this case.

tmarcu: I think they're very unlikely for mixer, and the error reporting is quite verbose. It would be better to assure the environment and dirs are populated correctly first than handle every error.

PR #23, opened 2017-08-03
Improving how commands are organized

cmarcelo:
Some fixes included.

### DIFF
--- a/src/mixer/main.go
+++ b/src/mixer/main.go
@@ -11,140 +11,188 @@ import (
 	"helpers"
 )
 
-// PrintMainHelp emits useful help text to the console
+type Command struct {
+	Name        string
+	Description string
+	Run         func(args []string)
+}
+
+var commands []*Command
+
+func init() {
+	commands = []*Command{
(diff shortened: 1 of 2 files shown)


cmarcelo:
@phmccarty @tmarcu could you take a look?

src/mixer/main.go
```diff
@@ -11,143 +11,193 @@ import (
...
+	b.ReadBuilderConf()
+	b.InitMix(strconv.Itoa(*clearflag), strconv.Itoa(*mixflag), *allflag)
+}
+
+func cmdHelp(args []string) {
+	PrintMainHelp()
+}
+
+func cmdVersion(args []string) {
+	// Do nothing else since version is always printed in main.
```
tmarcu: Instead of having a blank function, why don't make this actually print something - or just remove it completely since the version info is printed? I'm not against having it print the version string in both places, but I'd like to avoid empty/no-op functions.

cmarcelo: I'll just drop it.

cmarcelo:
Rebased for 3.06. Ping?

PR #24, opened 2017-08-03
Merge git helper functions into one

cmarcelo:
And also dump the Stderr of the command to os.Stderr. Without a git
failure might not give enough information about what gone wrong,
e.g. when trying to call "git commit" without a user.email configured.

The commit message was changed to include the Clear version from which
the Mix was based off.

### DIFF
--- a/src/helpers/helpers.go
+++ b/src/helpers/helpers.go
@@ -180,39 +180,14 @@ func GetDirContents(dirname string) []os.FileInfo {
 	return files
 }
 
-// GitInit attempts to initialize an empty git repository in the current
-// directory, or exits on failure.
-func GitInit() {
-	gitcmd := exec.Command("git", "init")
-	gitcmd.Stdout = os.Stdout
-	err := gitcmd.Run()
+// Git runs git with arguments and exit in case of failure.
+func Git(args ...string) {
+	cmd := exec.Command("git", args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err := cmd.Run()
 	if err != nil {
-		PrintError(err)
-		fmt.Println("Failed to init git repo, exiting...")
-		os.Exit(1)
-	}
-}
-
-// GitAdd performs a 'git add .' in the current directory
-func GitAdd() {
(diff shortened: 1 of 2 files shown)


cmarcelo:
@tmarcu @phmccarty 

The example mentioned in the commit message happened when I was following the HOWTO from the site documentation in a new environment. 

By outputting the error, it is easier to see what is going on. One "easy" way is to add those instructions to the HOWTO, but will add just noise to the instructions. It seems to me that for a better out-of-the-box experience we should either:

(a) the tool always specify some predefined username / email for the commit (since it is somehow automated);
(b) don't try to create a git repository on behalf of the user, just set up the mix-bundles directory and let the user add versioning if they want.

I'm happy to write a patch for either case if it is desired.

cmarcelo:
Ping

tmarcu:
I like having it be a git repo for them as it allows them to keep history manually, and also opens the door for mixer to automatically keep history if we decide to add such features later.
IMHO the best way is either write in the docs: set your username/email as part of the setup of mixer tools, which is easy and they do once (if they have not already done it for previous projects).
Or fallback to doing a default username/email.

src/helpers/helpers.go
```diff
@@ -180,39 +180,14 @@ func GetDirContents(dirname string) []os.FileInfo {
 	return files
 }
 
-// GitInit attempts to initialize an empty git repository in the current
-// directory, or exits on failure.
-func GitInit() {
-	gitcmd := exec.Command("git", "init")
-	gitcmd.Stdout = os.Stdout
-	err := gitcmd.Run()
+// Git runs git with arguments and exit in case of failure.
```
tmarcu: +1 on the code, but can we extend this comment to say args must be validated before being passed to this command? Or that any input from an external source MUST be validated and escaped correctly, so that I can't do something like:
"git status; rm -rf .*" if it's receiving *args* that aren't hardcoded (like they are here).

cmarcelo: Note that issue like one you mentioned, using ";", doesn't happen because exec package doesn't go through the shell to execute the command. It just exec the "git" directly and pass everything else as arguments for it (including the ";").

PR #25, opened 2017-08-10
When HTTP download fails, don't keep the file around

cmarcelo:
Only create the file if the HTTP connection succeeds, and in case of
failure later on during download, remove the partial file.

This let the user retries the operation, otherwise the empty/partial
file will prevent the download. One of such cases is forgetting to set
the https_proxy.


cmarcelo:
Ping because a coworker was bitten by this problem today.

tmarcu:
+1 this is more ideal behavior.

Issue #28, opened 2017-11-21
Allow use of Environment Variable in Configuration File


mdhorn:
It would be very advantageous to allow the use of environment variables, such a `$HOME`, in the builder configuration file.  When used in an automated build environment like Jenkins or Travis, it would be helpful to use environment variables for the various paths.

NOTE: This will require a compatible change to the `bundle-chroot-builder` tool as it also reads the same configuration file.

See https://github.com/clearlinux/bundle-chroot-builder/issues/11

PR #29, opened 2017-11-21
Enable Environment Variables in Configuration File

mdhorn:
This code change enables the use of Linux Environment
variables in the configuration file. This is very useful
for paths to avoid having to hard-code paths.

For proper use, it will also require a similar patch to
the bundle-chroot-builder project.

Fixes  https://github.com/clearlinux/mixer-tools/issues/28

### DIFF
--- a/src/builder/builder.go
+++ b/src/builder/builder.go
@@ -115,9 +115,21 @@ func (b *Builder) ReadBuilderConf() {
 
 	for _, h := range fields {
 		r := regexp.MustCompile(h.re)
+		// Look for Environment variables in the config file
+		re := regexp.MustCompile(`\$\{?([[:word:]]+)\}?`)
 		for _, i := range lines {
 			if m := r.FindIndex([]byte(i)); m != nil {
-				*h.dest = i[m[1]:]
+				// We want the variable without the $ or {} for lookup checking
+				matches := re.FindAllStringSubmatch(i[m[1]:], -1)
+				for _, s := range matches {
+					if _, ok := os.LookupEnv(s[1]); !ok {
+						helpers.PrintError(fmt.Errorf("buildconf contains an undefined environment variable: %s", s[1]))
+						os.Exit(1)
+					}
+				}
+
+				// Replace valid Environment Variables
+				*h.dest = os.ExpandEnv(i[m[1]:])
 			}
 		}
 	}


tmarcu:
We purposely moved away from using environment variables in the past for swupd and mixer so everything is well-defined and in a specific location...so I'm not sure I want to enable this. The paths are still hardcoded if you use an environment variable, it's just abstracted out to being defined in your shell env instead, and the intent with mixer is to have a dedicated workspace for your builds. 
What use case does this enable? The dir will always reside in the same place (ideally) and if it moves the hardcoded env var will have to be updated anyway - just like the regular path in the conf would have to be.

kevincwells:
@tmarcu -- I can see your argument that this is just changing where you need to update paths (in your environment rather than in the config), but I suspect we're looking at it from different usages. I suspect you're looking at it from the perspective of a given devops flow that wants to ensure system stability. @mdhorn is looking at it from the perspective of someone needing to set up, replicate, or migrate devops flows. I'm looking at it from the perspective of someone wanting to experiment with mixes outside of a devops flow.

For one, this would enable easier portability between build machines. The example given in the issue (https://github.com/clearlinux/mixer-tools/issues/28) is `$HOME`, an environment variable you wouldn't have to manually update. With this, you could set your `SERVER_STATE_DIR` to something like `$HOME/mix/update`, and easily transfer your setup from one machine to another without changing anything. Alternatively, someone wanting to experiment with mixes could set the base of their path to `$PWD` and re-use config setups between experiments.

Additionally, I would argue that setting an environment variable is likely significantly easier than updating the path in the config file in many use cases, especially if you were trying to script the usage of these tools. Exporting an environment variable is trivial in a shell/python/whatever script; editing a field in a file (especially one that isn't in a format with common parser libraries like JSON or YAML), while certainly doable, requires significantly more complicated string parsing. I'm guessing this isn't a use case that has been needed yet, but it seems plausible to me.

I guess at the end of the day I'm not sure I understand the argument _against_ enabling environment variable parsing. It doesn't preclude you from using hardcoded paths to ensure behavior and stability identical to today. Are you just worried that it will encourage/enable bad, potentially error-prone setups?

tmarcu:
@kevincwells I guess I'm still trying to understand the use-case this is trying to enable, and if it's actually valid. It's trivial to script this out with: sed -i "s/STATE_DIR.*/STATE_DIR=$newpath" builder.conf, on any system, which is currently used in some automation scripts. I am indeed trying to limit it so there are less points of failure & bad setups.
Building content on separate machines doesn't make sense because every build is stateful and depends on the previous build before it, so while one machine builds e.g. version 20, no other machine would be able to build at the same time. Then when that finishes, moving to another machine would mean copying the entire ~multi-gig workspace over to the new machine so version 30 could be built and so on. This introduces huge latency in your process which is why building is always done on a single machine. Migrating to a new - dedicated - builder can be scripted out, but I don't see that happening very often in a correct setup except when EOL-ing machines.
Could someone clarify why multiple builders are necessary perhaps? Or the the kinds of mixes you are intending to do?

mdhorn:
@tmarcu In general, it is very standard to have configuration files allow the use of Environment variables regardless of the tool. This was my expectation from the beginning, and hence why I've created the PRs to enable the functionality. I already have the example of when I'm doing development testing using a different location than what is used for our "production" work.
Adding this functionality will not impact your current production use case, but it will resolve the problem we are currently having while doing development testing.

What's the concern about enabling this functionality other than existing work flows do not need it?

src/builder/builder.go
```diff
@@ -116,7 +116,7 @@ func (b *Builder) ReadBuilderConf() {
 		r := regexp.MustCompile(h.re)
 		for _, i := range lines {
 			if m := r.FindIndex([]byte(i)); m != nil {
-				*h.dest = i[m[1]:]
+				*h.dest = os.ExpandEnv(i[m[1]:])
```
tmarcu: IMHO this should be validated in the case that someone uses a $VAR and it's not set, which ExpandEnv will return "" for, and will cause problems later on in the build paths. Can we check this and exit if something sane isn't returned?

mdhorn: @tmarcu ExpandEnv is an API which replaces the environment variable with it's value. Although I already that if a variable which isn't defined is used will result in an empty string, there is not a simple way to error check this. 
The same will be true of the parallel PR to expand environment variables in the builder.conf used by the bundle-chroot-builder.py.

tmarcu: I think it's feasible to check that env variables are actually set, with something (roughly) like:
 re := regexp.MustCompile("\\$\\{?[[:word:]]+\\}?")
    matches := re.FindAllString(i[m[1]:], -1)
    for _, s := range matches {
        if _, ok := os.LookupEnv(s[1:]); !ok {
            //print error & os.Exit()
....
If it's ok then *h.dest = os.ExpandEvn(i[m[1]:]) like you're doing already. I think it'd also be reasonable to say we should expect any env variable to just not be an empty string, since setting one to empty is valid and you can't tell the difference with ExpandEnv in that case. However if you say variables must be defined and not empty then it's easier just to check if ExpandEnv(s)  != "" and fail. Saves a later crash and/or content being generated in system locations they're not meant to go in.

kevincwells: +1 screening with os.LookupEnv()

I think it's plausible someone might deliberately use an empty string (If it's the start of a path, for example, and not the entire value), though disallowing that seems fair enough. But at least checking that it's set seems both safer and relatively straightforward.

@tmarcu - Interesting, I didn't realize golang's regexp syntax accepted both \w and [[:word:]].

kevincwells: @mdhorn  FYI -- you can do something similar in Python for your other PR with `if "FOO" in os.environ:`

mdhorn:
@tmarcu Added the environment variable error checking

Issue #30, opened 2017-11-22
Store the ClearLinux distribution and version used for Mix


mdhorn:
Have Mixer save the upstream URL and version used when generating the new Mix.

This information is both useful historical information about the Mix as well as needed by any automation used to generate new mixes.

The primary use would be to identify the upstream's original Swupd Format when generating a new Mix. The Format of the upstream, and how many times it has bumped, are needed in order to generate the necessary mixes for upgrading the downstream client of the Mix.

Assuming a new Mix stream is created based on the Public ClearLinux distribution plus some additional bundles (i.e. for a production product), the following steps would be required.

A new production Mix, version 10 Format 1, is made based on the [ClearLinux 18030 which is Format 19 ](https://download.clearlinux.org/update/18030/). When the next Mix is needed, the public [ClearLinux is now at version 19110 which is Format 21](https://download.clearlinux.org/update/19110/); the Upstream Format has bumped 2 times.

One suggested solution is [saving the URL used for the Upstream and the version into files in the Mix when it is created](https://github.com/mdhorn/mixer-tools/commit/d102d77724f5fb958b533774824e41b2477c63b5).
https://github.com/mdhorn/mixer-tools/commit/d102d77724f5fb958b533774824e41b2477c63b5

With that information, the Format of the Upstream can be found:
https://download.clearlinux.org/update/18030/format = 19
Next, the latest release of Format 19 can be determined:
https://download.clearlinux.org/update/version/format19/latest = 18500

So the first new Mix (20; still at Mix Format 1) will need to be based on Upstream's 18500.
The next Mix (30; Mix Format 2) would need to be based on the First release of Format 20:
https://download.clearlinux.org/update/version/format20/first = 18510
And then another Mix (40; Mix Format 2) based on the Last release of Format 20:
https://download.clearlinux.org/update/version/format20/latest = 18930
Next Mix (50; Mix Format 3) based on the First release of Format 21:
https://download.clearlinux.org/update/version/format21/first = 18940

Finally, we can make our last Mix (60; Mix Format 3) based on the choose 19110.

The summary would be:

Upstream Version | Upstream Format | Mix Version | Mix Format
------------ | ------------- | ------------- | -------------
18030 | 19 | 10 | 1
18500 | 19 | 20 | 1
18510 | 20 | 30 | 2
18930 | 20 | 40 | 2
18940 | 21 | 50 | 3
19110 | 21 | 60 | 3

[...]

PR #31, opened 2017-11-22
Save Mixer Upstream URL and Version

mdhorn:
In order to known if there have been format bumps of the
version of ClearLinux this mix is based upon (upstream),
both the source URL and version will be needed.

See https://github.com/clearlinux/mixer-tools/issues/30

### DIFF
--- a/src/builder/builder.go
+++ b/src/builder/builder.go
@@ -34,6 +34,7 @@ type Builder struct {
 	Versiondir  string
 	Yumconf     string
 	Yumtemplate string
+	Upstreamurl string
 
 	Signing int
 	Bump    int
@@ -141,6 +142,14 @@ func (b *Builder) ReadVersions() {
 	}
 	b.Clearver = strings.TrimSpace(string(ver))
 	b.Clearver = strings.Replace(b.Clearver, "\n", "", -1)
+
+	ver, err = ioutil.ReadFile(b.Versiondir + "/.clearurl")
+	if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+	b.Upstreamurl = strings.TrimSpace(string(ver))
+	b.Upstreamurl = strings.Replace(b.Upstreamurl, "\n", "", -1)
 }
 
 // SignManifestMOM will sign the Manifest.Mom file in in place based on the Mix
@@ -302,24 +311,30 @@ func (b *Builder) AddBundles(bundles []string, force bool, git bool) int {
 
 // InitMix will initialise a new swupd-client consumable "mix" with the given
 // based Clear Linux version and specified mix version.
-func (b *Builder) InitMix(clearver string, mixver string, all bool) error {
+func (b *Builder) InitMix(clearver string, mixver string, all bool, upstreamurl string) error {
 	if clearver == "0" || mixver == "0" {
 		fmt.Println("ERROR: Please supply -clearver and -mixver")
 		os.Exit(1)
 	}
-	err := ioutil.WriteFile(b.Versiondir+"/.clearversion", []byte(clearver), 0644)
+	err := ioutil.WriteFile(b.Versiondir+"/.clearurl", []byte(upstreamurl), 0644)
 	if err != nil {
 		helpers.PrintError(err)
 		os.Exit(1)
 	}
-	b.Mixver = mixver
+	b.Upstreamurl = upstreamurl
+	err = ioutil.WriteFile(b.Versiondir+"/.clearversion", []byte(clearver), 0644)
+	if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+	b.Clearver = clearver
 
 	err = ioutil.WriteFile(b.Versiondir+"/.mixversion", []byte(mixver), 0644)
 	if err != nil {
 		helpers.PrintError(err)
 		os.Exit(1)
 	}
-	b.Clearver = clearver
+	b.Mixver = mixver
 
 	b.UpdateRepo(clearver, all)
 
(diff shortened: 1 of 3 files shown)


tmarcu:
This looks good, one question though is the URL expected to be:
https://download.clearlinux.org/update/18030
or
https://download.clearlinux.org/update/

In this case?

mdhorn:
@tmarcu The intent was to have the URL in one file and the version in another file as the URL could be used for multiple related queries about about the original upstream.

tmarcu:
@mdhorn  ok so it would be the 1-level up URL then with no version info associated? Although it would arguably easier to read a file off your system than query a URL either way, but I was just curious :)

mdhorn:
@tmarcu I'm still working off the assumption that you can determine the information based on the previously published versions instead of relying upon having the previous mix results still in around in a known state. As a developer I might not have kept things around or in a known, working state.

src/builder/builder.go
```diff
@@ -110,6 +111,7 @@ func (b *Builder) ReadBuilderConf() {
 		{`^SERVER_STATE_DIR\s*=\s*`, &b.Statedir},
 		{`^VERSIONS_PATH\s*=\s*`, &b.Versiondir},
 		{`^YUM_CONF\s*=\s*`, &b.Yumconf},
+		{`^CONTENTURL\s*=\s*`, &b.Contenturl},
```
tmarcu: This should be renamed to UPSTREAMURL or something like that, since CONTENTURL already exists in the builder.conf and refers to the location clients running the mixer should update from.

mdhorn: @tmarcu Done. New code pushed for review.

mdhorn:
Hold on this -- that last change breaks the code.
CONTENTURL was the correct entry because I'm looking for that value from the builder.conf to know what the upstream is.

mdhorn:
@tmarcu Correct code pushed.

src/builder/builder.go
```diff
@@ -111,6 +112,7 @@ func (b *Builder) ReadBuilderConf() {
 		{`^SERVER_STATE_DIR\s*=\s*`, &b.Statedir},
 		{`^VERSIONS_PATH\s*=\s*`, &b.Versiondir},
 		{`^YUM_CONF\s*=\s*`, &b.Yumconf},
+		{`^CONTENTURL\s*=\s*`, &b.Upstreamurl},
```
tmarcu: This variable in the builder.conf does not refer to the upstream (official) URL. It is in the form of **my-build-server.mydomain/update**, which gets injected into the image (by bundle-chroot-builder) so clients automatically look at this URL when doing swupd operations. So this would just save the current mixer _update/www_ dir into the upstream_url vairable each time this ran. Currently the mixer does not do anything with this option, only the chroot builder, which is why it's not being parsed in here. 

I think the right solution here is to just copy the current .clearversion file to -> www/MIXVER/upstream_url at the end/inside of BuildUpdate(), that way it's saved before it's incremented for the next build.

src/builder/builder.go
```diff
@@ -141,6 +142,14 @@ func (b *Builder) ReadVersions() {
...
 	b.Clearver = strings.TrimSpace(string(ver))
 	b.Clearver = strings.Replace(b.Clearver, "\n", "", -1)
+
+	ver, err = ioutil.ReadFile(b.Versiondir + "/.clearurl")
+	if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+	b.Upstreamurl = strings.TrimSpace(string(ver))
+	b.Upstreamurl = strings.Replace(b.Upstreamurl, "\n", "", -1)
```
kevincwells: I understand this is copying the existing convention used above.  @tmarcu, do you know what edge case this is trying to account for?  The [documentation for strings.TrimSpace](https://golang.org/pkg/strings/#TrimSpace) says that it already includes newlines in its definition of whitespace, so isn't this line redundant?

rchiossi: The edge case is where the new line is in the middle of the string since TrimSpace only remove leading and trailing characters.

tmarcu: This ^

kevincwells: To be honest, I find this surprising. You're expecting .clearversion to potentially contain, for example, "18\n910", but still want to treat that as valid input? Why not skip the strings.TrimSpace and just do a full regexp s/\s// on the entire string? As is, it allows "18 \t   \t   910".  Or, simpler, since clear version is always an integer, why not just do `if _, err := strconv.Atoi(ver)` and whitelist only valid input, rather than parse out invalid input?

rchiossi: Agreed. We shouldn't require so much tinkering to parse something so simple. It seems like a better approach to just invalidate anything that is not the plain number.  If some other piece of software is generating this in a weird format, it is probably better to just patch that software instead to output in a proper format.

mdhorn: This particular data is an URL, not a number.

mdhorn:
@kevincwells @rchiossi I don't disagree with the code clean-up, but it should be done universally as a separate PR IMHO

src/builder/builder.go
kevincwells: Haha, yes, `strconv.Atoi(ver)` wouldn't work for you like it would above. Sanitizing/whitelisting will be different for different expected input.

In your particular case, you certainly *could* try to sanitize the input against some kind of "regex for valid URLs", but I would probably err on the side of basic cleanup like `strings.TrimSpace`. It'll error out below if you try to use an invalid URL.

kevincwells:
+1 separation of concerns.  @rchiossi said he's doing some general cleanup right now, and I've asked him to separate it out, _at minimum_ as separate commit from new features. I, too, would lean to a separate PR altogether.

matthewrsj:
Whatever cleanup is done globally it should be coordinated with @tmarcu's effort to run go linting and testing under CI such as travis.

kevincwells:
Nice. Definitely support automated linting/testing.

mdhorn:
@tmarcu Are there still concerns regarding this pull request that I should be addressing?

PR #35, opened 2017-12-01
Add new args for build-image

mdhorn:
build-image wants a builder.conf file, but the -config option was
missing from this command.

build-image assumed the template file was a hard coded name; added
functionality to pass through with new -template argument.


mdhorn:
For content, I'm trying to run mixer build-images from a directory which does not contain the builder.conf nor the release-image-config.json

PR #36, opened 2017-12-02
Add -all flag to mixer add-bundles

kevincwells:
Adds support for an '-all' flag to mixer add-bundles.
This flag takes precedence over the list passed with
the '-bundles' flag, and still supports '-force' and
'-git' options.

Note: this commit also passes the '-q' ('--quiet')
flag to git when making a commit, supressing the
potentially long output.

### DIFF
--- a/src/builder/builder.go
+++ b/src/builder/builder.go
@@ -232,8 +232,9 @@ func (b *Builder) UpdateRepo(ver string, allbundles bool) {
 // Linux version to the mix-bundles directory
 // bundles: array slice of bundle names
 // force: override bundle in mix-dir when present
+// all: include all CLR bundles. Overrides bundles.
 // git: automatically git commit with bundles added
-func (b *Builder) AddBundles(bundles []string, force bool, git bool) int {
+func (b *Builder) AddBundles(bundles []string, force bool, allbundles bool, git bool) int {
 	var bundleAddCount int
 
 	bundledir := b.Bundledir
@@ -254,6 +255,24 @@ func (b *Builder) AddBundles(bundles []string, force bool, git bool) int {
 		b.UpdateRepo(b.Clearver, false)
 	}
 
+	// Add all bundles if -all passed
+	if allbundles {
+		files, err := ioutil.ReadDir(clrbundledir)
+		if err != nil {
+			helpers.PrintError(err)
+			os.Exit(1)
+		}
+
+		// Clear out bundles if not empty
+		if len(bundles) > 0 {
+			bundles = make([]string, len(files))
+		}
+
+		for _, file := range files {
+			bundles = append(bundles, file.Name())
+		}
+	}
+
 	var includes []string
 	for _, bundle := range bundles {
 		// Check if bundle exists in clrbundledir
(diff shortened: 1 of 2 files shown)


kevincwells:
GitHub's diff of builder.go is a little hard to parse (it got confused about what is a deletion versus just a move), so it might be more helpful to just view the raw source.

Bottom line is: if they want all, it copies them all, else if they want just some bundles, it does what it did before. Still honors -force and -git flags either way.

src/builder/builder.go
```diff
@@ -254,33 +254,53 @@ func (b *Builder) AddBundles(bundles []string, force bool, git bool) int {
...
+			helpers.PrintError(err)
 			os.Exit(1)
 		}
-		// Check if bundle exists in mix bundles dir
-		if _, err := os.Stat(bundledir + bundle); os.IsNotExist(err) || force {
-			// Parse bundle to get all includes
-			if ib, err := helpers.GetIncludedBundles(clrbundledir + bundle); err != nil {
-				helpers.PrintError(errors.New("Cannot parse bundle " + bundle + " from CLR version " + b.Clearver))
+		bundles = make([]string, len(files))
+		for _, file := range files {
```
rchiossi: Instead of duplicating the work, you could append everything to the bundles list here and check if the allBundles flag is set before checking for includes. This keeps the processing of bundles in a single block and remove the need to nest the entire code for the default case. It will make the code easier to maintain.

src/builder/builder.go
```diff
@@ -254,33 +254,53 @@ func (b *Builder) AddBundles(bundles []string, force bool, git bool) int {
...
+			// Check if bundle exists in mix bundles dir
+			if _, err := os.Stat(bundledir + file.Name()); os.IsNotExist(err) || force {
+				helpers.CopyFile(bundledir+file.Name(), clrbundledir+file.Name())
```
rchiossi: there is no handling of the error code returned by CopyFile. Same thing on the block below.

src/builder/builder.go
kevincwells: I thought about that, and @mdhorn had the same question.  I did it this way to avoid muddying the already somewhat complicated logic of recursion and file inspection below.  I would still need an `if allbundles` outside of the main loop that copies over all of the file.Name() elements into the bundles slice, and I would need control flow logic to prevent it from parsing the bundles and recursing.

I honestly thought this way would be easier to understand in the future, but since you both commented on it, I'll reconsider the flow.

src/builder/builder.go
kevincwells: This is an excellent catch. I followed the example in the UpdateRepos function, where it also doesn't check the error. I think I will fix all of these locations in a separate commit on this pull request.

rchiossi: Yeah, I also noticed it in other places. I'm adding the error handling for all users of CopyFile as part of my init patch that I'll post later today. I'm also changing it a bit to transfer the error handler to the caller.

src/builder/builder.go
rchiossi: Indeed, makes sense. But in practice it took me much longer to understand the code since the approach I mentioned was my expectation. I had a "wait a minute, whats going on here" moment when reading the code. It is just my impression though. If you believe your approach is better I'm also ok with it.

src/builder/builder.go
kevincwells: +1 for moving the error printing out of the helper libraries and into the calling functions -- that's the whole point of returning an error to the caller. I know @cmarcelo will cheer too.

If you can, please separate general code cleanup out of your feature enhancement commits. Separation of concerns. I'd be happy with them being separate commits in the same PR, but @tmarcu might want a cleaner divide.

src/builder/builder.go
kevincwells: I've given it a go, and the alternative was less intrusive than I had originally anticipated -- I basically get the recursion avoidance for free, because it's already wrapped in a `if len(includes) > 0`.

It was originally my gut instinct to write it that way, and you and @mdhorn both expected it to be that way, so I definitely think it's the right move.

I'll update the PR in a bit when I'm done testing my change.

kevincwells:
Updated PR to reflect suggestions from @rchiossi and @mdhorn. Definitely cleaner this way.

rchiossi:
+1

PR #37, opened 2017-12-04
Initial work for swupd server integration into mixer

matthewrsj:
This PR adds some files and manifest functions and tests for the added functions.

It has been pointed out to me that table-based tests are usually done in a different way (https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) so I will address that before merging.

### DIFF
--- a/src/swupd/files.go
+++ b/src/swupd/files.go
@@ -1,0 +1,143 @@
+package swupd
+
+import (
+	"fmt"
+	"os"
+)
+
+type ftype int
+type fmodifier int
+type fstatus int
+
+const (
+	TYPE_UNSET ftype = iota
+	FILE
+	DIRECTORY
+	LINK
+	MANIFEST
+)
+
+const (
+	MODIFIER_UNSET fmodifier = iota
+	CONFIG
+	STATE
+	BOOT
+)
+
+const (
+	STATUS_UNSET fstatus = iota
+	DELETED
+	GHOSTED
+)
+
+// File represents an entry in a manifest
+type File struct {
+	Name    string
+	Hash    string
+	Version uint32
+
+	// flags
+	Type     ftype
+	Status   fstatus
+	Modifier fmodifier
+	Rename   bool
+
+	// renames
+	RenameScore uint16
+	RenamePeer  *File
+
+	Info      os.FileInfo
+	DeltaPeer *File
+}
+
+// setFileTypeFromFlag set file type based on flag byte
+func setFileTypeFromFlag(flag byte, f *File) error {
+	switch flag {
+	case 'F':
+		f.Type = FILE
+	case 'D':
+		f.Type = DIRECTORY
+	case 'L':
+		f.Type = LINK
+	case 'M':
+		f.Type = MANIFEST
+	case '.':
+	default:
+		return fmt.Errorf("invalid file type flag: %v", flag)
+	}
+
+	return nil
+}
+
+// setStatusFromFlag set status based on flag byte
+func setStatusFromFlag(flag byte, f *File) error {
+	switch flag {
+	case 'd':
+		f.Status = DELETED
+	case 'g':
+		f.Status = GHOSTED
+	case '.':
+	default:
+		return fmt.Errorf("invalid file status flag: %v", flag)
+	}
+
+	return nil
+}
+
+// setModifierFromFlag set modifier from flag byte
+func setModifierFromFlag(flag byte, f *File) error {
+	switch flag {
+	case 'C':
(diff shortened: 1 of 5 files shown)


src/swupd/files.go
```diff
@@ -0,0 +1,155 @@
...
+		f.Rename = false
+	default:
+		return fmt.Errorf("invalid file rename flag: %v", flag)
+	}
+
+	return nil
+}
+
+// set file flags from string
+func (f *File) SetFlags(flags string) error {
```
tmarcu: Golint will complain about this comment if the function stays exported, should be in the format:
// SetFlags ...<blah blah>...

src/swupd/files.go
```diff
@@ -0,0 +1,155 @@
...
+
+	Info      os.FileInfo
+	DeltaPeer *File
+}
+
+// set file type based on flag
+// F: file
+// D: directory
+// L: link
+// M: manifest
```
tmarcu: This is redundant :) do we need to explain these in the comment?

src/swupd/files.go
matthewrsj: This probably shouldn't be exported anyways, but I'll fix up the comments.

matthewrsj:
Work will be continued on https://github.com/matthwrsj/mixer-tools

src/swupd/files.go
```diff
@@ -0,0 +1,143 @@
...
+	// renames
+	RenameScore uint16
+	RenamePeer  *File
+
+	Info      os.FileInfo
+	DeltaPeer *File
+}
+
+// setFileTypeFromFlag set file type based on flag byte
+func setFileTypeFromFlag(flag byte, f *File) error {
```
cmarcelo: Minor comment: I would just make this and other functions below return a value of the respective numeric type -- so they convert from byte to the enum, and then set it at the caller. Still would return error as a second return value.

src/swupd/files_test.go
```diff
@@ -0,0 +1,178 @@
+package swupd
+
+import "testing"
+
+func TestSetFileTypeFromFlagFile(t *testing.T) {
+	flags := map[byte]ftype{
```
cmarcelo: Minor comment: order in maps is not guaranteed to be the same, so it is common to use an array/slice of a structs instead of a map.

matthewrsj:
thank you @cmarcelo I'll address your comments in the new PR in my fork (https://github.com/matthewrsj/mixer-tools/pull/1)

cmarcelo:
> It has been pointed out to me that table-based tests are usually done in a different way (https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) so I will address that before merging.

OK, but it is worthwhile to keep your use of t.Run(), so that is easy to identify subtests :-)

For the reference, see section "Table-driven tests using subtests" https://blog.golang.org/subtests

src/swupd/files.go
matthewrsj: that would be cleaner, thank you.

src/swupd/manifest.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// readManifestFileHeaderLine Read a header line from a manifest
+func readManifestFileHeaderLine(fields []string, m *Manifest) error {
+	var err error
+	var parsed uint64
+
+	// Only search for defined fields
+	switch fields[0] {
+	case "MANIFEST":
```
cmarcelo: Question: does the manifest format assumes that MANIFEST appears only once and only in the first line? (It seems to me useful to assume that from the format)

cmarcelo: The "Only" once thing might apply to other properties too.

Because of those properties that span multiple line (e.g. headers that must be unique, who needs to appear first), would be useful to later on just inline this function into the actual caller code parsing the header.

src/swupd/manifest.go
```diff
@@ -0,0 +1,118 @@
...
+	case "includes:":
+		m.Header.Includes = append(m.Header.Includes, Manifest{Name: fields[1]})
+	}
+
+	return nil
+}
+
+// readManifestFileEntry
+// fields: "<fflags, 4 chars>", "<hash, 64 chars>", "<version>", "<filename>"
+func readManifestFileEntry(fields [4]string, m *Manifest) error {
```
cmarcelo: Minor comment: fields [4]string will copy the array, that contain 4 references to the strings. fields []string would accept a slice, that would be (slightly) smaller. Call site will have to change to pass a slice too. Usually slices are preferred as function parameters.

PR #38, opened 2017-12-04
helpers: check http.Get status code

cmarcelo:
Also change the caller code to show what error happened: useful to
distinguish 404 from timeouts.

Fix the issue of writing invalid files (without error) if the
.clearversion do not exist. The contents would have the actual error
code.

### DIFF
--- a/src/builder/builder.go
+++ b/src/builder/builder.go
@@ -182,7 +182,7 @@ func (b *Builder) UpdateRepo(ver string, allbundles bool) {
 	URL := "https://github.com/clearlinux/clr-bundles/archive/" + ver + ".tar.gz"
 	err := helpers.Download(repo, URL)
 	if err != nil {
-		fmt.Println("ERROR: Failed to download new clr-bundles, make sure the version is valid")
+		fmt.Fprintf(os.Stderr, "ERROR: Failed to download clr-bundles, make sure the version is valid: %s\n", err)
 		os.Exit(1)
 	}
 
--- a/src/helpers/helpers.go
+++ b/src/helpers/helpers.go
@@ -171,14 +171,16 @@ func CopyFile(dest string, src string) error {
 func Download(filename string, url string) (err error) {
 	infile, err := http.Get(url)
 	if err != nil {
-		PrintError(err)
 		return err
 	}
 	defer infile.Body.Close()
 
+	if infile.StatusCode != http.StatusOK {
+		return fmt.Errorf("Get %s replied: %d (%s)", url, infile.StatusCode, http.StatusText(infile.StatusCode))
+	}
+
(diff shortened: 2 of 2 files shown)


src/builder/builder.go
```diff
@@ -182,7 +182,7 @@ func (b *Builder) UpdateRepo(ver string, allbundles bool) {
 	URL := "https://github.com/clearlinux/clr-bundles/archive/" + ver + ".tar.gz"
 	err := helpers.Download(repo, URL)
 	if err != nil {
-		fmt.Println("ERROR: Failed to download new clr-bundles, make sure the version is valid")
+		fmt.Printf("ERROR: Failed to download new clr-bundles, make sure the version is valid: %s\n", err)
```
tmarcu: Just being picky...can we change this to fmt.Fprintf(os.Stderr, ...etc etc) for the error case? I'd like to follow that convention going forward, and will be updating the rest of the code to be consistent with this.

phmccarty: Also might be nice to have the a wrapper function for printing errors to stderr and in the correct format.

cmarcelo: Updated to use os.Stderr in the output. Letting wrapper function for a different patch.

The wrapper function to print stuff could print+exit, just make sure that it can take printf formatting, for printing richer error messages.

kevincwells: +1 for a unified output library/wrapper/whatnot, and at minimum consistent usage throughout.

In my recent code, I actually used `helpers.PrintError(errors.New("error_message_here"))` just so that my error messages would have the same format as others in the application, but wrapping a string in a new Error object just to use the helper's PrintError logger is indeed a little bit silly.

I was considering a patch that converts all the output to use a standard logging package ([log](https://golang.org/pkg/log/) maybe? [glog](https://godoc.org/github.com/golang/glog) perhaps?). I would be wary of auto-calling exit (or using log.Fatal, for example), simply because that preempts other deferred wrap-up code.

cmarcelo: Just to be clear: I was thinking of the Print/Exit cases already in the code base.

I suggest trying package log, it has both print and print+exit functions.

tmarcu: Yes the error reporting needs to be updated, but I also don't want to just mass log everything without careful consideration over what we're logging and if it'll actually be useful. I am also a fan of the https://godoc.org/github.com/pkg/errors#Wrap type functions, which provide much more useful context and traces.

tmarcu:
+1

src/builder/builder.go
cmarcelo: I agree. That's why I took the print calls from the helper function ;-)

PR #39, opened 2017-12-06
mixer init

rchiossi:
This patchset is separated into 2 dependent parts.

The first two commits are focused on changing the CopyFile helper function so the code can be reused in the following patch. The main change here is the addition of a selective O_EXCL flag when creating the destination file. This allows the copy to fail if the target file already exists. The error handling was also removed from the function and now the caller decides what to do with the error. Proper error handling was added where missing. Note that this changes the behavior of the program. Previously the program would print the error and keep running if CopyFile failed to execute. Now the execution is terminated with an error code.

The second part adds a new command: `mixer init`. The first patch adds the basic functionality which creates and initializes `builder.conf` and the `rpms` and `local` directories. The second patch extends this functionality a bit further and integrates `init-mix` into the `init` command. 

Right now the configuration of `builder.conf` is done by string replacing as there is no proper parsing for the TOML conf file. Ideally the builder object should be able to load and save itself to file format, so there would be no need to copy the template from a hardcoded path nor use regex to load/save `builder.conf` . I plan to work on this next.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -61,6 +61,77 @@ func NewFromConfig(conf string) *Builder {
 	return b
 }
 
+// CreateDefaultConfig creates a default builder.conf using the active
+// directory as base path for the variables values.
+func (b *Builder) CreateDefaultConfig(builderconf string) {
+	pwd, err := os.Getwd()
+	if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+
+	if builderconf == "" {
+		builderconf = pwd + "/builder.conf"
+	}
+
+	err = helpers.CopyFile(builderconf, "/usr/share/defaults/bundle-chroot-builder/builder.conf", false)
+	if os.IsExist(err) {
+		fmt.Printf("File '%s' already exists. Skipping...\n", builderconf)
+		return
+	} else if os.IsNotExist(err) {
+		fmt.Println("Template for 'builder.conf' not found. Please create it manually.")
+		return
+	} else if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+
+	fmt.Println("Creating new builder.conf configuration file...")
+
+	raw, err := ioutil.ReadFile(builderconf)
+	if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+
+	data := strings.Replace(string(raw), "/home/clr/mix", pwd, -1)
+	data += "\n[Mixer]\n"
+	data += "RPMDIR=" + pwd + "/rpms\n"
+	data += "REPODIR=" + pwd + "/local\n"
+
+	err = ioutil.WriteFile(builderconf, []byte(data), 0666)
+	if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+}
+
+// CreateRpmDirs creates the RPM directories
+func (b *Builder) CreateRpmDirs() {
+	pwd, err := os.Getwd()
+	if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+
+	// Make the folder to store rpms
+	err = os.Mkdir(pwd+"/rpms", 0755)
+	if os.IsExist(err) {
+		fmt.Printf("Directory '%s' already exists. Skipping...\n", pwd+"/rpms")
+	} else if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+
+	err = os.Mkdir(pwd+"/local", 0755)
+	if os.IsExist(err) {
+		fmt.Printf("Directory '%s' already exists. Skipping...\n", pwd+"/local")
+	} else if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+}
+
 // LoadBuilderConf will read the builder configuration from the command line if
 // it was provided, otherwise it will fall back to reading the configuration from
 // the local builder.conf file.
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -222,7 +222,7 @@ func (b *Builder) UpdateRepo(ver string, allbundles bool) {
 				os.Exit(1)
 			}
 			for _, file := range files {
-				if err := helpers.CopyFile(bundles+"/"+file.Name(), clrbundles+file.Name()); err != nil {
+				if err := helpers.CopyFile(bundles+"/"+file.Name(), clrbundles+file.Name(), true); err != nil {
 					helpers.PrintError(err)
 					os.Exit(1)
 				}
@@ -230,22 +230,22 @@ func (b *Builder) UpdateRepo(ver string, allbundles bool) {
 		} else {
 			// Install only a minimal set of bundles
 			fmt.Println("Adding os-core, os-core-update, kernel-native, bootloader to mix-bundles...")
-			if err := helpers.CopyFile(bundles+"/os-core", clrbundles+"os-core"); err != nil {
+			if err := helpers.CopyFile(bundles+"/os-core", clrbundles+"os-core", true); err != nil {
 				helpers.PrintError(err)
 				os.Exit(1)
 			}
 
-			if err := helpers.CopyFile(bundles+"/os-core-update", clrbundles+"os-core-update"); err != nil {
+			if err := helpers.CopyFile(bundles+"/os-core-update", clrbundles+"os-core-update", true); err != nil {
 				helpers.PrintError(err)
 				os.Exit(1)
 			}
 
-			if err := helpers.CopyFile(bundles+"/kernel-native", clrbundles+"kernel-native"); err != nil {
+			if err := helpers.CopyFile(bundles+"/kernel-native", clrbundles+"kernel-native", true); err != nil {
 				helpers.PrintError(err)
 				os.Exit(1)
 			}
 
-			if err := helpers.CopyFile(bundles+"/bootloader", clrbundles+"bootloader"); err != nil {
+			if err := helpers.CopyFile(bundles+"/bootloader", clrbundles+"bootloader", true); err != nil {
 				helpers.PrintError(err)
 				os.Exit(1)
 			}
@@ -333,7 +333,7 @@ func (b *Builder) AddBundles(bundles []string, force bool, allbundles bool, git 
 			}
 
 			fmt.Printf("Adding bundle %q\n", bundle)
-			if err = helpers.CopyFile(bundledir+bundle, clrbundledir+bundle); err != nil {
+			if err = helpers.CopyFile(bundledir+bundle, clrbundledir+bundle, true); err != nil {
 				helpers.PrintError(err)
 				os.Exit(1)
 			}
@@ -484,7 +484,7 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
 		}
 		chrootcert := certdir + "/Swupd_Root.pem"
 		fmt.Println("Copying Certificate into chroot...")
-		if err = helpers.CopyFile(chrootcert, b.Cert); err != nil {
+		if err = helpers.CopyFile(chrootcert, b.Cert, true); err != nil {
 			helpers.PrintError(err)
 			return err
 		}
@@ -665,7 +665,7 @@ func (b *Builder) AddRPMList(rpms []os.FileInfo) {
 		fmt.Printf("Hardlinking %s to repodir\n", rpm.Name())
 		err := os.Link(b.Rpmdir+"/"+rpm.Name(), b.Repodir+"/"+rpm.Name())
 		if err != nil {
-			if err = helpers.CopyFile(b.Repodir+"/"+rpm.Name(), b.Rpmdir+"/"+rpm.Name()); err != nil {
+			if err = helpers.CopyFile(b.Repodir+"/"+rpm.Name(), b.Rpmdir+"/"+rpm.Name(), true); err != nil {
 				helpers.PrintError(err)
 				os.Exit(1)
 			}
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -222,15 +222,33 @@ func (b *Builder) UpdateRepo(ver string, allbundles bool) {
 				os.Exit(1)
 			}
 			for _, file := range files {
-				helpers.CopyFile(bundles+"/"+file.Name(), clrbundles+file.Name())
+				if err := helpers.CopyFile(bundles+"/"+file.Name(), clrbundles+file.Name()); err != nil {
+					helpers.PrintError(err)
+					os.Exit(1)
+				}
 			}
 		} else {
 			// Install only a minimal set of bundles
 			fmt.Println("Adding os-core, os-core-update, kernel-native, bootloader to mix-bundles...")
-			helpers.CopyFile(bundles+"/os-core", clrbundles+"os-core")
-			helpers.CopyFile(bundles+"/os-core-update", clrbundles+"os-core-update")
-			helpers.CopyFile(bundles+"/kernel-native", clrbundles+"kernel-native")
-			helpers.CopyFile(bundles+"/bootloader", clrbundles+"bootloader")
+			if err := helpers.CopyFile(bundles+"/os-core", clrbundles+"os-core"); err != nil {
+				helpers.PrintError(err)
+				os.Exit(1)
+			}
+
+			if err := helpers.CopyFile(bundles+"/os-core-update", clrbundles+"os-core-update"); err != nil {
+				helpers.PrintError(err)
+				os.Exit(1)
+			}
+
+			if err := helpers.CopyFile(bundles+"/kernel-native", clrbundles+"kernel-native"); err != nil {
+				helpers.PrintError(err)
+				os.Exit(1)
+			}
+
+			if err := helpers.CopyFile(bundles+"/bootloader", clrbundles+"bootloader"); err != nil {
+				helpers.PrintError(err)
+				os.Exit(1)
+			}
 		}
 
 		// Save current dir so we can get back to it
@@ -466,8 +484,7 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
 		}
 		chrootcert := certdir + "/Swupd_Root.pem"
 		fmt.Println("Copying Certificate into chroot...")
-		err = helpers.CopyFile(chrootcert, b.Cert)
-		if err != nil {
+		if err = helpers.CopyFile(chrootcert, b.Cert); err != nil {
 			helpers.PrintError(err)
 			return err
 		}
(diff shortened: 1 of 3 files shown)


src/builder/builder.go
```diff
@@ -200,15 +200,38 @@ func (b *Builder) UpdateRepo(ver string, allbundles bool) {
...
+			if err != nil {
+				helpers.PrintError(err)
+				os.Exit(1)
+			}
+
+			err = helpers.CopyFile(bundles+"/bootloader", clrbundles+"bootloader", true)
+			if err != nil {
+				helpers.PrintError(err)
+				os.Exit(1)
+			}
```
kevincwells: One of my absolute least favorite things about Go's "return an error as one of your return values" paradigm is that it results in code constantly being peppered with `if err != nill {  // Print error & exit }` blocks, just like this, where 4 lines become 16. You're not doing anything wrong here -- it's "how Go works" -- but it makes code a lot a lot harder to read IMHO.

kevincwells: As a separate issue, the actual formatting of these (and many others) could/should be changed to be more idiomatic, using Go's initializer statement for `if` statements. Something like:

```go
if err = helpers.CopyFile(bundles+"/bootloader", clrbundles+"bootloader", true); err != nil {
    helpers.PrintError(err)
    os.Exit(1)
}
````

File: src/mixer/main.go [hunk 2]
```diff
@@ -224,6 +226,69 @@ func cmdAddBundles(args []string) {
...
+
+		re := regexp.MustCompile("<(.)+>\n")
+		data = re.ReplaceAllString(data, "file://"+pwd+"/update/www\n")
+
+		err = ioutil.WriteFile(pwd+"/builder.conf", []byte(data), 0666)
+	}
+
+	if *confOnlyflag {
+		return
+	}
```
kevincwells: Please move the work payload (line 237 through this one) into the builder class. Main.go is currently used for arg parsing and input validation, then passing off to the builder class to do the actual work.

src/mixer/main.go
```diff
@@ -224,6 +226,69 @@ func cmdAddBundles(args []string) {
...
+	} else {
+		raw, err := ioutil.ReadFile(pwd + "/builder.conf")
+		if err != nil {
+			helpers.PrintError(err)
+			os.Exit(1)
+		}
+
+		data := strings.Replace(string(raw), "/home/clr/mix", pwd, -1)
+		data += "\nRPMDIR=" + pwd + "/rpms\n"
+		data += "REPODIR=" + pwd + "/local\n"
```
kevincwells: This might not do what you want it to.  This will append it to the end of the file, but that will put it in the [Server] section of the TOML* file.

\* Technically it's not TOML, it's INI. The file exists for the bundle-chroot-builder, which is written in Python and uses [the configparser library](https://docs.python.org/3/library/configparser.html), which only handles INI not TOML.

src/mixer/main.go
```diff
@@ -224,6 +226,69 @@ func cmdAddBundles(args []string) {
...
+	}
+
+	if *confOnlyflag {
+		return
+	}
+
+	b := builder.New()
+	b.LoadBuilderConf(pwd + "/builder.conf")
+	b.ReadBuilderConf()
+	b.InitMix(strconv.Itoa(*clearflag), strconv.Itoa(*mixflag), *allflag)
```
kevincwells: Up for discussion, but I would consider just merging this functionality with the init-mix command, rather than wrapping it/duplicating all of its parameters.

kevincwells:
My personal preference would be to pull the helpers.CopyFile rewrites out, squash them into a single commit, and submit that as a separate PR. Then this one, also squashed, can rebase on top of that.

kevincwells:
PS - +1 for proper config file parsing. I, too, had a plan in the back of my head for replacing all of the config regex parsing with a standard library. While I am a fan of TOML, and there are many libraries for parsing TOML files ([toml](https://github.com/BurntSushi/toml) for just parsing it, [viper](https://github.com/spf13/viper) for a more full-featured config option, etc), we do need to be careful not to use the more exotic TOML extensions to INI.  As mentioned above in my review, the builder.conf is technically *not* TOML, as it exists for (and is also parsed by) the bundle-chroot-builder, which is written in python and uses [the configparser library](https://docs.python.org/3/library/configparser.html), which is limited to just INI syntax.  It's a moot point for now (basic TOML == INI), but we'd need to update bundle-chroot-builder.py in the future if we ever wanted to use, say, TOML's lists or nested sections.

src/builder/builder.go
rchiossi: Yeah, it was a bit painful to change that, it looked very neat before. I couldn't find any nicer way to handle the errors, so i guess we have to live with that. As for the if initializer, I still look it quite ugly, but I guess that's the Go way.... I'll update it.

src/mixer/main.go
rchiossi: Right now, at least for mixer, the sections are just for show. The "use mixer tool" documentation also don't specify where to put them, so they will usually end up at the end of the file. But I agree, that's not the best place to have it. My plan was to handle it properly while reworking the config parser.

src/mixer/main.go
rchiossi: It's been in and out of the builder.go file during my revisions. My main argument here is that the work performed is not exclusive related to the builder. But I see an argument to do it otherwise.

src/mixer/main.go
rchiossi: Hmm, I think ideally I would have a third separate function that is called by both.

src/mixer/main.go
kevincwells: I see your point, but it's analogous to the work done by builder.InitMix. In general, I am mostly voting for keeping things consistent with how the rest of the code is laid out, for clarity and maintainability.

src/mixer/main.go
kevincwells: They are "for show" for mixer right now, but the configparse library used by bundle-chroot-builder actually does separate things out by section correctly.

The [Server] section is parsed out [here](https://github.com/clearlinux/bundle-chroot-builder/blob/master/bundle-chroot-builder.py#L205) and used [here](https://github.com/clearlinux/bundle-chroot-builder/blob/master/bundle-chroot-builder.py#L182). Based on how they're using it, these new values would be ignored and not hurt anything, but it's a dangerous coupling that results from two different tools relying on the same config file.

While I agree this should be "done properly" in a config file rewrite, for now you could insert it in to the right section by finding the location of "\n\n[swupd]" and inserting them before that.

kevincwells: ... though, I guess REPODIR and RPMDIR are only used by mixer, so there isn't really a "right section". Maybe it should go in its own [Mixer] section?

src/builder/builder.go
tmarcu: I think this is ok for now... if everyone agrees on a consistent implementation let's propagate it throughout the rest of the codebase in another PR.

src/mixer/main.go
tmarcu: @mdhorn @kevincwells  would it be useful to have a --prefix or --path option to this so you can mixer init --path /some/diff/path/than/currdir? Probably out of scope for this PR but good to think about for future improvement.

src/mixer/main.go
tmarcu: I would like to keep code out of this as well esp if cobra will be replacing the UI here, it will get moved around anyway.

src/mixer/main.go
tmarcu: I also do not agree with duplicating the command like this. Can this be reworked and be its own separate thing? I wouldn't be against it replacing the current one even...

src/mixer/main.go
kevincwells: @tmarcu I think that was originally part of the proposal @rchiossi had suggested. I can see uses for that, but I'd be fine with it being a future improvement.

src/mixer/main.go
rchiossi: @tmarcu do you mean dropping init-mix and keeping init only?

src/mixer/main.go
tmarcu: I agree for future, just wanted to mention it.

src/mixer/main.go
tmarcu: @rchiossi yeah..it essentially looks like the exact same code but with nicer dir creation logic and config-only options :) Doesn't seem like it'd interfere with anyone currently using mixer init-mix. I'd argue to just replace the current one and replace it with your implementation, however keep the call init-mix because people currently using mixer may have it scripted/automated to call "init-mix" and not "init" so we don't want to break them right now. Can update that in a separate PR and send out a disclaimer email along with it to make sure everyone moves over smoothly over time. Another option would be to support both options and call the same code for either.

rchiossi:
Did some cleanup on the first two patches and integrated the functionality of init into init-mix as we talked about in IRC. 
With this new version, init-mix keeps the same syntax, but if no builder.conf is found it will generate one. If one is provided, the creation process is skipped. The creation of rpms and local directories was moved into UpdateRepo() as the other dirs were already created here.

src/builder/builder.go
```diff
@@ -190,6 +231,29 @@ func (b *Builder) SignManifestMOM() {
...
+	}
+
+	err = os.Mkdir(pwd+"/local", 0755)
+	if os.IsExist(err) {
+		fmt.Println("Directory '/local' already exists. Skipping...")
+	} else if err != nil {
+		helpers.PrintError(err)
+		os.Exit(1)
+	}
+
```
kevincwells: I don't think this logic can live here yet. It makes sense to me for these things to be be together eventually, but `UpdateRepo` needs to be refactored first.

Right now `UpdateRepo` does two things: 1) it downloads and unpacks a local cache of CLR bundle definitions, and 2) it populates/initializes the `mix-bundles` directory if it doesn't already exist.

Right now it's called in three places: 1) by `mixer init-mix`, 2) by `mixer get-bundles`, and 3) by `mixer add-bundles`

`UpdateRepo` is called in `AddBundles` to take advantage of the CLR bundle download/unpacking. `AddBundles` doesn't want the `mix-bundles` initialization, and takes advantage of the fact it'll skip it if it exists. This coupling is a bit dangerous. With this patch, it would attempt to add these  `rpm` and `local` directories every time you call `mixer add-bundles`, too.

Likewise, these directories would be re-added (and warn with a "already exists" message) anytime someone called `mixer get-bundles`, too. However, I am pushing for this command to be made obsolete, with the tool handling bundle caching automatically for you the way `AddBundles` does.

So really what we probably want is one higher-level (`helper.go`?) function that retrieves and unpacks bundles that can be called by both `UpdateRepo` and `AddBundles`, and then `UpdateRepo` could be thought of more as a "initialize all the directories" function.

src/builder/builder.go
```diff
@@ -61,6 +61,47 @@ func NewFromConfig(conf string) *Builder {
...
+	data := strings.Replace(string(raw), "/home/clr/mix", pwd, -1)
+	data += "\n[Mixer]\n"
+	data += "RPMDIR=" + pwd + "/rpms\n"
+	data += "REPODIR=" + pwd + "/local\n"
+
+	re := regexp.MustCompile("<(.)+>\n")
+	data = re.ReplaceAllString(data, "file://"+pwd+"/update/www\n")
```
kevincwells: Does this actually work as a valid value for `CONTENTURL` or `VERSIONURL`? 

If I understand correctly, this is the value that gets baked into images, telling the resultant swupd-client where to look for updates. 1) Does swupd-client actually understand the `file://` protocol? And 2) this URL is a path on the build host; even if swupd-client understands `file://`, this wouldn't be a valid path, right?

rchiossi: Those values came from here: https://github.com/clearlinux/swupd-client/blob/master/swupd-add-pkg
I figured if this is the template being generated by swupd we should use the same.

src/builder/builder.go
```diff
@@ -61,6 +61,47 @@ func NewFromConfig(conf string) *Builder {
...
+		os.Exit(1)
+	}
+
+	if builderconf == "" {
+		builderconf = pwd + "/builder.conf"
+	}
+
+	err = helpers.CopyFile(builderconf, "/usr/share/defaults/bundle-chroot-builder/builder.conf", false)
+	if os.IsExist(err) {
+		fmt.Println("File 'builder.conf' already exists. Skipping...")
```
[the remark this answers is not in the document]
rchiossi: Indeed

src/builder/builder.go
rchiossi: After reading your proposals for the tool I figured this was a bit out of place. Still, I don't want this PR to keep hanging until the refactor happens. So I think we have two options now: Leave it here and once there is a proper isolated function for creating the directories we move it along or isolate it in a separate function and once the refactor happens remember to move it to the appropriate location. I'm ok with either solution.

kevincwells: I vote for the latter -- wrap the repo directory creation in a separate function and call it during init. 

And, actually, I think it needs to be an optional thing (created and defined in the conf only if you pass "--local-repo" or something like that). This is a little trickier than I realized. Yesterday @mdhorn reminded me that the code in `BuildChroots` checks if `b.Repodir !- ""` to determine how to use `m4` for expanding the yum config. (The section telling it about the local RPMS is only included if the ifdef passes.) This means that as it is right now, your code will *always* define these repos, and so `b.Repodir` will *never* be "", so `m4` will *always* include the local RPM section, even if it's empty.

@tmarcu, can you double check what I just said above?

src/builder/builder.go
tmarcu: Ugh sorry, that's a very specific nuanced feature that was just recently added to swupd..I would ignore it & anything related completely for mixer. Swupd does understand file://, but @kevincwells is right that those URL variables get injected into the image so the client points to the server the mix content is hosted on -> which is in all cases, except that weird swupd one, some other remote location.

rchiossi: Fair enough. In that case, what should we have as defaults?

src/builder/builder.go
tmarcu: Yes, that is what happens @kevincwells . I didn't comment previously since it technically works still...it'll just make yum look into your local repo every time before checking upstream, but that's probably misleading a bit. I like the idea of a --local-repo flag, and/or just removing the code that creates /local and /rpms from this patch for now. The current documentation describes those 2 folders needing to be made, so it's not a big deal if we just keep that functionality as is, and then we can merge this patch and iterate on it afterwards.

src/builder/builder.go
rchiossi: @tmarcu @kevincwells

kevincwells: I don't think this is something you can provide a good default value for. It needs to be the URL at which deployed swupd-clients can reach valid update content -- something equivalent to https://cdn.download.clearlinux.org/update/, but for your derivative mix. The tool cannot predict what that likely is, and hence why it's "&lt;CONTENT URL GOES HERE&gt;" kind of values right now -- something that clearly indicates you need to set it.

If you want the tool to be able to fill these, the best you can likely do is allow the user to pass in the value to `init-mix`.

@tmarcu any better idea?

src/builder/builder.go
kevincwells: @tmarcu if you spot stuff like this, please definitely speak up, as they create dead paths in the code.

I'm not saying that these directories can't get created automatically (I like the idea of not having to make them) or that it's even a bad thing for YUM to look into your local repo every time (is it that much of an optimization to make it not?). But if we go that route, we definitely need to take out the check for these repos, the use of `m4`, and the dead paths in the code. One or the other.

src/builder/builder.go
tmarcu: @rchiossi @kevincwells I can't say what a good default is 'cause it depends fully on where the mix content ends up going. The trivial solution is "<current ip address>/update", assuming you host your content on the build machine under the /update folder, wherever your httpd directory is. However it could be anywhere the httpd conf points, the content might get built on one server but rsync'd over to another release server, etc, so it's hard to make a good default for these options.

rchiossi:
I removed the default values to CONTENTURL and VERSIONURL and isolated the new functionality on their own functions. Other than that, the patch is pretty much he same.

builder/builder.go
```diff
@@ -333,7 +333,7 @@ func (b *Builder) AddBundles(bundles []string, force bool, allbundles bool, git
 			}
 
 			fmt.Printf("Adding bundle %q\n", bundle)
-			if err = helpers.CopyFile(bundledir+bundle, clrbundledir+bundle); err != nil {
+			if err = helpers.CopyFile(bundledir+bundle, clrbundledir+bundle, true); err != nil {
```
cmarcelo: Instead of a boolean parameter, I would suggest one of the following to improve the code at the callsites

- Have two functions: CopyFile and CopyOverrideFile, or something along those lines. They should use a non-exported function that take e.g. the flags, so you don't duplicate implementation.
- Use flags instead of a boolean, so callsite can see either the word override or exclusive.
- Allow passing "extra" flags, so call sites would pass O_EXCL (kind of lazy version of the above).

cmarcelo:
@rchiossi suggestion: unless this will get merged soon I suggest you push a PR with the CopyFile changes (since they are more straightforward to get in).

tmarcu:
Sorry to have pushed this one out so far, I can help with rebasing this onto the new CLI/structure if you'd like to get it all in.

rchiossi:
I don´t mind rebasing when I get back from vacations, unless you need this feature in before the end of next week.

tmarcu:
There's no rush for this specifically - let's catch up when you're back!

kevincwells:
Hey @rchiossi, I was hoping you could update this PR and rebase, or submit a new one with a subset of the functionality sooner than later.

The stuff I'm working on right now involves changing where a few files live, but the paths to those files are defined in the `builder.conf` (`VERSIONS_PATH`, for example). To do this, I need to edit the default values in the conf, and add a few other things. What I really want is your `CreateDefaultConfig` function from your 3rd commit merged, so I can do what I need to in there. I have a couple hacks in my code to keep me going for now, but I'd like to rebase on top of this.

PS - When you're updating it, make sure to return errors back up to their calling function rather than `fmt.Println()` and `os.Exit(1)` from within `builder.go`; that's the new convention we're trying to enforce in the code.
PPS - I think the `CopyFile` changes are absolutely worth getting merged too, they're just not urgent.

kevincwells:
Hey @rchiossi -- I took out the hacks I'd put in place and instead just cherry-picked the commit from your branch that I needed. Please look [my branch](https://github.com/kevincwells/mixer-tools/commits/bundle-refactor) (and specifically ef3056ad95e3f737a3a95fe116ced692fba0acab) to see what I did. As you can see, you're still the author of the commit.

I had to change a couple small things to integrate it
1. I passed all the errors back up to the calling function as I mentioned
1. I took out the `builderconf` paramater you had passed to `CreateDefaultConfig`. Technically your code would see the passed-in config and skip it when CopyFile errors, but it seemed easier (and safer) to just avoid this path altogether.  Now, a default config file is only created if the user doesn't pass in their own.
1. I moved the call to `createRpmDirs` to inside `InitMix` to line up with where the other directory creation occurs and made it un-exported ("C"-&gt;"c"). Unlike `CreateDefaultConfig`, this one doesn't need to occur before any config parsing, and can thus safely be called from within `InitMix`.
1. I had to take out the last parameter in your call to `CopyFile`, 'cause I didn't rebase your changes to `helpers.go`. I still definitely want those to get in.

rchiossi:
Hey Kevin,
After reviewing this PR I decided to drop most of it and use the save function from Viper lib to create the default config file instead of copying it from a hardcoded path. Since we know the default  values, there is not really a need to copy it. That would also make the changes to the copy function I made unnecessary for this feature.
As for you cherry pick, I don't think it is a good idea to use it without the changes to CopyFile since the behavior will change. In your branch, if the config exists, it will be overwritten and yield a message saying that it already exists and was skipped.
I'll fix this branch and open a new PR today so you can get things going. It is OK for me to replace the code later with the config save.

PR #47, opened 2017-12-19
builder: create a temp dir and use as stage dir for ister

cmarcelo:
Mixer calls ister, that calls swupd (client), that uses a staging
directory to cache downloaded files. If not specified, it defaults to
the system one, which is not what we want, since the system might not
be using the mix we are creating.

Without passing this, build-image works only if there is no conflict
with some previous version (from any stream) that it used before.

Fix based on diagnostic by Tudor and William.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -611,12 +611,22 @@ func (b *Builder) BuildImage(format string, template string) {
 		template = "release-image-config.json"
 	}
 
+	// swupd (client) called by itser will need a temporary directory to act as its stage dir.
+	wd, _ := os.Getwd()
+	tempStage, err := ioutil.TempDir(wd, "ister-swupd-client-")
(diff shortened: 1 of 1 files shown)


builder/builder.go
```diff
@@ -611,12 +611,22 @@ func (b *Builder) BuildImage(format string, template string) {
 		template = "release-image-config.json"
 	}
 
+	// swupd (client) called by itser will need a temporary directory to act as its stage dir.
+	wd, _ := os.Getwd()
+	tempStage, err := ioutil.TempDir(wd, "ister-swupd-client-")
+	if err != nil {
+		// TODO: This should return a proper error and the caller deals with printing.
+		helpers.PrintError(err)
```
tmarcu: In another PR, lets change address this issue.

PR #49, opened 2017-12-20
Update the Makefile

### LINKED ISSUE #40: proposal: follow Go conventions for mixer-tools repository
At the moment the mixer-tools repository is organized that it assumes it will be set as the GOPATH to operate. This is not the conventional way to do things, which means that some of the tooling around Go either don't work or need special care to work.

The conventional way is to assume the sources will be inside an existing GOPATH, in the github.com/clearlinux/mixer-tools directory.

Changes needed:
- Move the contents of src/ into the root of the repository
- Change import paths to use the new name of the package
- Update the Makefile to work in this new setting. Make it so that the Makefile works regardless where the mixer-tools sources is (so it would still set the GOPATH for now). For this change would be helpful to know what Makefile targets are used besides "make", "make check" and "make install".

cmarcelo:
Update the Makefile to work when we are outside a GOPATH. Note that developers are still encouraged to be inside a GOPATH. The Makefile is mostly intended for package creation.

Fixes #40.

This also fixes the recent regression of "make" not working with release tarballs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ gopath:
 	@echo "Code already in existing GOPATH=${GOPATH}"
 endif
 
-.PHONY: all compliant build install clean
+.PHONY: all compliant build install clean check
 
 .DEFAULT_GOAL := all
 all: compliant build
@@ -51,6 +51,9 @@ install: gopath
 	install -m 00755 pack-maker.sh $(DESTDIR)/usr/bin/mixer-pack-maker.sh
 	install -m 00755 superpack-maker.sh $(DESTDIR)/usr/bin/mixer-superpack-maker.sh
 	install -D -m 00644 yum.conf.in $(DESTDIR)/usr/share/defaults/mixer/yum.conf.in
+
+check: gopath
+	go test ${GO_PACKAGE_PREFIX}/...
 
 clean:
 ifeq (,${LOCAL_GOPATH})
--- a/Makefile
+++ b/Makefile
@@ -60,11 +60,15 @@ else
 endif
 	rm -f mixer-tools-*.tar.gz
 
-VERSION = 3.2.0
 release:
 	@if [ ! -d .git ]; then \
 		echo "Release needs to be used from a git repository"; \
 		exit 1; \
 	fi
-	git archive --format=tar.gz --verbose -o mixer-tools-$(VERSION).tar.gz HEAD --prefix=mixer-tools-$(VERSION)/
+	@VERSION=$$(grep -e 'const Version' mixer/main.go | cut -d '"' -f 2) ; \
+	if [ -z "$$VERSION" ]; then \
+		echo "Couldn't extract version number from the source code"; \
+		exit 1; \
+	fi; \
+	git archive --format=tar.gz --verbose -o mixer-tools-$$VERSION.tar.gz HEAD --prefix=mixer-tools-$$VERSION/
 
--- a/Makefile
+++ b/Makefile
@@ -1,28 +1,70 @@
-PROJECT_ROOT := src/
-VERSION = 3.2.0
+# Makefile used to create packages for mixer-tools. It doesn't assume
+# that the code is inside a GOPATH, and always copy the files into a
+# new workspace to get the work done. Go tools doesn't reliably work
+# with symbolic links.
+#
+# For historical purposes, it also works in a development environment
+# when the repository is already inside a GOPATH.
+
+.NOTPARALLEL:
+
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,28 +1,67 @@
...
-install:
+install: gopath
 	test -d $(DESTDIR)/usr/bin || install -D -d -m 00755 $(DESTDIR)/usr/bin;
 	install -m 00755 $(GOPATH)/bin/mixer $(DESTDIR)/usr/bin/.
 	install -m 00755 pack-maker.sh $(DESTDIR)/usr/bin/mixer-pack-maker.sh
 	install -m 00755 superpack-maker.sh $(DESTDIR)/usr/bin/mixer-superpack-maker.sh
 	install -D -m 00644 yum.conf.in $(DESTDIR)/usr/share/defaults/mixer/yum.conf.in
 
+clean:
+	rm -f $(GOPATH)/bin/mixer
```
icarus-sparry: GOPATH can be a colon separated list of directories.

Use ```go clean -i -x``` 

Should it clean up pack-maker.sh etc as well?

Makefile
```diff
@@ -1,28 +1,67 @@
...
+LOCAL_GOPATH :=
+GOPATH ?= ${HOME}/go
+gopath:
+	@echo "Code already in existing GOPATH=${GOPATH}"
+endif
+
+.PHONY: all compliant build install clean
 
 .DEFAULT_GOAL := all
+all: gopath compliant build
```
icarus-sparry: I would think about removing gopath from all.

I am not sure how good an idea it is to be potentially running compliant and build in parallel, if compliant updates a file as build is reading it. Probably will not happen in practice as the code will be formatted almost all the time.

Makefile
```diff
@@ -1,28 +1,67 @@
...
 
-CHANGES := $(shell go fmt ./... 2>&1)
+compliant: gopath
+	CHANGED_FILES=$$(go fmt ${GO_PACKAGE_PREFIX}/... 2>&1) ; \
+	if [ -n "$$CHANGED_FILES" ]; then \
+		echo -e "ERROR: go fmt had to fix the following files:\n$$CHANGED_FILES"; \
```
icarus-sparry: printf 'ERROR: go fmt had to fix the following:\n%s\n' "$CHANGED_FILES" >&2 \

Makefile
cmarcelo: Thanks for the go clean tip :-)

pack-maker.sh is part of the sources.

Makefile
```diff
@@ -1,28 +1,67 @@
...
+# For historical purposes, it also works in a development environment
+# when the repository is already inside a GOPATH.
+
+GO_PACKAGE_PREFIX := github.com/clearlinux/mixer-tools
+
+.PHONY: gopath
+
+# Strictly speaking we should check if it the directory is inside an
+# actual GOPATH, but the directory structure matching is likely enough.
+ifeq (,$(findstring ${GO_PACKAGE_PREFIX},${PWD}))
```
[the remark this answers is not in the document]
cmarcelo: Changed to CURDIR.

Makefile
```diff
@@ -1,28 +1,67 @@
...
+gopath:
+	@rm -rf ${LOCAL_GOPATH}/src
+	@mkdir -p ${LOCAL_GOPATH}/src/${GO_PACKAGE_PREFIX}
+	@cp -af * ${LOCAL_GOPATH}/src/${GO_PACKAGE_PREFIX}
+	@echo "Prepared a local GOPATH=${GOPATH}"
+else
+LOCAL_GOPATH :=
+GOPATH ?= ${HOME}/go
+gopath:
+	@echo "Code already in existing GOPATH=${GOPATH}"
```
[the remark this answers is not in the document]
cmarcelo: I've kept since it will help diagnosing trouble with the new Makefile. We can drop it later.

Makefile
cmarcelo: Thanks for noticing this. Added .NOTPARALLEL to the whole Makefile.

Makefile
cmarcelo: Done.

cmarcelo:
@icarus-sparry @tmarcu could you please take another look?

Makefile
icarus-sparry: I was rather thinking of removing the copy of` pack-maker.sh` from `/usr/bin` but perhaps the longer term fix is to add a `make uninstall` target?

tmarcu:
My testing shows it working as the previous setup, lgtm.

PR #51, opened 2017-12-23
Makefile: add a 'make lint' to run gometalinter.v1

cmarcelo:
Make easy to run the same checks as travis.

Signed-off-by: Caio Marcelo de Oliveira Filho <caio.oliveira@intel.com>

PR #52, opened 2017-12-23
Do not enforce a specific cyclomatic complexity

cmarcelo:
Some sequential series of steps are better handled in a long function
than in multiple small functions. Forcing a low number might make
those functions "pass the test" but less readable in the end.
    
The gocyclo can be an useful tool but it shouldn't be used to enforce
a certain limit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ lint: gopath
 	--enable=vet \
 	--enable=ineffassign \
 	--enable=gofmt \
-	--enable=gocyclo --cyclo-over=10 \
+	$${CYCLO_MAX:+--enable=gocyclo --cyclo-over=$${CYCLO_MAX}} \
 	--enable=golint \
 	--enable=deadcode \
 	--enable=varcheck \


cmarcelo:
Depends on #51.

tmarcu:
@matthewrsj @icarus-sparry @phmccarty I agree with @cmarcelo that this is sometimes a ridiculous constraint, but I don't know about taking it out completely either. What is the consensus on either:
1.) Making it higher (it should be higher anyway IMHO)
2.) Making it off/high by default but able to be passed in to change, i.e add to Makefile `CYCLO_MAX ?= 30` (or 100 whatever), set with `make lint CYCLO_MAX=12 `for local tests.
This does not measure sequential steps, it measure paths which are independent of each other. And this can make code awful to read and maintain -> refer to the endless functions in swupd source code. I'm more in favor of #2 so it's easy to use still when needed.

cmarcelo:
To be clear: sequential set of steps with error checking and conditionals in the middle. E.g. the main.go for create-fullfiles program.

Makefile
```diff
@@ -57,17 +57,19 @@ check: gopath
 
 .PHONY: lint
 lint: gopath
-	@gometalinter.v1 --deadline=10m --tests --vendor --disable-all \
+	@extra_args= ;\
+	[ -z "${CYCLO_MAX}" ] || extra_args="--enable=gocyclo --cyclo-over=${CYCLO_MAX}" ;\
+	gometalinter.v1 --deadline=10m --tests --vendor --disable-all \
```
tmarcu: This won't run if gometalinter is not installed to /usr/bin, should probably use `$(GOPATH)/bin/gometalinter.v1`

cmarcelo: It seems to work for the travis. I have $GOPATH/bin as part of my $PATH -- I guess by now most of us do.

In the case we are using a "new GOPATH" we'll need to keep track of the old one for the sake of this command line call. So, before I enter that rabbit hole, is this a real issue for you or an hypothetical case?

tmarcu: Yes this works for Travis, but it doesn't work for a local build which doesn't install gometalinter or add GOPATH to $PATH. If $GOPATH/bin is already in your path just calling it with $GOPATH/bin/<program> shouldn't cause any problems and cover both cases IMO. It's already being used for the install target too  :)

tmarcu:
@cmarcelo  can you also update the travis config in this PR as well?

Makefile
tmarcu: The compliant target is not needed anymore with this change since fmt and vet will be run, and it exits appropriately with an exit by default.

icarus-sparry: You can not in general use $GOPATH/bin/metalinter.v1. This is because GOPATH, like PATH, is a colon separated list of directories. I agree that the common case is for it to be just a single one, but just as you can't say $PATH/ls, you can't append bin/metalinter.v1 to $GOPATH.

If you want to do something like that, this, then I suggest the following is the correct way to do it. Works with an empty GOPATH, one with a single directory in it, works with multiple directories in it. 

     _G=${GOPATH:-$HOME/go}:
     PATH=${_G//:/\/bin:}$PATH metalinter.v1

(works by setting _G to the GOPATH or the default one if it is not set, and appends a colon, then changes all colons to  slash bin colon, and invokes metalinter.v1 with this prepended onto the PATH so it is found first.

Makefile
```diff
@@ -57,17 +57,19 @@ check: gopath
...
 	--enable=vet \
 	--enable=ineffassign \
 	--enable=gofmt \
-	--enable=gocyclo --cyclo-over=10 \
 	--enable=golint \
 	--enable=deadcode \
 	--enable=varcheck \
 	--enable=structcheck \
 	--enable=unused \
+	$$extra_args \
```
icarus-sparry: The proposed change could be written more succulently. If the intention is to use a Makefile variable, then use a makefile construct to generate the line. If the intention is to use a shell variable, then use parameter expansion, e.g.

            ${CYCLO_MAX:+ --enable-gocyclo --over-cyclo=$CYCLO_MAX} \
(double the dollars for inside a Makefile of course)

cmarcelo: Changed to use ":+". Thanks for the tip.

Makefile
cmarcelo: > Yes this works for Travis, but it doesn't work for a local build which doesn't install gometalinter or add GOPATH to $PATH.

This is more about #51. It wasn't my intention to go all the way to automatically install gometalinter -- since it is a bit of a rabbit hole to deal with all the cases. Note that before the commit in question, we had to copy/paste to run these tests :-P

Having a target for installing the lint tools could be useful, but I thought it was less important than 'make lint' itself.

cmarcelo: > This is because GOPATH, like PATH, is a colon separated list of directories.

Yep. The Makefile still uses it in a way that is not 100% guaranteed to work (but was there before). We'll fix that eventually.

All this becomes more complicated because we are using this Makefile both as "helper for developer that follows Go conventions" and "helper for packaging that don't care about go conventions". Wondering if maybe the real fix is to teach autospec a bit more about go. E.g. something like Travis' go_import_path keyword would make things easier.

Makefile
tmarcu: Nice!

PR #60, opened 2017-12-26
Remove 'make compliant' target from Makefile

cmarcelo:
Redundant now that 'make lint' is around. Our integration system
already runs 'make lint', so not necessary to enforce 'make lint' in
all the builds.


cmarcelo:
Per @tmarcu request.

PR #67, opened 2018-01-03
Clean up the 'mixer add-rpms' codepath

cmarcelo:


cmarcelo:
@tmarcu rewrote the commit message of one of the patches and the PR title.

cmarcelo:
> Or, update the rest of the paths to return error instead of calling helpers.PrintErrror() and os.Exit(1)... I think everything should return a useful error, with only a few places really needing to os.Exit(1) immediately.

Agreed. There will be other patches taking care of other paths in builder.

PR #68, opened 2018-01-03
mixer --version and mixer --check

cmarcelo:
This command will always output the mixer version, regardless of the
dependencies (so callers don't need to worry about extraneous output
in their scripts).

Also make the error message for failed dependency less redundant: the
error from LookPath already says the name of the program that wasn't
found.

### DIFF
--- a/mixer/cmd/root.go
+++ b/mixer/cmd/root.go
@@ -25,13 +25,29 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// Version of Mixer. Also used by the Makefile for releases.
+const Version = "3.2.1"
+
 var config string
 
 // RootCmd represents the base command when called without any subcommands
 var RootCmd = &cobra.Command{
 	Use:  "mixer",
 	Long: `Mixer is a tool used to compose OS update content and images.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if rootCmdFlags.version {
+			fmt.Printf("Mixer %s\n", Version)
+			os.Exit(0)
+		}
+		// Use cmd here to print exactly like other prints of Usage (that might be
+		// configurable).
+		cmd.Print(cmd.UsageString())
+	},
 }
+
+var rootCmdFlags = struct {
+	version bool
+}{}
 
 type initCmdFlags struct {
 	all         bool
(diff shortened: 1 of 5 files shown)


cmarcelo:
@mdhorn could you take a look?

mdhorn:
Does not Cobra directly support the -v  and --version? Can we not centralize the the location of the version string and use that?

For the dependency checking, how about creating a function in helpers which takes a list/slice of dependencies to verify. The each base command can call the helper function with a list of required commands for that specific command. We only check the dependencies needed for that command base.
This way `mixer bundle add` does not fail if `ister` is not found, but `mixer build image` would.

cmarcelo:
> Now with that in mind, I'd ultimately prefer version to be a flag (-v and --version), which Cobra will automatically generate if you set the Version field on the root cmd:

Note that the latest release of Cobra (the one we are using) doesn't has this yet.

I will change the patch to use a flag instead of a command. I will not use "-v" shorthand (Cobra newer code doesn't use it too) since I like that to be reserved for a flag to control verbosity.

For the whitelisting, I'll avoid it by trying Mark's suggestion. I'm a bit unsure if the granularity is worthwhile, but let's see.

cmarcelo:
> Does not Cobra directly support the -v and --version?

It gives you automatic --version, but that's from a non-released version of cobra.

> Can we not centralize the the location of the version string and use that?

Not sure what you mean. The string is already centralized.

cmarcelo:
> For the dependency checking, how about creating a function in helpers which takes a list/slice of dependencies to verify. (...)

Done, not exactly the way you specified but I hope the effect is the same.

cmarcelo:
@mdhorn @kevincwells could you please take another look?

Main differences:
- Use a option: 'mixer --version'.
- Make dependency check per command.
- And a standalone: 'mixer --check' that performs check of all dependencies and put up a report.

cmarcelo:
Forgot to mention:
- Per request: remove the version from every output.

mixer/cmd/root.go
```diff
@@ -63,36 +99,78 @@ func Execute() {
...
+			continue
+		}
+		_, err := exec.LookPath(dep)
+		if err != nil {
+			fmt.Printf("  %-*s not found\n", max, dep)
+			ok = false
+		} else {
+			fmt.Printf("  %-*s ok\n", max, dep)
+		}
+	}
```
kevincwells: Both here and in checkCmdDeps (at line 130), you have a "loop through dependencies, skipping duplicates, and call exec.LookPath" block that I probably would have pulled out and made a common thing. But I see you're doing a different thing in the output, and the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) says if you need it twice just copy, so I'm fine with it how it is.

PR #71, opened 2018-01-04
helpers: remove unused ENOVERSION constant

cmarcelo:


kevincwells:
+1

I don't know the origin of this variable, but it does appear to be dead code now.

PR #72, opened 2018-01-04
Write three more tests for manifest creation

matthewrsj:
* Deduplicate bundle includes list and add test 
* Set deleted file hash to AllZeroHash 
  When a file is deleted set its hash to all zeros. Add a test to confirm this.
* Restructure processBundles to perform tasks in correct order 
  An added test identified issues with how processBundles was adding, removing, and changing files. Certain steps need to happen to *all* bundles before the next group of steps can be completed. For instance, all manifest.Files manipulation must be done before subtracting included bundles, which in turn must be done before detecting if a file has changed.

### DIFF
--- a/swupd/config.go
+++ b/swupd/config.go
@@ -161,6 +161,16 @@ func readLastVerFile(path string) (uint32, error) {
 	return uint32(parsed), nil
 }
 
+func appendUnique(ss []string, s string) []string {
+	for _, e := range ss {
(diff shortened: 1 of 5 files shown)


matthewrsj:
This PR is to the *swupd* branch.

matthewrsj:
Those indeed can be combined, thanks! I'll update that before merging.

PR #74, opened 2018-01-05
builder: clean up BuildUpdate

cmarcelo:
Main changes:

- Rename parameters to reflect better their intention. In particular
  "signflag" -> "skipSigning"

- Remove unnecessary Stat check for dir creation: MkdirAll will do
  that and not fail if the directory already exists.

- Changed commands to do like the swupd_create_update and their output
  the error directly to the os.Stdout/os.Stderr.

- Changed most of file path building to use filepath.Join.

- Inline setVersion, since these two steps make sense at build
  update level.

- Only create upstreamver file if we create an upstreamurl file, and do
  that before the final publishing step.

- Added comments to clarify some of the steps.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -522,136 +522,119 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
 	return nil
 }
 
-// Set the published versions to what was just built
-func (b *Builder) setVersion(publish bool) {
-	if publish == false {
-		return
-	}
-
-	// Create the www/version/format# dir if it doesn't exist
-	formatdir := b.Statedir + "/www/version/format" + b.Format
-	if _, err := os.Stat(formatdir); os.IsNotExist(err) {
-		if err = os.MkdirAll(formatdir, 0777); err != nil {
-			helpers.PrintError(err)
-			os.Exit(1)
-		}
-	}
-
-	fmt.Println("Setting latest version to " + b.Mixver)
-	err := ioutil.WriteFile(formatdir+"/latest", []byte(b.Mixver), 0644)
-	if err != nil {
-		helpers.PrintError(err)
-		os.Exit(1)
-	}
-
-	err = ioutil.WriteFile(b.Statedir+"/image/LAST_VER", []byte(b.Mixver), 0644)
-	if err != nil {
-		helpers.PrintError(err)
-		os.Exit(1)
-	}
-
-	if b.Upstreamurl != "" {
-		fmt.Println("Saving the upstream version URL " + b.Upstreamurl)
-		upstreamurl := b.Statedir + "/www/" + b.Mixver + "/upstreamurl"
-		err = ioutil.WriteFile(upstreamurl, []byte(b.Upstreamurl), 0644)
-		if err != nil {
-			helpers.PrintError(err)
-			os.Exit(1)
-		}
-	}
-	fmt.Println("Saving the upstream version " + b.Clearver)
-	upstreamver := b.Statedir + "/www/" + b.Mixver + "/upstreamver"
-	err = ioutil.WriteFile(upstreamver, []byte(b.Clearver), 0644)
-	if err != nil {
-		helpers.PrintError(err)
-		os.Exit(1)
-	}
-
-}
-
-// CleanChroots will remove chroots based on what bundles are defined
-func (b *Builder) CleanChroots() {
-	files := helpers.GetDirContents(b.Bundledir)
-	basedir := b.Statedir + "/image/" + b.Mixver + "/"
-
-	for _, f := range files {
-		if f.Name() == "full" {
-			continue
-		}
-		err := os.RemoveAll(basedir + f.Name())
-		if err != nil {
-			helpers.PrintError(err)
-		}
-	}
-}
-
 // BuildUpdate will produce an update consumable by the swupd client
-func (b *Builder) BuildUpdate(prefixflag string, minvflag int, formatflag string, signflag bool, publishflag bool, keepchrootsflag bool) error {
-	if formatflag != "" {
-		b.Format = formatflag
+func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string, skipSigning bool, publish bool, keepChroots bool) error {
+	if format != "" {
+		b.Format = format
 	}
 
-	if _, err := os.Stat(b.Statedir + "/www/version/format" + b.Format); os.IsNotExist(err) {
-		if err = os.MkdirAll(b.Statedir+"/www/version/format"+b.Format, 0777); err != nil {
-			helpers.PrintError(err)
-			return err
-		}
+	// Ensure the format dir exists.
+	formatDir := filepath.Join(b.Statedir, "www", "version", "format"+b.Format)
+	err := os.MkdirAll(formatDir, 0777)
+	if err != nil {
+		return errors.Wrapf(err, "couldn't create the format directory")
 	}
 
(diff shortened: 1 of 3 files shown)


builder/builder.go
```diff
@@ -522,137 +522,135 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
+	formatDir := filepath.Join(b.Statedir, "www", "version", "format"+b.Format)
+	err := os.MkdirAll(formatDir, 0777)
 	if err != nil {
-		helpers.PrintError(err)
-		os.Exit(1)
+		return errors.Wrapf(err, "couldn't create the format directory")
 	}
 
-	err = ioutil.WriteFile(b.Statedir+"/image/LAST_VER", []byte(b.Mixver), 0644)
+	//
```
tmarcu: extra comment?

builder/builder.go
```diff
@@ -522,137 +522,135 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
+	// Create update metadata for the mix.
+	//
```
tmarcu: here too

builder/builder.go
```diff
@@ -522,137 +522,135 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
-func (b *Builder) CleanChroots() {
-	files := helpers.GetDirContents(b.Bundledir)
-	basedir := b.Statedir + "/image/" + b.Mixver + "/"
-
-	for _, f := range files {
-		if f.Name() == "full" {
-			continue
-		}
-		err := os.RemoveAll(basedir + f.Name())
+	//
```
tmarcu: can we not do these kinds of comments please :)

builder/builder.go
```diff
@@ -522,137 +522,135 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
-		helpers.PrintError(err)
-		return err
+		return errors.Wrapf(err, "couldn't create zero packs")
 	}
 
-	// We only need the full chroot from this point on, so cleanup the others to save space
-	if keepchrootsflag == false {
-		b.CleanChroots()
+	//
+	// Hardlink the duplicate files. This helps when keeping the bundle chroots, so
```
tmarcu: s/so/to

builder/builder.go
```diff
@@ -522,137 +522,135 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
+	hardlinkcmd.Stderr = os.Stderr
+	err = hardlinkcmd.Run()
+	if err != nil {
+		return errors.Wrapf(err, "couldn't perform hardlink step")
 	}
 
-	// Step 1.5: sign the Manifest.MoM that was just created
-	if signflag == false {
-		err = b.SignManifestMoM()
+	//
```
tmarcu: here too

builder/builder.go
```diff
@@ -522,137 +522,135 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
 	if err != nil {
-		helpers.PrintError(err)
-		return err
+		return errors.Wrapf(err, "couldn't create zero packs")
 	}
 
-	// We only need the full chroot from this point on, so cleanup the others to save space
-	if keepchrootsflag == false {
-		b.CleanChroots()
+	//
```
tmarcu: and here

builder/builder.go
```diff
@@ -522,137 +522,135 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
+	if err != nil {
+		return errors.Wrapf(err, "couldn't create fullfiles")
 	}
 
-	if _, err := os.Stat(b.Statedir + "/www/version/format" + b.Format); os.IsNotExist(err) {
-		if err = os.MkdirAll(b.Statedir+"/www/version/format"+b.Format, 0777); err != nil {
-			helpers.PrintError(err)
-			return err
-		}
+	//
```
tmarcu: por que

builder/builder.go
```diff
@@ -522,137 +522,135 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
+		upstreamURLFile := filepath.Join(b.Statedir, "www", b.Mixver, "/upstreamurl")
+		err = ioutil.WriteFile(upstreamURLFile, []byte(b.Upstreamurl), 0644)
 		if err != nil {
-			helpers.PrintError(err)
-			return err
+			return errors.Wrapf(err, "couldn't write upstreamurl file")
+		}
+		fmt.Printf("Saving the upstream version: %s\n", b.Clearver)
+		upstreamVerFile := filepath.Join(b.Statedir, "www", b.Mixver, "upstreamver")
+		err = ioutil.WriteFile(upstreamVerFile, []byte(b.Clearver), 0644)
```
tmarcu: I think this is fine, @mdhorn you said this was OK to move to a different part of the code as well right?

builder/builder.go
```diff
@@ -522,137 +522,135 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
+			fmt.Fprintf(os.Stderr, "Ignored error when cleaning bundle chroots: %s", err)
+		}
+		basedir := filepath.Join(b.Statedir, "image", b.Mixver)
+		for _, f := range files {
+			if f.Name() == "full" {
+				continue
+			}
+			err = os.RemoveAll(filepath.Join(basedir, f.Name()))
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Ignored error when cleaning bundle chroots: %s", err)
```
tmarcu: @phmccarty @matthewrsj Is this something worth printing? Do we care whether or not those files really got removed?

helpers/helpers.go
```diff
@@ -222,18 +222,6 @@ func Download(filename string, url string) (err error) {
 	return nil
 }
 
-// GetDirContents is an an assert-style helper to get the contents of a
-// directory, or to exit on failure.
-func GetDirContents(dirname string) []os.FileInfo {
```
tmarcu: Why?

builder/builder.go
matthewrsj: I think this would work better as a `TODO` comment in the code if we should be caring about the errors.

cmarcelo: Mixer says it will clean those (unless you pass a flag), but it couldn't. Since the process can proceed, we just warn the user about the problem.

helpers/helpers.go
cmarcelo: This function was not being used anymore.

I've moved the only user of the function to use ioutil.ReadDir directly, because I did not want the "PrintError/os.Exit" which was the only extra thing provided by GetDirContents.

builder/builder.go
cmarcelo: Idea was to separate the main sections, but it failed to accomplish that. Removed.

builder/builder.go
cmarcelo: Rewrote the text.

builder/builder.go
tmarcu: We could print at the end instead of in the loop as well, to prevent spamming things. This would be interesting if it really failed though, @cmarcelo we can keep this and see how it goes if people actually run into it.

helpers/helpers.go
tmarcu: Ok :+1:  :)

PR #75, opened 2018-01-05
Add more tests and fix identified issues

matthewrsj:
* Add more tests for manifest creation
  Add two more tests for CreateManifests. One to test that the state flag is being set correctly for state files and another to test that empty directories are correctly added to the manifest.
* Fix Manifest.MoM writing
  There were several issues with Manifest.MoM writing. The filenames included the "Manifest." prefix and the wrong old version was being used to populate the old bundle manifests, resulting in manifests that would update every other version regardless of changed content. This adds a
test and fixes the bugs.

This PR is against the **swupd** branch. I created the `swupd` label to indicate swupd-server rewrite PRs and issues in the future.

### DIFF
--- a/swupd/create_manifests_test.go
+++ b/swupd/create_manifests_test.go
@@ -54,16 +54,12 @@ func TestCreateManifests(t *testing.T) {
 	defer removeIfNoErrors(t, testDir)
 	mustInitStandardTest(t, testDir, "0", "10", []string{"test-bundle"})
 	mustGenFile(t, testDir, "10", "test-bundle", "foo", "content")
-	if err = CreateManifests(10, false, 1, testDir); err != nil {
-		t.Fatal(err)
-	}
+	mustCreateManifestsStandard(t, 10, testDir)
 
 	// set last version to 10
 	mustInitStandardTest(t, testDir, "10", "20", []string{"test-bundle"})
 	mustGenFile(t, testDir, "20", "test-bundle", "foo", "new content")
-	if err = CreateManifests(20, false, 1, testDir); err != nil {
-		t.Error(err)
-	}
+	mustCreateManifestsStandard(t, 20, testDir)
 
 	lines := []struct {
 		sub      []byte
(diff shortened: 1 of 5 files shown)


swupd/create_manifests_test.go
```diff
@@ -398,3 +398,38 @@ func TestCreateManifestIncludeVersionBump(t *testing.T) {
 	mustNotExist(t, filepath.Join(testDir, "www/30/Manifest.test-bundle"))
 	mustNotExist(t, filepath.Join(testDir, "www/30/Manifest.included2"))
 }
+
+func TestCreateManifestsState(t *testing.T) {
+	testDir := mustSetupTestDir(t, "state")
+	defer removeIfNoErrors(t, testDir)
+	mustInitStandardTest(t, testDir, "0", "10", []string{})
+	mustGenFile(t, testDir, "10", "os-core", "var/lib/test", "test")
+	if err := CreateManifests(10, false, 1, testDir); err != nil {
```
icarus-sparry: Perhaps we need a MustCreateManifests function for testing?

swupd/create_manifests_test.go
```diff
@@ -398,3 +398,38 @@ func TestCreateManifestIncludeVersionBump(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	re := regexp.MustCompile("D\\.s\\.\t.*\t10\t/var/lib\n")
```
icarus-sparry: Here we are doing a very textual test, seeing that we have a particular string (well regular expression) in the file. Would it be better to have a more logical test, saying do we have a File{Flags: something, Version: 10, Name: "/var/lib"} in the manifest? This would make the test independent of the file format.

swupd/filesystem.go
```diff
@@ -79,9 +79,10 @@ func (m *Manifest) createManifestRecord(rootPath string, path string, fi os.File
 		fmt.Fprint(os.Stderr, err)
 	}
 
-	// remove leading "/" from manifest record
+	// the most recently added file is the one we need to update
 	f := m.Files[len(m.Files)-1]
-	f.Name = strings.TrimLeft(f.Name, "/")
+	// Only the bundle name should be part of the name in the manifest
+	f.Name = strings.Replace(f.Name, "/Manifest.", "", -1)
```
icarus-sparry: This stuff screams "race condition" to me. As a general design would it not be better to have something that creates a record and a separate step that appends the record, which could then be done under lock? This function would then be "create record, hack the name, type and version fiends, append record". Of course we could just declare that Manifest types are not thread safe, and so can only be used from one go routine at a time.

matthewrsj: I see what you mean, yes I suppose appending should be done last. I was just trying to reuse code that already existed.

swupd/create_manifests_test.go
matthewrsj: Well, I do want to check the *format* of the text manifest, since that is important to us at this point. I do like the idea of abstracting that away though, so a test can just call the function to test that a file exists in the manifest and the function can do the heavy lifting of checking the format in the manifest file.

swupd/create_manifests_test.go
matthewrsj: sure

swupd/create_manifests_test.go
matthewrsj: I'm going to defer this one to later, since I have some other pressing things to work on.

PR #79, opened 2018-01-06
Use new swupd package in mixer

cmarcelo:
Add a global flag --new-swupd to enable use of the new swupd package.
When using that flag, swupd.CreateFullfiles will be called.
See individual commit messages for the details.

This PR depends on #74.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -20,6 +20,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -539,13 +540,9 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string, 
 	var err error
 	// TODO: move this to parsing configuration time.
 	var mixVerUint uint32
-	{
-		var parsed uint64
-		parsed, err = strconv.ParseUint(b.Mixver, 10, 32)
-		if err != nil {
-			return errors.Wrapf(err, "couldn't parse mix version value %q", b.Mixver)
-		}
-		mixVerUint = uint32(parsed)
+	mixVerUint, err = parseUint32(b.Mixver)
+	if err != nil {
+		return errors.Wrapf(err, "couldn't parse mix version")
 	}
 
 	// Ensure the format dir exists.
@@ -557,12 +554,30 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string, 
 
 	// Create update metadata for the mix.
 	timer.Start("create update")
-	updatecmd := exec.Command(prefixflag+"swupd_create_update", "-S", b.Statedir, "--minversion", strconv.Itoa(minVersion), "-F", b.Format, "--osversion", b.Mixver)
-	updatecmd.Stdout = os.Stdout
-	updatecmd.Stderr = os.Stderr
-	err = updatecmd.Run()
-	if err != nil {
-		return errors.Wrapf(err, "failed to create update metadata")
+	if UseNewSwupdServer {
+		// TODO: move this to parsing configuration / parameter time.
+		if minVersion < 0 || minVersion > math.MaxUint32 {
+			return errors.Errorf("minVersion %d is out of range", minVersion)
+		}
+		// TODO: move this to parsing configuration / parameter time.
+		// TODO: should this be uint64?
+		var formatUint uint32
+		formatUint, err = parseUint32(b.Format)
+		if err != nil {
+			return errors.Errorf("invalid format")
+		}
+		_, err = swupd.CreateManifests(mixVerUint, uint32(minVersion), uint(formatUint), b.Statedir)
+		if err != nil {
+			return errors.Wrapf(err, "failed to create update metadata")
+		}
+	} else {
+		updatecmd := exec.Command(prefixflag+"swupd_create_update", "-S", b.Statedir, "--minversion", strconv.Itoa(minVersion), "-F", b.Format, "--osversion", b.Mixver)
+		updatecmd.Stdout = os.Stdout
+		updatecmd.Stderr = os.Stderr
+		err = updatecmd.Run()
+		if err != nil {
+			return errors.Wrapf(err, "failed to create update metadata")
+		}
 	}
 	timer.Stop()
 
@@ -817,4 +832,12 @@ func checkRPM(path string) error {
 		return errors.Errorf("file is not a RPM: %s", string(output))
 	}
 	return nil
+}
+
+func parseUint32(s string) (uint32, error) {
+	parsed, err := strconv.ParseUint(s, 10, 32)
+	if err != nil {
+		return 0, errors.Wrapf(err, "error parsing value %q", s)
+	}
+	return uint32(parsed), nil
 }
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -530,14 +530,23 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
 
 // BuildUpdate will produce an update consumable by the swupd client
 func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string, skipSigning bool, publish bool, keepChroots bool) error {
-	var timer stopWatch
-	defer timer.WriteSummary(os.Stdout)
+	var err error
+
+	if minVersion < 0 || minVersion > math.MaxUint32 {
+		return errors.Errorf("minVersion %d is out of range", minVersion)
+	}
 
 	if format != "" {
 		b.Format = format
 	}
+	// TODO: move this to parsing configuration / parameter time.
+	// TODO: should this be uint64?
+	var formatUint uint32
+	formatUint, err = parseUint32(b.Format)
+	if err != nil {
+		return errors.Errorf("invalid format")
+	}
 
-	var err error
 	// TODO: move this to parsing configuration time.
 	var mixVerUint uint32
 	mixVerUint, err = parseUint32(b.Mixver)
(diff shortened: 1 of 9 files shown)


builder/builder.go
```diff
@@ -522,137 +529,156 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
 	if err != nil {
-		helpers.PrintError(err)
-		os.Exit(1)
+		return errors.Wrapf(err, "couldn't create the format directory")
 	}
 
-	err = ioutil.WriteFile(b.Statedir+"/image/LAST_VER", []byte(b.Mixver), 0644)
+	// Create update metadata for the mix.
+	timer.Start("create update")
+	updatecmd := exec.Command(prefixflag+"swupd_create_update", "-S", b.Statedir, "--minversion", strconv.Itoa(minVersion), "-F", b.Format, "--osversion", b.Mixver)
```
matthewrsj: Why not also use the option to run the new `CreateManifests()` function as well?

cmarcelo: I'm working on those :-)

matthewrsj:
@cmarcelo once #82 and #84 are merged you can add CreateManifests to this PR.

cmarcelo:
@matthewrsj zero packs are in. I'll call CreateManifests tomorrow, but think this PR is already good to go.

builder/builder.go
```diff
@@ -572,36 +596,112 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string,
 		fmt.Println("Signed Manifest.MoM")
 	}
 
+	fullChrootDir := filepath.Join(b.Statedir, "image", b.Mixver, "full")
+	if _, err = os.Stat(fullChrootDir); err != nil {
+		return errors.Wrapf(err, "couldn't access the full chroot")
+	}
+
+	outputDir := filepath.Join(b.Statedir, "www")
```
matthewrsj: We are ignoring the output directory specified in server.ini again. I don't think this is a big deal right now, but after we have the swupd code fully in mixer we should consolidate configuration files and access *one* configuration in *one* place.

builder/builder.go
```diff
@@ -572,36 +596,112 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string,
...
+	timer.Start("create zero packs")
+	if UseNewSwupdServer {
+		var bundles map[string]*swupd.BundleToPack
+		bundles, err = swupd.FindBundlesToPack(0, mixVerUint, b.Statedir)
+		if err != nil {
+			return errors.Wrapf(err, "couldn't find bundles to pack")
+		}
+
+		// TODO: Use goroutines.
+		for _, b := range bundles {
```
matthewrsj: Are you shadowing the `b *Builder` receiver on purpose?

builder/builder.go
```diff
@@ -720,3 +733,50 @@ func checkRPM(path string) error {
 	}
 	return nil
 }
+
+// stopWatch keeps track of a sequence of durations. Use Start and Stop to mark the sections, then
+// write the final result with WriteSummary.
+type stopWatch struct {
```
matthewrsj: These things seem useful enough to put in their own file or in helpers.go to me.

builder/builder.go
cmarcelo: Mixer itself ignores that configuration, so I suggest we handle this consolidation in separate (possibly after we get things working).

builder/builder.go
cmarcelo: Will move to its own file.

builder/builder.go
cmarcelo: Not really. Will fix this.

cmarcelo:
> I feel like all this error checking should be happening in the swupd library functions themself so that all the caller has to do is define the filepaths and return an error if the library function fails.

This was somehow "by design" in this PR. I wanted to get the code in BuildUpdate actually working to have a better vision of what extra functions should be in swupd library.

Would be OK to address this in a separate commit after this PR?

builder/builder.go
matthewrsj: +1

builder/builder.go
```diff
@@ -524,25 +532,56 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
-	err = updatecmd.Run()
-	if err != nil {
-		return errors.Wrapf(err, "failed to create update metadata")
+	timer.Start("create update")
+	if UseNewSwupdServer {
+		// TODO: move this to parsing configuration / parameter time.
+		if minVersion < 0 || minVersion > math.MaxUint32 {
+			return errors.Errorf("minVersion %d is out of range", minVersion)
+		}
+		// TODO: support format strings in CreateManifests. See
```
matthewrsj: We agree this isn't necessary, correct?

cmarcelo: Correct. I haven't pushed the updated version yet.

cmarcelo:
@matthewrsj updated version.

There is still something wrong with the packs (verify install is failing), but is something to be fixed independent of this PR.

builder/builder.go
```diff
@@ -655,14 +656,36 @@ func (b *Builder) buildUpdateWithNewSwupd(timer *stopWatch, mixVersion uint32, m
...
+	}
+
+	thisVersionDir := filepath.Join(outputDir, fmt.Sprint(mixVersion))
+	fmt.Println("Compressing Manifest.MoM")
+	momF := filepath.Join(thisVersionDir, "Manifest.MoM")
+	if skipSigning {
+		err = createCompressedArchive(momF+".tar", momF)
 	} else {
-		fmt.Println("SKIPPING manifest signing.")
+		err = createCompressedArchive(momF+".tar", momF, momF+".sig")
```
cmarcelo: @tmarcu @matthewrsj after writing this, I've noticed that when using Mixer we don't do this. But the Manifest.MoM.tar in Clear Linux do have the sig included. Should I take it off (not include the signature in the tar)?

cmarcelo:
> Would be OK to address this in a separate commit after this PR?

The patches on top reduced the work done by the caller, in part by reusing the MoM information we now have at hand.

builder/builder.go
matthewrsj: if it is done elsewhere we should consider doing it here to utilize the very helpful `createCompressedArchive`

builder/builder.go
```diff
@@ -639,6 +639,157 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string,
 	return nil
 }
 
+func (b *Builder) buildUpdateWithNewSwupd(timer *stopWatch, mixVersion uint32, minVersion uint32, format uint32, skipSigning bool) error {
+	var err error
+
+	outputDir := filepath.Join(b.Statedir, "www")
+	if _, err = os.Stat(outputDir); err != nil {
+		return errors.Wrapf(err, "couldn't access the swupd output dir (www)")
```
matthewrsj: This isn't strictly necessary since `CreateManifests` makes the `outputDir` if it doesn't exist since it might not exist for the very first update built ever.

builder/builder.go
```diff
@@ -639,6 +639,157 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string,
...
+	// TODO: CreateFullfiles should return us feedback on what was
+	// done so we can report here.
+	err = swupd.CreateFullfiles(mom.FullManifest, fullChrootDir, fullfilesDir)
+	if err != nil {
+		return err
+	}
+	timer.Stop()
+
+	timer.Start("CREATE ZERO PACKS")
+	for _, bundle := range mom.Files {
```
matthewrsj: Is this worth being a function, maybe `CreatePacks` or `CreateAllPacks`, in the swupd library?

builder/builder.go
```diff
@@ -720,3 +871,74 @@ func checkRPM(path string) error {
 	}
 	return nil
 }
+
+func parseUint32(s string) (uint32, error) {
```
matthewrsj: helpers.go ? The swupd library itself could use this function actually.

builder/builder.go
```diff
@@ -720,3 +871,74 @@ func checkRPM(path string) error {
...
+	parsed, err := strconv.ParseUint(s, 10, 32)
+	if err != nil {
+		return 0, errors.Wrapf(err, "error parsing value %q", s)
+	}
+	return uint32(parsed), nil
+}
+
+func createCompressedArchive(dst string, srcs ...string) (err error) {
```
matthewrsj: Any way to re-use what you wrote for fullfiles in the swupd library?

matthewrsj:
LGTM with just a few suggestions/questions. This'll need a review from @tmarcu though because it changes some mixer core code.

builder/builder.go
cmarcelo: Makes sense. Removed the check.

builder/builder.go
cmarcelo: We do take advantage of printing output during the loop in this code, so moving to the library we will have to decide how to handle that (e.g. pass a io.Writer, etc). I rather do this when we explore using goroutines, as one might influence another (having a channel to get the intermediate information).

builder/builder.go
cmarcelo: You noticed below the reuse of ExternalWriter :-). The other bits are more specific for each use case (e.g. tradeoff whether we compress everything and write vs. streaming, what file types to support), so not worth trying to consolidate now.

builder/builder.go
cmarcelo: I'll skip doing this now since the PR is already a long one.

cmarcelo:
> There is still something wrong with the packs (verify install is failing), but is something to be fixed independent of this PR.

Fixed :-). It was a trivial thing.

builder/builder.go
```diff
@@ -852,3 +875,66 @@ func parseUint32(s string) (uint32, error) {
...
+		}
+	}
+
+	archive := &bytes.Buffer{}
+
+	xw, err := swupd.NewExternalWriter(archive, "xz")
+	if err != nil {
+		return err
+	}
+	defer func() {
```
icarus-sparry: I find the error and close handling here a bit forced. Perhaps moving  lines 914 to 933 into its own function and just doing

    xw, err := swupd.NewExternalWriter()
    if err != nil { return err }
    err = some_new_function(xw, headers, srcs)
    cerr := xw.Close()
    if err { return err }
    if cerr { err = cerr /* because of magic err handling! */ ; return err }
    return ioutil.WriteFile()

would make me feel better?
Perhaps move all but 880 to 884 into createCompressedArchiveInternal, and remove the magic handling of "err" as well?

     func createCompressedArchive(dst string, srcs ...string) (err error) {
          err := createCompressedArchiveInternal(dst, srcs...)
          return errors.Wrapf(err, "couldn't create compressed archive %s", dst)
     }

builder/builder.go
```diff
@@ -852,3 +875,66 @@ func parseUint32(s string) (uint32, error) {
...
+	}()
+
+	tw := tar.NewWriter(xw)
+	for i, hdr := range headers {
+		err = tw.WriteHeader(hdr)
+		if err != nil {
+			return err
+		}
+		var srcData []byte
+		srcData, err = ioutil.ReadFile(srcs[i])
```
icarus-sparry: Probably ought to think if we want to read large amounts of data into memory, just to pass it on to tar or if it makes sense to read it in chunks, e.g. 256KB. Something for the future, not now.

swupd/fullfiles.go
```diff
@@ -49,6 +49,12 @@ func CreateFullfiles(m *Manifest, chrootDir, outputDir string) error {
 			// if the file could be compressed.
 			output := filepath.Join(outputDir, name+".tar")
 
+			// Don't regenerate if file exists.
```
tmarcu: Only if the file exists in a location that is user editable, or could have been a failed download/creation.

cmarcelo: Could you elaborate a bit on what cases we should skip or not?

To identify a failed creation, we need to verify the hash of the file. So we should just skip if the hash matches, and try to rewrite if it doesn't?

builder/builder.go
cmarcelo: This function job was to archive two small regular files (Manifest.MoM and Manifest.MoM.sig), so I opted for the tradeoff of doing all in memory so I will only write a file when it is good (makes error handling easier as I'll never need to care about invalid file being written).

I'll clarify that scope in the function comment.

swupd/cmd/create-pack/main.go
```diff
@@ -74,153 +72,74 @@ func main() {
 		}
 	}
 
-	if *outputDir == "" {
-		tempDir, err := ioutil.TempDir(".", "packs-")
+	var err error
+	var bundles map[string]*swupd.BundleToPack
+
+	if *allBundles {
```
tmarcu: It would be really nice to have this moved into the regular Mixer CLI so we're not carrying 2 cli implementations in the same project. Out of scope for this PR, but that should be updated soon IMO.

cmarcelo: These programs let me test the functionality without having to do all the mixer configuration and steps. I'll probably nuke them once I set up better automated testing for the functionality.

Agree out of scope of this PR to remove them.

builder/builder.go
tmarcu: +1 kill the serveri.ini

builder/builder.go
cmarcelo: Reorganize that function to make it clearer. Split off tar-specific code into a helper function. 

Removed the defer that was wrapping error, and explicit wrap the errors. (I still prefer wrapping using defer, as it is less error prone, but not really that important).

builder/builder.go
```diff
@@ -557,12 +554,30 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string,
...
-	err = updatecmd.Run()
-	if err != nil {
-		return errors.Wrapf(err, "failed to create update metadata")
+	if UseNewSwupdServer {
+		// TODO: move this to parsing configuration / parameter time.
+		if minVersion < 0 || minVersion > math.MaxUint32 {
+			return errors.Errorf("minVersion %d is out of range", minVersion)
+		}
+		// TODO: move this to parsing configuration / parameter time.
+		// TODO: should this be uint64?
```
tmarcu: I don't think we'll max the format in our lifetime if it's a uint64.

cmarcelo: Format is "uint" with size not specified, the TODO was more to remember if we want to set the size of the number that holds format.

builder/builder.go
cmarcelo: @icarus-sparry could you please take another look?

builder/builder.go
```diff
@@ -850,3 +874,72 @@ func parseUint32(s string) (uint32, error) {
...
 	return uint32(parsed), nil
 }
+
+// createCompressedArchive will use tar and xz to create a compressed
+// file. It does not stream the sources contents, doing all the work
+// in memory before writing the destination file.
+func createCompressedArchive(dst string, srcs ...string) error {
+	// Helper function to avoid repeating the destination
+	// information in every error we make.
+	wrapError := func(err error) error {
```
icarus-sparry: This is still missing my point - sorry I have not been clearer.

If you want to write it like this then it should say something like

    func createCompressedArchive(dst string, srcs ...string) error {
        err := func (dst string, srcs ...string) error { // anonymous internal function
               var err error
               archive := &bytes.Buffer{}
               xw, err := swupd.NewExternalWriter(archive, xz)
               if err := nil {
                      return err   // just return err from the anonymous function
               }
               // lots of omitted lines
               return nil
            }(dst, srcs...)
         return errors.Wrapf(err, "couldn't create compressed archive %s", dst)
    }

cmarcelo: @icarus-sparry  Could you take another look?

builder/builder.go
```diff
@@ -850,3 +874,70 @@ func parseUint32(s string) (uint32, error) {
 	}
 	return uint32(parsed), nil
 }
+
+// createCompressedArchive will use tar and xz to create a compressed
+// file. It does not stream the sources contents, doing all the work
+// in memory before writing the destination file.
+func createCompressedArchive(dst string, srcs ...string) error {
+	err := createCompressedArchiveInternal(dst, srcs...)
```
tmarcu: I like this implementation.

builder/builder.go
```diff
@@ -850,3 +874,70 @@ func parseUint32(s string) (uint32, error) {
...
+		return err
+	}
+
+	err = xw.Close()
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(dst, archive.Bytes(), 0644)
+	if err != nil {
```
matthewrsj: just return err here and remove last 5 lines?

cmarcelo: Fixed.

builder/builder.go
```diff
@@ -524,26 +532,50 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
 
 // BuildUpdate will produce an update consumable by the swupd client
 func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string, skipSigning bool, publish bool, keepChroots bool) error {
+	var err error
+
+	if minVersion < 0 || minVersion > math.MaxUint32 {
```
icarus-sparry: If we had a type for "version", we could specify that minVersion was of this type and remove this check.

builder/builder.go
```diff
@@ -639,6 +640,151 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string,
...
+		zeropackArgs = append(zeropackArgs, "--repodir", prefixflag)
+	}
+	zeropackcmd := exec.Command("mixer-pack-maker.sh", zeropackArgs...)
+	zeropackcmd.Stdout = os.Stdout
+	zeropackcmd.Stderr = os.Stderr
+	err = zeropackcmd.Run()
+	if err != nil {
+		return errors.Wrapf(err, "couldn't create zero packs")
+	}
+	timer.Stop()
```
icarus-sparry: Move the timer.Stop before the error handling?

builder/builder.go
```diff
@@ -720,3 +866,73 @@ func checkRPM(path string) error {
...
+	if err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(dst, archive.Bytes(), 0644)
+}
+
+func archiveFiles(w io.Writer, srcs []string) error {
+	tw := tar.NewWriter(w)
+	for _, src := range srcs {
```
icarus-sparry: Probably need either goto or defer to close the tw in the event of error.

cmarcelo: We don't need to Close a tar.Writer.

cmarcelo: ...In case of failure.

builder/builder.go
cmarcelo: Agreed. Either that or simply using Uint32.

I had to pick my battles since this would cause a lot of changes outside the scope of what I was doing.

builder/builder.go
icarus-sparry: OK.

PR #80, opened 2018-01-06
Add test helpers to check for strings in manifests

matthewrsj:

### DIFF
--- a/swupd/create_manifests_test.go
+++ b/swupd/create_manifests_test.go
@@ -1,7 +1,6 @@
 package swupd
 
 import (
-	"bytes"
 	"io/ioutil"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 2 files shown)


swupd/create_manifests_test.go
```diff
@@ -49,81 +48,46 @@ func TestInitBuildDirs(t *testing.T) {
...
-	// do not use the fileContains helper for this because we are checking a lot
-	// of strings in the same file
-	b, err := ioutil.ReadFile(filepath.Join(testDir, "www/10/Manifest.test-bundle"))
-	if err != nil {
-		t.Fatal("could not read test file for checking")
+	nExpSubs := []string{
+		"\t0\t/foo",
+		".d..\t",
 	}
+	checkManifestContains(t, testDir, "10", "test-bundle", nExpSubs, false)
```
cmarcelo: It will be better for our sanity if we had a pair of functions: checkManifestContains and checkManifestNotContains (not sure this is proper english). I know what you mean here with the false in the end of the call, but will forget soon.

(Other solutions that make explicit Contains/NotContains distinction are also valid)

swupd/helpers_test.go
```diff
@@ -218,3 +200,37 @@ func mustCreateManifests(t *testing.T, ver uint32, minVer bool, format uint, tes
...
+			if expRes {
+				t.Errorf("%s/Manifest.%s did not contain expected '%s'", ver, name, sub)
+			} else {
+				t.Errorf("%s/Manifest.%s contained unexpected '%s'", ver, name, sub)
+			}
+		}
+	}
+}
+
+func checkManifestContainsRes(t *testing.T, testDir, ver, name string, res []*regexp.Regexp) {
```
cmarcelo: Optional: checkManifestMatchREs? checkManifestMatches?

swupd/helpers_test.go
```diff
@@ -218,3 +200,37 @@ func mustCreateManifests(t *testing.T, ver uint32, minVer bool, format uint, tes
 		t.Fatal(err)
 	}
 }
+
+func checkManifestContains(t *testing.T, testDir, ver, name string, subs []string, expRes bool) {
```
cmarcelo: Optional: assuming you drop the bool, you can use varargs to allow both single string and list of strings.

    func checkManifestContains(..., subs ...string) {
        // can use subs as a []string, so implementation don't change.
    }

    // Then we can call:
    checkManifestContains(t, dir, "10", "MoM", "alalallaa")
    // or
    mySubs := []string{...}
    checkManifestContains(t, dir, "10", "MoM", mySubs...)

swupd/helpers_test.go
```diff
@@ -218,3 +200,37 @@ func mustCreateManifests(t *testing.T, ver uint32, minVer bool, format uint, tes
...
+func checkManifestContainsRes(t *testing.T, testDir, ver, name string, res []*regexp.Regexp) {
+	manFpath := filepath.Join(testDir, "www", ver, "Manifest."+name)
+	b, err := ioutil.ReadFile(manFpath)
+	if err != nil {
+		t.Error(err)
+		return
+	}
+
+	for _, re := range res {
+		if string(re.Find(b)) == "" {
```
icarus-sparry: `Find` is a bit heavyweight, and then having to convert the array of bytes to a string to compare to the empty string seems to be adding insult to injury!

> if re.Match(b) ==false {

swupd/helpers_test.go
icarus-sparry: As far as the interface is concerned, I am not sure there is a lot of value in being able to pass in multiple expressions. As @cmarcelo points out for checkManifestContains, using a varargs approach probably makes sense here as well if we do want multiple expressions.

swupd/helpers_test.go
icarus-sparry: Even if we don't drop the bool (which I strongly support), we could move the `expRes` before `subs`.

matthewrsj:
Thanks @cmarcelo and @icarus-sparry, changes incorporated. I like having the option to pass multiple substrings to match, so I went with the varargs suggestion.

swupd/helpers_test.go
```diff
@@ -30,6 +30,11 @@ func mustDirExistsWithPerm(t *testing.T, path string, perm os.FileMode) {
 }
 
 func mustSetupTestDir(t *testing.T, testName string) string {
+	oldDirs, _ := filepath.Glob("./testdata/cmtest-" + testName + "*")
```
icarus-sparry: This works fine as long as no testName is a prefix of another, or we don't run the tests in parallel.

I think it would help to have an extra character in the directory names, for example `.`, and change this line to

     oldDirs, _ := filepath.Glob("./testdata/cmtest-" + testName + ".*")

swupd/helpers_test.go
```diff
@@ -30,6 +30,11 @@ func mustDirExistsWithPerm(t *testing.T, path string, perm os.FileMode) {
...
+	for _, d := range oldDirs {
+		removeAllIgnoreErr(d)
+	}
+
 	testDir, err := ioutil.TempDir("./testdata", "cmtest-"+testName)
```
icarus-sparry: and change this to

    testDir, err := ioutil.TempDir("./testdata", "cmtest-"+testName +".")

to avoid the issue (at least until people start adding their of periods to test names).
Obviously one can use a different character than `.`, any of `!@#%+` are reasonable choices, or even adding `@@`

matthewrsj: Good suggestion, thank you.

PR #81, opened 2018-01-08
 Maximize file version in Manifest.full

matthewrsj:
When a file exists in one bundle, then is added to another bundle as
well, Manifest.full should report the file version as the most recent
changed version. Add test for this case as well.

This PR is based on #80. Once that PR is merged this will contain one commit containing the Manifest.full maximize changes only.

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -158,10 +158,6 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 		// detect modifier flag for all files in the manifest
 		// must happen after finding newDeleted files to catch ghosted files.
 		bundle.applyHeuristics()
-		// sort manifest by version (then by filename)
-		// this must be done after subtractManifests has been done for all manifests
-		// because subtractManifests sorts the file lists by filename alone
-		bundle.sortFilesVersionName()
 		// Assign final FileCount based on the files that made it this far
 		bundle.Header.FileCount = uint32(len(bundle.Files))
 		// If we made it this far, this bundle has a change and should be written
@@ -258,6 +254,14 @@ func CreateManifests(version uint32, minVersion bool, format uint, statedir stri
 
 	// write manifests then add them to the MoM
 	for _, bMan := range newManifests {
+		// full is just another manifest, grab it from here and maximize versions
+		if bMan.Name == "full" {
+			maximizeFull(bMan, newManifests)
+		}
+
+		// sort by version then by filename, previously to this sort these bundles
+		// were sorted by file name only to make processing easier
+		bMan.sortFilesVersionName()
 		manPath := filepath.Join(verOutput, "Manifest."+bMan.Name)
 		if err = bMan.WriteManifestFile(manPath); err != nil {
 			return err
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -531,3 +531,38 @@ func getManifestVerFromMoM(mom *Manifest, b *Manifest) uint32 {
 
 	return 0
 }
+
+// maxFullFromManifest maximizes all file entries in mf that are also in bundle
+func maxFullFromManifest(mf *Manifest, bundle *Manifest) {
+	mf.sortFilesName()
+	bundle.sortFilesName()
+	i := 0
+	j := 0
+
+	for i < len(mf.Files) && j < len(bundle.Files) {
+		ff := mf.Files[i]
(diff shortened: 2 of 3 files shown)


swupd/manifest.go
```diff
@@ -531,3 +531,38 @@ func getManifestVerFromMoM(mom *Manifest, b *Manifest) uint32 {
 
 	return 0
 }
+
+// maxFullFromManifest maximizes all file entries in mf that are also in bundle
+func maxFullFromManifest(mf *Manifest, bundle *Manifest) {
+	mf.sortFilesName()
```
icarus-sparry: Should maxFullFromManifest sort the files in the manifests as a side effect, or should it make a copy of the array of file pointers and sort that copy? Obviously we want the mf.Files[x].Version fields updated, but not clear we want the order of the mf.Files and bundle.Files altered.

 Should maxFullFromMinifest be a method on mf?

swupd/manifest.go
```diff
@@ -531,3 +531,38 @@ func getManifestVerFromMoM(mom *Manifest, b *Manifest) uint32 {
...
+			i++
+		} else {
+			// check next filename in bundle
+			j++
+		}
+	}
+}
+
+// maximizeFull maximizes all file entries in mf for each bundle in bundles
+func maximizeFull(mf *Manifest, bundles []*Manifest) {
```
icarus-sparry: Should maximizeFull be a method on mf?

swupd/manifest.go
matthewrsj: > Should maxFullFromManifest sort the files in the manifests as a side effect
If you look in #82 I have a reason for wanting to actually sort the manifests. This may or may not be a good idea because of side effects, but it allows us to skip sorting if we walk the manifest several times.

> Should maxFullFromManifest be a method on mf?
I avoided that because it is very specifically meant to be used on the Manifest.full and no other manifest.

swupd/manifest.go
matthewrsj: Same comment as above.

swupd/manifest.go
```diff
@@ -531,3 +531,38 @@ func getManifestVerFromMoM(mom *Manifest, b *Manifest) uint32 {
...
+func maxFullFromManifest(mf *Manifest, bundle *Manifest) {
+	mf.sortFilesName()
+	bundle.sortFilesName()
+	i := 0
+	j := 0
+
+	for i < len(mf.Files) && j < len(bundle.Files) {
+		if mf.Files[i].Name == bundle.Files[j].Name {
+			// files match, maximize versions if appropriate
+			ff := mf.Files[i]
```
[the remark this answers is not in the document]
matthewrsj: yes, I'll update that.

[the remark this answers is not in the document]
matthewrsj: updated

swupd/manifest.go
icarus-sparry: Fair enough. Just noting that #82 didn't exist when I wrote my review, and it doesn't prevent there being a separate sort step. I think that a few extra comments explaining the reasoning would help. The array being sorted makes concurrent access to the manifest more fragile, even if it is only read access.

OK also on the reasoning behind not making it a method.

swupd/create_manifests.go
```diff
@@ -258,6 +254,14 @@ func CreateManifests(version uint32, minVersion bool, format uint, statedir stri
 
 	// write manifests then add them to the MoM
 	for _, bMan := range newManifests {
+		// full is just another manifest, grab it from here and maximize versions
+		if bMan.Name == "full" {
+			maximizeFull(bMan, newManifests)
```
cmarcelo: I think this maximize step should happen before this loop, but I might be missing something.

The potential problem I'm seeing is newManifests contains "a", "full" and "b".

- "a" gets sorted by version then name
- "full" causes resort of "a" and "b" to be name
- "full" itself is sorted by version then name
- "b" is sorted by version then name

At the end of the loop, "a" sorting is incorrect.

swupd/manifest.go
cmarcelo: Full manifest is being sorted multiple for each manifest.

One sort in maximizeFull should be enough. And document this expectation (or inline this function).

matthewrsj:
@cmarcelo @icarus-sparry since many of the issues brought up in this review are/will be addressed in #82 I will just close this PR and make changes in that PR. #82 already has the commit from this PR so no change there.

PR #82, opened 2018-01-08
Swupd performance + maximize full

matthewrsj:
Maximize Manifest.full
* When a file exists in one bundle, then is added to another bundle as
well, Manifest.full should report the file version as the most recent
changed version. Add test for this case as well.

Make various performance fixes for manifest creation.
* Added an appendFile helper function to unset the Sorted field.
* Optimize linkPeersAndChange 
  Instead of an O(N^2) operation (N^3 counting the calling loop) walk each
manifest file list at the same time for a O(N) operation (N^2 counting
the calling loop). Fold filesAdded() and newDeleted() into
linkPeersAndChange so they can take advantage of the double list walk.

A couple minor test fixes
* Fix misnamed test from "format" to "ghosted"
* Change test error to fatal
  This test should be fatal so the f.DeltaPeer.Name is not checked when
f.DeltaPeer is nil.

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -148,7 +148,6 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 		oldMPath := filepath.Join(c.outputDir, fmt.Sprint(ver), "Manifest."+bundle.Name)
 		oldM := getOldManifest(oldMPath)
 		changedIncludes := compareIncludes(bundle, oldM)
-		bundle.sortFilesName()
 		oldM.sortFilesName()
 		changedFiles, added, deleted := bundle.linkPeersAndChange(oldM)
 		// if nothing changed, skip
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -158,10 +158,6 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 		// detect modifier flag for all files in the manifest
 		// must happen after finding newDeleted files to catch ghosted files.
 		bundle.applyHeuristics()
-		// sort manifest by version (then by filename)
-		// this must be done after subtractManifests has been done for all manifests
-		// because subtractManifests sorts the file lists by filename alone
-		bundle.sortFilesVersionName()
 		// Assign final FileCount based on the files that made it this far
 		bundle.Header.FileCount = uint32(len(bundle.Files))
 		// If we made it this far, this bundle has a change and should be written
@@ -258,6 +254,14 @@ func CreateManifests(version uint32, minVersion bool, format uint, statedir stri
 
 	// write manifests then add them to the MoM
 	for _, bMan := range newManifests {
+		// full is just another manifest, grab it from here and maximize versions
+		if bMan.Name == "full" {
+			maximizeFull(bMan, newManifests)
+		}
+
+		// sort by version then by filename, previously to this sort these bundles
+		// were sorted by file name only to make processing easier
+		bMan.sortFilesVersionName()
 		manPath := filepath.Join(verOutput, "Manifest."+bMan.Name)
 		if err = bMan.WriteManifestFile(manPath); err != nil {
 			return err
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -112,13 +112,6 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 			bundle.removeDebuginfo(c.debuginfo)
 		}
 
-		oldMPath := filepath.Join(c.outputDir, fmt.Sprint(ui.lastVersion), "Manifest."+bundle.Name)
-		oldM := getOldManifest(oldMPath)
-		// add old deleted files if the format has incremented
-		if ui.oldFormat == ui.format {
-			bundle.addDeleted(oldM)
-		}
-
 		bundle.sortFilesName()
 		tmpManifests = append(tmpManifests, bundle)
 	}
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -147,9 +147,7 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 		oldMPath := filepath.Join(c.outputDir, fmt.Sprint(ver), "Manifest."+bundle.Name)
 		oldM := getOldManifest(oldMPath)
 		changedIncludes := compareIncludes(bundle, oldM)
-		changedFiles := bundle.linkPeersAndChange(oldM)
-		added := bundle.filesAdded(oldM)
-		deleted := bundle.newDeleted(oldM)
+		changedFiles, added, deleted := bundle.linkPeersAndChange(oldM)
 		// if nothing changed, skip
 		if changedFiles == 0 && added == 0 && deleted == 0 && !changedIncludes {
 			continue
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -76,6 +76,7 @@ func getOldManifest(path string) *Manifest {
 }
 
 func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
+	var newFull *Manifest
 	tmpManifests := []*Manifest{}
 	// first loop sets up initial bundle manifests and adds files to them
 	for _, bundleName := range ui.bundles {
@@ -87,6 +88,12 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 				TimeStamp: ui.timeStamp,
 			},
 			Name: bundleName,
+		}
+
+		if bundleName == "full" {
+			// track full manifest so we can do extra processing later
+			// (maximizeFull)
+			newFull = bundle
 		}
 
 		bundleChroot := filepath.Join(c.imageBase, fmt.Sprint(ui.version), bundle.Name)
@@ -112,13 +119,14 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 			bundle.addDeleted(oldM)
 		}
 
+		bundle.sortFilesName()
 		tmpManifests = append(tmpManifests, bundle)
 	}
 
 	// second loop reads includes and then subtracts manifests using the file lists
 	// populated in the last step
 	for _, bundle := range tmpManifests {
-		if bundle.Name != "os-core" && bundle.Name != "full" {
+		if bundle.Name != "os-core" && bundle != newFull {
 			// read in bundle includes
 			if err := bundle.readIncludes(tmpManifests, c); err != nil {
 				return nil, err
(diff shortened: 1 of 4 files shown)


swupd/manifest.go
```diff
@@ -386,59 +386,100 @@ func (m *Manifest) prependFile(f *File) {
...
+			// it is added.
+			nf.Version = m.Header.Version
+			added++
+			// look at next file in current manifest
+			i++
+		} else {
+			// if the file exists in the old manifest and does not *yet* exist
+			// in the new manifest, it was deleted. Set up the deleted file
+			// and prepend it to the files list via newDeleted()
+			if of.Status == statusDeleted || of.Status == statusGhosted {
```
icarus-sparry: I am finding I am writing this quite a bit as well. Perhaps we might think about splitting this into two values, like the old is_deleted and is_ghosted bits?

matthewrsj: Yes that might be worth doing. It is kind of weird because with the current manifest format these flags occupy the same position (`.d/g..`) which is why they were combined into their own `enum`.

What if we instead check if `file.Status == statusUnset`? Not very extensible if we add more status flags, but much simpler.

cmarcelo: What about having a method in fstatus, so we can say things like: of.Status.IsGone()? (name to be defined)

swupd/manifest.go
```diff
@@ -317,9 +331,15 @@ func (m *Manifest) sortFilesName() {
...
 		return m.DeletedFiles[i].Name < m.DeletedFiles[j].Name
 	})
+	m.Sorted = sortedByName
 }
 
 func (m *Manifest) sortFilesVersionName() {
+	// do not re-sort
+	if m.Sorted == sortedByVersionName {
+		return
+	}
```
cmarcelo: I don't like this solution because it will hide from us when the expensive operation is happening, and led us to write code that is going to be more care to analyze. Might sound trivial know to understand, but put a few months of patches on top, and things become difficult.

Given we have a good idea of when things need to be sorted, _my vote is to make sorting more explicit_.

- We read all the bundle manifest files
- Sort them by name
- Run other algorithms that depend on sorted manifest
 -- Link Peers, Find Gone, Maximize Full
 -- Those that change the list, should fix it before return
- In the end, sort them by version then name

Easy fix for algorithms that change the list is to call Sort before they return, but this approach let us try other ways to keep the invariant. E.g. Find Gone can first store the deleted files list, and then create a new one (the right size) merging both Files and DeletedFiles, then just doing m.Files = mergedFiles.

That said, if you still want to pursue the approach in this patch, make a new type that is a struct containing the list of files and the sorted state, and create methods to operate on that. You lose the benefit of using the x[y] syntax, but you properly encapsulate and it will be clearer that we shouldn't be appending to it, but calling m.Files.appendFile().

cmarcelo:
For the two patches that fix tests: if they are tests already in master, +2 from me just push them.

If they are fixes to tests part of the PR, just squash them with the relevant commit.

swupd/manifest.go
matthewrsj: @cmarcelo that sounds like a good idea. I'd rather leave that for another PR though.

matthewrsj:
@cmarcelo @icarus-sparry I've updated this PR. I opted to remove the sorting from the marching functions altogether and put the sorting requirement on the calling functions.

swupd/manifest.go
```diff
@@ -363,59 +363,100 @@ func (m *Manifest) prependFile(f *File) {
...
 			} else {
-				match.Version = of.Version
+				// otherwise the version is the same as the old file
+				nf.Version = of.Version
+			}
+			// advance indices for both file lists since we had a match
+			i++
+			j++
+		} else if nf.Name < of.Name {
+			// if the file does not exist in the old manifest it is a new
```
icarus-sparry: Don't we want at this stage to add the nf to a list of potential new files, and then the rename detection after this loop?

swupd/manifest.go
```diff
@@ -363,59 +363,100 @@ func (m *Manifest) prependFile(f *File) {
...
+			if of.Status == statusDeleted || of.Status == statusGhosted {
+				j++
+				continue
 			}
+			m.newDeleted(of)
+			deleted++
+			// look at next file in old manifest
+			j++
+			// increment the index into m.Files as well since we increased
+			// length by prepending of
```
icarus-sparry: This i++ looks like we are doing the wrong thing!

I think that by the time we get to line 445 we should have a a local slice of each of the following

files which exist in both the old and new manifests
files which didn't exist in the old but do exist in the new
files which did exist in the old and don't exist in the new. (this is essentially the newDeleted stuff, but I don't think we want to alter the attributes yet)
The rename detection code then works on matching things in the second and third lists, if it finds a match then it removes the pair of files from these two lists and adds it to the first.

At this point we can compute "changed", added, and deleted.

We then merge the 3 local slices to form the Files slice of the new manifest.

swupd/manifest.go
matthewrsj: Well, it isn't just new files that could be a renamed-to file if the rename is from an older version, but I suppose the rename status will be copied over from the old manifest in that case. Alternatively we can just run rename detection over the manifest files for files that are last changed at the current version.

swupd/manifest.go
```diff
@@ -386,59 +386,100 @@ func (m *Manifest) prependFile(f *File) {
...
+	oldManifest.sortFilesName()
+
 	changed := 0
-	for _, of := range oldManifest.Files {
-		if match := of.findFileNameInSlice(m.Files); match != nil {
+	added := 0
+	deleted := 0
+	i := 0
+	j := 0
+	for i < len(m.Files) && j < len(oldManifest.Files) {
```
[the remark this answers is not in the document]
matthewrsj: Ok, I get what you mean now. Updated. I'll push an update shortly.

swupd/manifest.go
matthewrsj: I see your point for creating the three different slices so rename detection is faster... but I don't see the point of then merging these three slices at the end when we can just add them *in* the loop like we are already doing. If the three slices contain pointers to the files in the manifest.Files list then we can just update the rename status and the hash in the lists and the master list will have the changes.

swupd/manifest.go
```diff
@@ -342,72 +342,113 @@ func (m *Manifest) sortFilesVersionName() {
...
 }
 
-func (m *Manifest) addDeleted(oldManifest *Manifest) {
-	for _, df := range oldManifest.DeletedFiles {
-		if df.findFileNameInSlice(m.Files) == nil {
-			m.Files = append(m.Files, df)
-			m.DeletedFiles = append(m.DeletedFiles, df)
-		}
-	}
+func (m *Manifest) appendFile(f *File) {
```
cmarcelo: Optional: unless there is a reason to have (for future changes), I would remove this method.

swupd/create_manifests.go
```diff
@@ -147,9 +148,9 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 		oldMPath := filepath.Join(c.outputDir, fmt.Sprint(ver), "Manifest."+bundle.Name)
 		oldM := getOldManifest(oldMPath)
 		changedIncludes := compareIncludes(bundle, oldM)
-		changedFiles := bundle.linkPeersAndChange(oldM)
-		added := bundle.filesAdded(oldM)
-		deleted := bundle.newDeleted(oldM)
+		bundle.sortFilesName()
```
cmarcelo: Aren't those already sorted?

swupd/manifest.go
```diff
@@ -342,72 +342,113 @@ func (m *Manifest) sortFilesVersionName() {
...
+			// file in this manifest. Update the version and record that
+			// it is added.
+			nf.Version = m.Header.Version
+			added = append(added, nf)
+			// look at next file in current manifest
+			i++
+		} else {
+			// if the file exists in the old manifest and does not *yet* exist
+			// in the new manifest, it was deleted. Set up the deleted file
+			// and prepend it to the files list via newDeleted()
```
cmarcelo: Prepend comment here is probably not valid anymore?

swupd/manifest.go
```diff
@@ -342,72 +342,113 @@ func (m *Manifest) sortFilesVersionName() {
...
+	for i < mFilesLen && j < omFilesLen {
+		nf := m.Files[i]
+		of := oldManifest.Files[j]
+		if nf.Name == of.Name {
+			// if it is the same name check if anything about the file has changed.
+			// if something has changed update the version to current version and
+			// record that the file has changed
 			if of.Status == statusDeleted || of.Status == statusGhosted {
+				// don't worry about ghosted or deleted files
+				i++
```
icarus-sparry: I think we do need to worry here, extending added with nf. This allows for

1. Version 10
    1. fred - deleted
    1. fred.new - normal file
1. Version 20
     1. fred - which may be a rename of the previous fred.new

swupd/manifest.go
matthewrsj: +1

swupd/manifest.go
matthewrsj: ewps, yep

swupd/create_manifests.go
matthewrsj: yes

swupd/manifest.go
```diff
@@ -342,72 +342,113 @@ func (m *Manifest) sortFilesVersionName() {
...
+			i++
+		} else {
+			// if the file exists in the old manifest and does not *yet* exist
+			// in the new manifest, it was deleted. Set up the deleted file
+			// and prepend it to the files list via newDeleted()
+			if of.Status == statusDeleted || of.Status == statusGhosted {
+				j++
+				continue
 			}
+			m.newDeleted(of)
```
icarus-sparry: We could just delay the call to m.newDeleted until after we do the rename stuff, and just do it on the removed slice

swupd/manifest.go
```diff
@@ -342,72 +342,113 @@ func (m *Manifest) sortFilesVersionName() {
...
-	for _, of := range oldManifest.Files {
-		if match := of.findFileNameInSlice(m.Files); match != nil {
+// if the file in the oldManifest is not deleted or ghosted.
+// Expects m and oldManifest files lists to be sorted by name only
+func (m *Manifest) linkPeersAndChange(oldManifest *Manifest) (int, int, int) {
+	var unchanged, changed, removed, added []*File
+	i := 0
+	j := 0
+	mFilesLen := len(m.Files)
+	omFilesLen := len(oldManifest.Files)
```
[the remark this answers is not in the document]
matthewrsj: I was putting that in another PR, but I'll add it here so we can stop thrashing this function in so many PRs.

swupd/manifest.go
matthewrsj: ok

PR #85, opened 2018-01-10
[RFC don't merge] swupd: Do not print to stderr from library calls

tmarcu:
If this is indeed a library, then we should remove all printing to stdout/stderr and return proper errors, or don't return anything in cases like this where it's going to continue anyways. Libraries should not print to console at any time.

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -176,10 +176,6 @@ func CreateManifests(version uint32, minVersion bool, format uint, statedir stri
 	var err error
 	var c config
 	c, err = getConfig(statedir)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "Found server.ini, but was unable to read it. "+
-			"Continuing with default configuration\n")
-	}
 
 	if err = initBuildEnv(c); err != nil {
 		return err


swupd/create_manifests.go
```diff
@@ -176,10 +176,6 @@ func CreateManifests(version uint32, minVersion bool, format uint, statedir stri
 	var err error
 	var c config
 	c, err = getConfig(statedir)
```
cmarcelo: Should still return error?

cmarcelo: I know there is a fallback, but an server.ini we can't read might indicate a problem, so worthwhile failing?

matthewrsj: Maybe this should be a real failure then. If a server.ini exists but we can't read the file the user had configuration that we don't know about.

I think we should return the error.

tmarcu: If this could output a bad manifest or update, we should set the error and return it at the end anyway. If not - then throw it out and ignore it. Optionally one could add a nonzero return value to check and log the error in the calling function.

matthewrsj: Looks like we all fell all over ourselves to agree we should return the error.

tmarcu: +1 :) I will update this PR with more changes then, wanted to get the ball rolling and your opinions on it.

cmarcelo:
Ping? :-)

tmarcu:
Haven't had time to look back at this PR yet.

PR #86, opened 2018-01-11
Refactor upstream bundle fetching

kevincwells:
This PR automates the caching of upstream bundle definitions for mixer. 

The user does not need to know how upstream bundle definitions are fetched and stored locally; this is something the tool can and should do automatically for them. All commands that need upstream bundle definitions to function now handle fetching them. Thus, the `mixer bundle get` command is obsolete and has been removed.

Future commands that deal with upstream bundle definitions can easily ensure they exist by calling
```go
getUpstreamBundles(ver string, prune bool) error
```
Sample usage: 
```go
// Fetch upstream bundles if needed
if err := b.getUpstreamBundles(b.Clearver, true); err != nil {
	return err
}
```
The command checks if there is already a cache for the upstream version requested, and returns if it already exists. The second parameter, `prune`, will clean up existing cached upstream bundle definitions before fetching the new ones. The default value should be `true` as part of normal tool clean-up, but can be set to `false` for commands that specifically need to work with multiple versions of upstream bundles simultaneously. Note that `prune` is observed even if the function returns without having to download anything new.

This PR also includes general clean-up and error handling for all of the commands it touches.

This PR has the following side effects:
* The upstream bundles are now considered temporary, sausage-making data. As such, they now live in `.mixer/upstream-bundles/`, and the `.tar.gz` file used to fetch them is cleaned up.
* Fetching upstream bundles no longer relies on an `exec` call out to `tar`. It was never listed as a dependency, but now it doesn't need to be.
* A bug in `mixer bundle add --all` introduced in the CLI port has been fixed. It's left as a separate commit because it is unrelated to the rest of this, but I was already cleaning up that code and wanted to fix it ASAP.
* `mixer init-mix` has been renamed to `mixer init` just for nitpickiness.
* Throughout the effected functions, "clear bundles" and the like has been rephrased as "upstream bundles". This is in keeping with changes to support derivative-of-derivative builds, for which mainline CLR is not the upstream.
* For the time being, with `mixer bundle get` removed, `mixer bundle add` is the only subcommand under the `mixer bundle` command grouping. This won't last long; promise ;)


mixer/cmd/bundles.go
```diff
@@ -40,35 +41,28 @@ var addBundlesCmd = &cobra.Command{
 	Use:   "add [bundle(s)]",
 	Short: "Add clr-bundles to your mix",
 	Long:  `Add clr-bundles to your mix`,
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
```
kevincwells: @cmarcelo - just wanted to point out that this is a good example of a case where we do still want to use `RunE`. Cobra doesn't have a declarative way of saying "This command needs exactly one argument, except when a specific flag is passed", so that kind of logic has to live in the `Run` function. Using `RunE` here lets us pass that error back up to Cobra, and thus get usage data as expected, while still handling internal errors with `fail`.

helpers/helpers.go
```diff
@@ -153,6 +156,70 @@ func GetIncludedBundles(filename string) ([]string, error) {
 	return includes, nil
 }
 
+// UnpackFile unpacks a .tar or .tar.gz/.tgz file to a given directory
+// Should be roughly equivalent to "tar -x[z]f file -C dest"
```
cmarcelo: Document what is the expected behavior when the files being unpacked already exist.

helpers/helpers.go
```diff
@@ -153,6 +156,70 @@ func GetIncludedBundles(filename string) ([]string, error) {
...
+func UnpackFile(file string, dest string) error {
+	fr, err := os.Open(file)
+	defer fr.Close()
```
cmarcelo: This and other defer statements must happen after the error handling. When error is returned, many functions return nil as the other value.

helpers/helpers.go
```diff
@@ -153,6 +156,70 @@ func GetIncludedBundles(filename string) ([]string, error) {
...
+		switch hdr.Typeflag {
+		case tar.TypeDir:
+			if _, err := os.Stat(out); err != nil {
+				if err = os.Mkdir(out, 0777); err != nil {
+					return errors.Wrapf(err, "Error unpacking directory: %s", out)
+				}
+			}
+		case tar.TypeReg:
+			of, err := os.OpenFile(out, os.O_CREATE|os.O_RDWR|os.O_EXCL, os.FileMode(hdr.Mode))
+			defer of.Close()
```
cmarcelo: I don't recommend using defer in this case. Just close "of" directly in the lines below right after the Copy happens, and before checking for the Copy error.


Defer works by adding a the function call to a "cleanup list" in the scope of the function (not the block). This means that we are only closing the files after all of them were opened.

helpers/helpers.go
```diff
@@ -153,6 +156,70 @@ func GetIncludedBundles(filename string) ([]string, error) {
...
+	fr, err := os.Open(file)
+	defer fr.Close()
+	if err != nil {
+		return err
+	}
+
+	var tr *tar.Reader
+
+	// If it's a compressed tarball
+	if strings.HasSuffix(file, ".tar.gz") || strings.HasSuffix(file, ".tgz") {
```
cmarcelo: FYI: somewhere in swupd we have code that can identify from the file itself whether it is gzipped or not, might make this function more robust. I wouldn't worry about this now, but might be useful to pull it later.

mixer/cmd/bundles.go
cmarcelo: OK!

helpers/helpers.go
cmarcelo: Alternative is to pull up the code to a function (either in local scope or toplevel). In this case, since you can resolve it with a single Close, don't think it is worth.

helpers/helpers.go
kevincwells: Really? That'd be pretty cool. I googled around for a while for a good way to do this in go and eventually gave up.

cmarcelo: swupd/packs.go

I would defer it for later, as this doesn't seem the main goal of the PR.

helpers/helpers.go
kevincwells: > This and other defer statements must happen after the error handling.

I thought about this and found examples of both on the internet. Why do you say they need to be after? Wouldn't putting this after the `if err != nil` block mean the file never gets closed in an error state?

>When error is returned, many functions return nil as the other value.

Are you talking about in `AddBundles` where I return `bundleAddCount` instead of 0 (the nil of ints)? I did this because, at the point the error occured, `bundleAddCount` is actually a valid representation of how many bundles were added by the time the error occurred. If there's somewhere else I returned an odd value, I'm not aware of it.

helpers/helpers.go
kevincwells: Ah, that's a really good point about them all being opened at once.

As mentioned above, though, I still don't understand how doing it that way (after two different possible return statements) doesn't leave the possibility of the file hanging open after the function returned.

helpers/helpers.go
cmarcelo: > Wouldn't putting this after the if err != nil block mean the file never gets closed in an error state?

No. The idiom is for functions to return "nil, err". This is true for the functions we are looking at in this PR. If the file wasn't successfully open, you don't need to close it, since there is no file in the first place.

The commit needs to be fixed.

> Are you talking about in AddBundles where I return bundleAddCount instead of 0 (the nil of ints)?

No. It was a bad attempt to explain the above. Sometimes returning "empty" containers is useful for the calling code, but for things like "give me a 'resource'", you usually return nil and an error.

icarus-sparry: Agree 100% with @cmarcelo , the commit does need to be fixed.

It is more usual to say "the zero value of Something", than "the nil of Something". Go has an idiom that you try and make the zero value of Something be something that is usable.

helpers/helpers.go
icarus-sparry: Let me see if I can explain.

    of, err = os.OpenFile(...)

will either

1. Set `of` to `nil` and set `err` to something of type `error` or
1. Set `of` to be a pointer to some kind of data structure and set `err` to `nil`

In the second case there is nothing to close, so you just need to check `err` and if it is not `nil` return it. No `defer` is needed.

In the first case you do need to close `of`. However all you are doing before you want to close it is to call `io.Copy()` so the suggestion is that you write

      of, err := os.OpenFile(.....)
      if err != nil { return err ; }
      _, err = io.Copy()
      cerr := of.Close()
      if err != nil { return errors.Wrapf(err, ...) }
      if cerr != nil { return errors.Wrapf(cerr, "Probably out of disk space"....) }

(which means you also handle any errors you get when closing the output file that are currently missed).

kevincwells:
Pushed update to fix @cmarcelo's feedback re: unpacking.

helpers/helpers.go
kevincwells: Definitely really appreciate you and @cmarcelo's help, here and on IRC. Still new to Go, but usually a quick learner ;)

BTW, thanks for reminding me that I'm ignoring the err return on `.Close()`. I noticed on c3fb081eafa5bdfe31762e45312d6c00529d1c67 that @tmarcu replaced all `defer variable.Close()` with
```go
defer func() {
 	_ = variable.Close()
 }()
```
to get rid of linting warnings. Guessing he probably wants me to do the same.

While I'm at it, I'm going to go through the Travis output to see all the other places linting yells at me.

icarus-sparry: @kevincwells You do know how to install the linter locally?

Handling errors associated with close is always something for debate.Traditionally the C stdio library fclose function causes a flush of the buffer, and that is a place where you can get 'out of disk' or 'out of quota' errors returned from write.  The actual errors that get returned from close(2) are very limited, essentially EIO, which there is almost nothing you can do about. With go, you are encouraged to do the flush yourself, to make the error handling more explicit.

helpers/helpers.go
kevincwells: Just checked it out -- nice! Now my brain is off thinking about how to organize helpers like this that would be useful across all of mixer...

helpers/helpers.go
```diff
@@ -153,6 +156,75 @@ func GetIncludedBundles(filename string) ([]string, error) {
...
+// UnpackFile unpacks a .tar or .tar.gz/.tgz file to a given directory.
+// Should be roughly equivalent to "tar -x[z]f file -C dest". Does not
+// overwrite; returns error if file being unpacked already exists.
+func UnpackFile(file string, dest string) error {
+	fr, err := os.Open(file)
+	if err != nil {
+		return err
+	}
+	defer func() {
+
```
cmarcelo: Minor: you can drop this extra newline.

helpers/helpers.go
```diff
@@ -153,6 +156,75 @@ func GetIncludedBundles(filename string) ([]string, error) {
...
+		hdr, err := tr.Next()
+		if err == io.EOF {
+			// end of archive
+			break
+		} else if err != nil {
+			return errors.Wrapf(err, "Error reading contents of tarball: %s", file)
+		}
+
+		// Skip GitHub generated "extended header" file
+		if strings.Compare(hdr.Name, "pax_global_header") == 0 {
```
cmarcelo: I think you can check for hdr.Typeflag == TypeXGlobalHeader.

helpers/helpers.go
```diff
@@ -153,6 +156,75 @@ func GetIncludedBundles(filename string) ([]string, error) {
...
+			if err != nil {
+				return errors.Wrapf(err, "Error unpacking file: %s", out)
+			}
+
+			if _, err := io.Copy(of, tr); err != nil {
+				return errors.Wrapf(err, "Error unpacking file: %s", out)
+			}
+			_ = of.Close()
+		default:
+			return errors.New("Error unpacking file: " + out)
```
cmarcelo: Optional: prefer Errorf.

helpers/helpers.go
```diff
@@ -153,6 +156,75 @@ func GetIncludedBundles(filename string) ([]string, error) {
...
+		case tar.TypeReg:
+			of, err := os.OpenFile(out, os.O_CREATE|os.O_RDWR|os.O_EXCL, os.FileMode(hdr.Mode))
+			if err != nil {
+				return errors.Wrapf(err, "Error unpacking file: %s", out)
+			}
+
+			if _, err := io.Copy(of, tr); err != nil {
+				return errors.Wrapf(err, "Error unpacking file: %s", out)
+			}
+			_ = of.Close()
```
cmarcelo: If copy fails, you still need to Close the file.

    err = io.Copy(...)
    _ = of.Close
    if err != nil { ... }

Alternatively just call of.Close from the error codepath too.

mixer/cmd/bundles.go
```diff
@@ -40,17 +41,20 @@ var addBundlesCmd = &cobra.Command{
...
+				return errors.New("bundle add requires at least 1 argument if --all is not passed")
 			}
 		}
 		bundles := strings.Split(args[0], ",")
 		b := builder.NewFromConfig(config)
-		// TODO change this to return (int, error)
-		numadded := b.AddBundles(bundles, bundleFlags.force, bundleFlags.all, bundleFlags.git)
+		numadded, err := b.AddBundles(bundles, bundleFlags.force, bundleFlags.all, bundleFlags.git)
 		fmt.Println(numadded, " bundles were added")
+		if err != nil {
```
cmarcelo: When we fail, is numadded relevant? IOW: should we move the print here after the error check?

builder/builder.go
```diff
@@ -355,62 +413,103 @@ func (b *Builder) AddBundles(bundles []string, force bool, allbundles bool, git
...
+		if _, err := b.AddBundles(defaultBundles, false, all, false); err != nil {
+			return err
+		}
+
+		// Init the mix-bundles git repo
+		curr, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		if err = os.Chdir(b.Bundledir); err != nil {
```
cmarcelo: Pass "-C", b.Bundledir extra arguments to Git.

Don't do the getwd/chdir/chdir sequence.

cmarcelo: This also applies to other places you use Getwd/Chdir/Git combo.

General guideline is to avoid changing the current process directory unless there is a good reason. Rationale is that is easy to conflict with other parts of the code.

If Git didn't supported -C, the suggestion would be to teach the helper about a directory parameter, and set it in the cmd.Dir before executing.

builder/builder.go
```diff
@@ -294,26 +357,24 @@ func (b *Builder) AddBundles(bundles []string, force bool, allbundles bool, git
 
 	// Check if mix bundles dir exists
 	if _, err := os.Stat(bundledir); os.IsNotExist(err) {
-		helpers.PrintError(errors.New("Mix bundles directory does not exist. Run mixer init-mix"))
-		os.Exit(1)
+		return bundleAddCount, errors.New("Mix bundles directory does not exist. Run mixer init-mix")
```
cmarcelo: Optional: unless there is a reason, I tend to return the zero values when returning errors. This signals to future code changes that they shouldn't care about keeping the count for this particular code path.

builder/builder.go
```diff
@@ -201,83 +201,58 @@ func (b *Builder) SignManifestMoM() error {
...
+func getUpstreamBundlesDir(ver string) string {
+	return filepath.Join(upstreamBundlesBaseDir, fmt.Sprintf(upstreamBundlesVerDir, ver), upstreamBundlesBundleDir)
+}
 
-	URL := "https://github.com/clearlinux/clr-bundles/archive/" + ver + ".tar.gz"
-	err := helpers.Download(repo, URL)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "ERROR: Failed to download clr-bundles, make sure the version is valid: %s\n", err)
-		os.Exit(1)
+func (b *Builder) getUpstreamBundles(ver string, prune bool) error {
```
cmarcelo: Optional: I'd consider making prune a separated function.

Optional (fallback): Unless you have some concrete local change that uses the "false" codepath, remove the argument and always assume true.

builder/builder.go
```diff
@@ -201,83 +201,58 @@ func (b *Builder) SignManifestMoM() error {
...
+	// Make the folder to store upstream bundles if does not exist
+	if _, err := os.Stat(upstreamBundlesBaseDir); os.IsNotExist(err) {
+		if err = os.MkdirAll(upstreamBundlesBaseDir, 0777); err != nil {
```
cmarcelo: MkdirAll return nil if the path already exists, so you don't need the Stat call.

builder/builder.go
```diff
@@ -201,83 +201,58 @@ func (b *Builder) SignManifestMoM() error {
...
-			if err != nil {
-				helpers.PrintError(err)
-				os.Exit(1)
-			}
-			for _, file := range files {
-				if err = helpers.CopyFile(bundles+"/"+file.Name(), clrbundles+file.Name()); err != nil {
-					helpers.PrintError(err)
+		for _, file := range files {
+			// Skip the current version cache, but delete all others
+			if strings.Compare(file.Name(), fmt.Sprintf(upstreamBundlesVerDir, ver)) != 0 {
```
cmarcelo: Use ```if a != b``` for comparing the strings.

Optional: save the result of fmt.Sprintf in a local variable and reuse it.

helpers/helpers.go
```diff
@@ -153,6 +156,75 @@ func GetIncludedBundles(filename string) ([]string, error) {
...
+		if strings.Compare(hdr.Name, "pax_global_header") == 0 {
+			continue
+		}
+
+		out := filepath.Join(dest, hdr.Name)
+
+		switch hdr.Typeflag {
+		case tar.TypeDir:
+			if _, err := os.Stat(out); err != nil {
+				if err = os.Mkdir(out, 0777); err != nil {
```
cmarcelo: Optional: consider using MkdirAll that gives you Stat for free.

builder/builder.go
```diff
@@ -201,83 +201,58 @@ func (b *Builder) SignManifestMoM() error {
...
+
+	tmptarfile := filepath.Join(upstreamBundlesBaseDir, ver+".tar.gz")
+	URL := "https://github.com/clearlinux/clr-bundles/archive/" + ver + ".tar.gz"
+	if err := helpers.Download(tmptarfile, URL); err != nil {
+		return errors.Wrapf(err, "Failed to download bundles for upstream version %s", ver)
 	}
 
-	fmt.Println("Downloaded " + repo)
+	if err := helpers.UnpackFile(tmptarfile, upstreamBundlesBaseDir); err != nil {
+		return errors.Wrapf(err, "Error unpacking bundles for upstream version %s\n%s left for debuging", ver, tmptarfile)
```
cmarcelo: If Unpack fails, you probably want to remove the directory, otherwise it might leave a partial unpack and your cache check some lines above think everything is fine in a second execution. I'm mentioning this because this kind of issue already happened in Mixer in the past.

Suggestion: you are caching the uncompressed contents. Cache only the tarballs instead, tar+gz is cheap and it might give you less moving parts to care about in your cache (one file vs a lot of stuff).

builder/builder.go
```diff
@@ -201,83 +201,58 @@ func (b *Builder) SignManifestMoM() error {
...
-func (b *Builder) UpdateRepo(ver string, allbundles bool) {
-	// Make the folder to store all clr-bundles version
-	if _, err := os.Stat("clr-bundles"); err != nil {
-		if err = os.Mkdir("clr-bundles", 0777); err != nil {
-			fmt.Fprintf(os.Stderr, "ERROR: Failed to create clr-bundles: %s\n", err.Error())
-		}
-	}
+const (
+	upstreamBundlesBaseDir   = ".mixer/upstream-bundles/"
+	upstreamBundlesVerDir    = "clr-bundles-%s"
```
cmarcelo: Optional: might worth adding Format or Fmt to the prefix to indicate this is not a dir to be used directly.

builder/builder.go
kevincwells: >Optional: I'd consider making prune a separated function.

I considered that, but I didn't want the vast majority of functions to have to make two calls. (I wish Go supported default values for variables like Python does. Or even nil values for boolean primitives would allow default programming easier. I could always make it a variadic function to make it so folks don't have to pass the prune flag if they don't want). Bottom line, I'll pull `prune` out into a separate function when we hit a use case where people need to prune outside of fetching.

>Optional (fallback): Unless you have some concrete local change that uses the "false" codepath, remove the argument and always assume true.

The concrete example using false is the forthcoming "update my bundles to a new upstream version #", where you need to have the old CLR bundles and the new CLR bundles at the same time to create diffs.

builder/builder.go
kevincwells: Ah, that's a nice simplification.

builder/builder.go
kevincwells: >Use if a != b for comparing the strings.

Just read the [comments in strings.Compare source](https://golang.org/src/strings/compare.go). Good to know.

>Optional: save the result of fmt.Sprintf in a local variable and reuse it.

I might be missing your point, but I'm not re-calculating that value anywhere in this function, so not sure what this would do.

builder/builder.go
kevincwells: Yeah, that's the convention I tried to follow elsewhere.  For this, though, the call affects the file system. If it thinks it has successfully added any bundles, that means those files actually now exist in the `mix-bundles` directory. As is, the user gets notified that bundles were in fact added, even if there was a failure along the way.

A cleaner implementation (for the user) would be to keep track of which bundles were actually added (as opposed to skipped for already being there, etc.) and attempt to rewind/clean-up in the event of an error. All of this is going away with the local/upstream bundle rewrite I'm doing this week, though, so I don't plan to implement that kind of bookkeeping for the time being.

builder/builder.go
kevincwells: To which? The format number itself (as opposed to the word) is in name of the directory the bundles get unpacked into.

As is, it ends up as `<dir where mixer is run>/.mixer/upstream-bundles/clr-bundles-1930/bundles/<bundle file>`.  Everything after `upstream-bundles` is a side effect of the structure within the tarball. (Going forward, we probably want a more generic solution to support derivatives that don't package their bundle definitions the same as CLR, but that's a bigger problem to solve for later.)

If you had two copies checked out simultaneously (`prune` as `false`), you'd have, say, `.mixer/upstream-bundles/clr-bundles-1930` and `.mixer/upstream-bundles/clr-bundles-1980` side-by-side.

Going forward (as mentioned in my proposals and in the meeting this morning), there will be other directories inside `.mixer`. The `upstream-bundles` directory is one that will be in the `.gitignore`.

builder/builder.go
cmarcelo: Optional: the suggestion was outside the for loop, you create a variable (e.g. 'verDir') that holds the result of fmt.Sprintf(...) call. Since that will not change for each iteration, you can reuse that for each iteration.

builder/builder.go
cmarcelo: Optional: since there is a reason for those partial values even when there is an error, document it on the function comment something along the lines: "when there is error, the int value will contain the bundles added so far".

"All of this is going away with the local/upstream bundle rewrite I'm doing this week, though, so I don't plan to implement that kind of bookkeeping for the time being." -> OK!

builder/builder.go
cmarcelo: "concrete example using false is the forthcoming" -> Fair enough.

Another alternative is to create constants and take a flags numeric argument, so that call sites look like

    getUpstreamBundles(ver, Prune)

Since this code is in flux at the moment, I don't mind keeping as is.

builder/builder.go
kevincwells: Yeah, that's good practice. Kind of gotten out of habit for things like that, as most compilers for other languages will optimize that stuff automatically these days; does Go not?

builder/builder.go
kevincwells: Yeah, good call. This is really old code I've just moved around, but you're right -- it's worth cleaning it up now.

builder/builder.go
cmarcelo: My guess in this case is that Go compiler can't infer that Sprintf will not have side effects (unless it inlines it, which I don't think happen in this case), so it can't automatically make this transformation. (We should just check compiler output!)

Note: in this case I think the change is worth even if it doesn't optimize anything.

builder/builder.go
cmarcelo: Clarifying, I meant a suffix (not prefix! sorry) to the variable name (not directory names in the filesystem), the suggestion was to call upstreamBundlesVerDir something like upstreamBundlesVerDirFormat (the one in the line this comment refers to).

Optional: When in those situations I noticed I end up creating a function to produce the right path for me instead of giving the format out. The calling code looks a bit better.

cmarcelo: IOW, I was not talking about anything related to the paragraphs 2-4 of your reply ;-)

builder/builder.go
kevincwells: Fixed in forthcoming force push.

builder/builder.go
kevincwells: Fixed in forthcoming push.

builder/builder.go
kevincwells: Fixed in forthcoming push.

builder/builder.go
kevincwells: Ohhhhhh, completely misread your comment. Good idea. Fixed in forthcoming push.

builder/builder.go
kevincwells: Fixed in forthcoming push.

icarus-sparry: @kevincwells Actually few compilers will do that kind of optimization of function calls. In particular gcc C will not unless the function being called is annotated either `pure` or `const`. Of course some languages are side effect free and compilers for those languages will do the code hoisting you mention.

builder/builder.go
kevincwells: >If Unpack fails, you probably want to remove the directory, otherwise it might leave a partial unpack and your cache check some lines above think everything is fine in a second execution. I'm mentioning this because this kind of issue already happened in Mixer in the past.

Great catch. Fixed in upcoming push.

>Suggestion: you are caching the uncompressed contents. Cache only the tarballs instead, tar+gz is cheap and it might give you less moving parts to care about in your cache (one file vs a lot of stuff).

I considered that, but I decided that
1. I'd rather not have to untar+gz for each command, even if it's cheap. Just feels like unnecessary work to do each time.
1. While they're uncompressed, they're small text files, so compression isn't buying us a lot of savings from that perspective.
1. It's a _little bit_ more future-proof to rely on the bundle output itself, rather than the way mainline CLR currently serves up its bundle definitions as a .tgz. This isn't a great argument, 'cause it's still pretty darn hard-coded to the way CLR does it right now. As I mentioned elsewhere, we have a problem for derivatives of CLR: the way we fetch upstream definitions isn't generic/hot-swappable for something else. We could parameterize the URL we fetch, but that still locks derivatives into serving their bundle definitions as .tgz's with the same internal file structure as mainline CLR.

builder/builder.go
kevincwells: @icarus-sparry & @cmarcelo -- yeah, you're both right. I'm running on little sleep today, so I wasn't thinking about function side effects when I made my remark.

helpers/helpers.go
kevincwells: Weird. Thanks for catching that. Fixed in forthcoming push.

helpers/helpers.go
kevincwells: Fixed in forthcoming push.

helpers/helpers.go
kevincwells: Nice catch. Fixed in forthcoming push.

helpers/helpers.go
kevincwells: Fixed in forthcoming push.

kevincwells:
@cmarcelo thanks for the thorough review! I just did a push that I believe addresses each of your comments.

helpers/helpers.go
cmarcelo: > Just checked it out -- nice! Now my brain is off thinking about how to organize helpers like this that would be useful across all of mixer...

If multiple packages are using it the functionality, we either make one of them export, or if it is out of scope for that package, export in a separate package. Depending on the nature of the feature, it might make sense to make it an internal package, i.e. only our packages can import it, but external code not. This separates what we want to reuse internally vs. what we want to export as API to external users. See https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU

(Note: our helpers likely should be an internal package)

YMMV, but is better to have packages with purpose than a grab bag of utilities. (Thinking now, internal packages might reduce the issue with the grab bag style of packages, but unsure...). My strategy is to keep utilities that don't deserve a package for them inside the package that use them. See also the section on bad names https://blog.golang.org/package-names

The concrete answer for this case: we have various archive+compress related code spawning here and there, so soon we might have enough use cases to create a mixer-tools/internal/archive package or similar.

helpers/helpers.go
```diff
@@ -153,6 +156,71 @@ func GetIncludedBundles(filename string) ([]string, error) {
...
+		}
+
+		out := filepath.Join(dest, hdr.Name)
+
+		switch hdr.Typeflag {
+		// Skip GitHub generated "extended header" file
+		case tar.TypeXGlobalHeader:
+			continue
+		case tar.TypeDir:
+			if err = os.MkdirAll(out, 0777); err != nil {
```
cmarcelo: Optional: the tar header have the permissions for directories too, I wonder if you could just use them.

kevincwells: Huh, yeah, good catch. Simple fix. Push coming.

helpers/helpers.go
tmarcu: @cmarcelo why?

tmarcu: Ignore me.

helpers/helpers.go
tmarcu: It would be nice to not duplicate archive compress/extract in multiple parts if we can re-use and extend 1 package instead.

kevincwells:
@tmarcu -- is anything actually holding this up from being merged now? I know you'd said you wanted to update documentation as this changes where upstream bundles live, but the documentation is already out of date due to the entire CLI rewrite.

#107 is blocked, in part, on this being merged.

PR #90, opened 2018-01-12
Write update metadata to output directory

matthewrsj:
Instead of requiring the swupd-server code to create the "format" and
"swupd-server-src-version" metadata files enable BuildUpdate to do so.

Because the mixer version is now used instead of a swupd-server version
move the Version const to builder.go so it can be accessed trivially.
Update the Makefile parsing to point to builder.go and update the access
method in the top-level mixer/cmd/root.go.

RFC: should we change the name of `swupd-server-src-version` file to `mixer-src-version` or something similar?

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -35,6 +35,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Version of Mixer. Also used by the Makefile for releases.
+const Version = "3.2.1"
+
 // UseNewSwupdServer controls whether to use the new implementation of
 // swupd-server (package swupd) when possible. This is an experimental feature.
 var UseNewSwupdServer = false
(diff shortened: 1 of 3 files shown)


cmarcelo:
The key question is: what programs use "swupd-server-src-version" and for what?

cmarcelo:
Depending on the answer, it might be a breaking change to put mixer version there.

(The patch itself look good, we just need to know if we can do it :)

matthewrsj:
After speaking with @gtkramer it looks like this file is not used anywhere. It is nice to have though because it keeps a record of the source code version that built the update content. With that knowledge, I'll change the file name to `mixer-src-version`.

cmarcelo:
> With that knowledge, I'll change the file name to mixer-src-version.

Sounds good!

builder/builder.go
```diff
@@ -557,7 +560,8 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string,
 	}
 
 	// Ensure the format dir exists.
-	formatDir := filepath.Join(b.Statedir, "www", "version", "format"+b.Format)
+	outputDir := filepath.Join(b.Statedir, "www")
+	formatDir := filepath.Join(outputDir, "version", "format"+b.Format)
```
tmarcu: why not just 
`formatDir := filepath.Join(b.Statedir, "www", "version", "format"+b.Format)` ? I don't see it being used under here.

matthewrsj: Oops, leftover change from before splitting out new-swupd and old-swupd code. Fixed now.

PR #93, opened 2018-01-14
Add 'mixer build delta-packs' when using --new-swupd

cmarcelo:
This is on top of #79.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -697,6 +697,7 @@ func (b *Builder) buildUpdateWithNewSwupd(timer *stopWatch, mixVersion uint32, m
 	timer.Stop()
 
 	timer.Start("CREATE ZERO PACKS")
+	chrootDir := filepath.Join(b.Statedir, "image")
 	for _, bundle := range mom.Files {
 		// TODO: Evaluate if it's worth using goroutines.
 		name := bundle.Name
@@ -714,7 +715,7 @@ func (b *Builder) buildUpdateWithNewSwupd(timer *stopWatch, mixVersion uint32, m
 		fmt.Printf("Creating zero pack for %s to version %d\n", name, version)
 
 		var info *swupd.PackInfo
-		info, err = swupd.CreatePack(name, 0, version, outputDir, fullChrootDir)
+		info, err = swupd.CreatePack(name, 0, version, outputDir, chrootDir)
 		if err != nil {
 			return errors.Wrapf(err, "couldn't make pack for bundle %s", name)
 		}
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -937,3 +937,128 @@ func archiveFiles(w io.Writer, srcs []string) error {
 	}
 	return tw.Close()
 }
+
+// BuildDeltaPacks between two versions of the mix.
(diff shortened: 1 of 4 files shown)


cmarcelo:
The two last commits are the ones for this PR, others come from #79.

builder/builder.go
```diff
@@ -939,3 +939,128 @@ func createCompressedArchive(dst string, srcs ...string) (err error) {
...
+
+	if to == 0 {
+		to = mixVersion
+	} else {
+		if to > mixVersion {
+			return errors.Errorf("--to version must be at most the latest mix version (%d)", mixVersion)
+		}
+	}
+	if from >= to {
+		return errors.Errorf("the --from version must not be larger than the --to version")
```
matthewrsj: "must not be larger than **or equal to** the --to version"

builder/builder.go
```diff
@@ -650,20 +651,42 @@ func (b *Builder) buildUpdateWithNewSwupd(timer *stopWatch, mixVersion uint32, m
 	for _, f := range mom.Files {
 		fmt.Printf("- %-20s %d\n", f.Name, f.Version)
 	}
-	if !skipSigning {
+
+	if skipSigning {
+		fmt.Println("SKIPPING manifest signing.")
```
tmarcu: Don't need to print this, just remove this if IMHO.

builder/builder.go
```diff
@@ -852,3 +875,66 @@ func parseUint32(s string) (uint32, error) {
 	}
 	return uint32(parsed), nil
 }
+
+func createCompressedArchive(dst string, srcs ...string) (err error) {
```
tmarcu: Should this be exported so we can use it elsewhere in the code?

builder/builder.go
pohly: It isn't clear what gets negated by  "not": "not (larger than or equal)" or "(not larger) or equal" - isn't language wonderful? ;-}

I think this is simpler: "the --from version must be smaller than the --to version".

builder/builder.go
cmarcelo: Removed this from the commit that added it (and other PR) and from this one.

builder/builder.go
cmarcelo: This function wasn't written with the intent of being exported. If we see value later in exporting it, we can revisit it to make it so.

builder/builder.go
cmarcelo: Yes, the "smaller" reads better. Fixed.

mixer/cmd/build.go
```diff
@@ -137,6 +137,69 @@ var buildImageCmd = &cobra.Command{
...
+}
+
+func runBuildDeltaPacks(cmd *cobra.Command, args []string) error {
+	if !builder.UseNewSwupdServer {
+		// TODO: Depending on how long we are going to live with both implementations, it
+		// might be worth making this command call the script when not using new swupd-server.
+		failf("build delta packs is only available with --new-swupd\nUse mixer-pack-maker.sh instead.")
+	}
+
+	fromChanged := cmd.Flags().Changed("from")
```
tmarcu: Cool, haven't used Changed() before :)

PR #94, opened 2018-01-16
mixer: add CPU profiling if flag passed

cmarcelo:
The flag --cpu-profile let the caller specify a filename to write the
CPU profile information to. That can be later read with
    
    $ go tool pprof FILENAME
    
More details on how to use the tool is in
https://blog.golang.org/profiling-go-programs.
    
The new flag is hidden from the default help message -- since it is
mostly a feature for developers.

### DIFF
--- a/mixer/cmd/root.go
+++ b/mixer/cmd/root.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"runtime/pprof"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmarcelo:
This PR is on top of #79. Only the last patch is new.

I had a positive experience using this flag to investigate create-fullfiles helper program, now that we will have more Go written functionality in mixer, profiling information might be useful.

matthewrsj:
+1

tmarcu:
+1 !

PR #95, opened 2018-01-16
Exporting more Hash related functions

cmarcelo:
Following new functionality is exported:
- Calculate swupd hashes without having a file on the disk
- Calculate swupd hashes without having a the whole data in memory

Exports a type containing exactly the metadata used by the swupd hash.

Doesn't change anything in the Hashval or interning, everything should work as before.

### DIFF
--- a/swupd/hash.go
+++ b/swupd/hash.go
@@ -20,7 +20,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"hash"
-	"io/ioutil"
+	"io"
 	"os"
 	"syscall"
 )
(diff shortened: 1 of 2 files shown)


cmarcelo:
There isn't an use case yet in swupd content generation. This is more useful for "client" type applications. It could also help smooth things for changes to support reading the UID/GID information from other sources.

I have non swupd-server code that uses it to validate the tarball before writing the extracted contents to disk. Would be useful to have it part of the swupd package, but I'm fine closing this and waiting until we have an official usage of this functionality.

swupd/hash.go
```diff
@@ -241,3 +173,67 @@ func (h *Hash) Sum() string {
...
+	if err != nil {
+		return "", fmt.Errorf("error creating hash for file %s: %s", filename, err)
+	}
+
+	if info.Mode&syscall.S_IFMT == syscall.S_IFREG {
+		f, err := os.Open(filename)
+		if err != nil {
+			return "", fmt.Errorf("read error for file %s: %s", filename, err)
+		}
+		defer func() {
```
icarus-sparry: My usual comment about not deferring something for 1 line.

swupd/hash.go
```diff
@@ -241,3 +173,67 @@ func (h *Hash) Sum() string {
...
+
+	hashInfo := &HashFileInfo{
+		Mode: info.Mode,
+		UID:  info.Uid,
+		GID:  info.Gid,
+		Size: info.Size,
+	}
+
+	if info.Mode&syscall.S_IFMT == syscall.S_IFLNK {
+		fmt.Printf("\n\n\n%s  %x  %x  %x\n\n\n\n", filename, info.Mode, syscall.S_IFLNK, info.Mode&syscall.S_IFLNK)
```
icarus-sparry: Debugging Printf ???

cmarcelo: Haha, sorry for that :-)  Fixed. Was learning how to use S_IFMT.

swupd/hash.go
cmarcelo: Done.

cmarcelo:
Pushed again to trigger travis.

swupd/hash.go
```diff
@@ -241,3 +173,63 @@ func (h *Hash) Sum() string {
 	hex.Encode(result[:], h.hmac.Sum(nil))
 	return string(result[:])
 }
+
+// GetHashFile calculate the swupd hash for a file in the disk.
+func GetHashFile(filename string) (string, error) {
```
tmarcu: Sorry if I'm being too picky...but GetHashFile sounds like it's returning me a - Hash "file", not a string. Would it make more sense to just call it GetHash? or GetFileHash()?

cmarcelo: Maybe GetHashForFile and GetHashForBytes?

icarus-sparry: @tmarcu You can (almost) never be too picky about exported Names

swupd/hash.go
```diff
@@ -134,8 +135,109 @@ func hmacComputeKey(info syscall.Stat_t) []byte {
...
+	}
+	hex.Encode(key[:], mac.Sum(nil))
+
+	// With the key in hand, create the HMAC struct that will be
+	// used to write data to.
+	h := &Hash{
+		hmac: hmac.New(sha256.New, key[:]),
+	}
+
+	// Pre-write data we known so that directories and symbolic
```
tmarcu: s/known/know

PR #100, opened 2018-01-19
Add delta creation code

tmarcu:
This isn't hardcoded to use Manifest.Full to be more general, but it should be passed in as the manifest to use for our case.

### DIFF
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -305,3 +305,29 @@ func Git(args ...string) error {
 	}
 	return nil
 }
+
+// RunCommand runs the given command with args and prints output
+func RunCommand(cmdname string, args ...string) error {
+	cmd := exec.Command(cmdname, args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	err := cmd.Run()
+	if err != nil {
+		return errors.Wrapf(err, "Failed to run %s %s: %v\n", cmdname, strings.Join(args, " "), err)
+	}
+
+	return nil
+}
+
+// RunCommandSilent runs the given command with args and does not print output
+func RunCommandSilent(cmdname string, args ...string) error {
+	cmd := exec.Command(cmdname, args...)
+	// Automatically connected to os.DevNull by implementation
+	// because cmd.Stdout/Stderr are nil in this case
+	err := cmd.Run()
+	if err != nil {
+		return errors.Wrapf(err, "Failed to run %s %s: %v\n", cmdname, strings.Join(args, " "), err)
+	}
+
+	return nil
+}
--- a/swupd/delta.go
+++ b/swupd/delta.go
@@ -1,0 +1,179 @@
+package swupd
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/clearlinux/mixer-tools/helpers"
+	"github.com/pkg/errors"
+)
+
+// CreateDeltas creates all delta files between the previous and current
+// version of the supplied manifest, returning a list of files that
+// delta creation failed for.
+func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
+	var c config
+
+	c, err := getConfig(statedir)
+	if err != nil {
+		return nil, err
+	}
+
+	var oldManifest *Manifest
+	var newManifest *Manifest
+
+	if oldManifest, err = ParseManifestFile(filepath.Join(c.outputDir, fmt.Sprintf("%d", from), manifest)); err != nil {
+		return nil, err
+	}
+	if newManifest, err = ParseManifestFile(filepath.Join(c.outputDir, fmt.Sprintf("%d", to), manifest)); err != nil {
+		return nil, err
+	}
+
+	// Must be sorted before passing into linkPeersAndChange()
+	oldManifest.sortFilesName()
+	newManifest.sortFilesName()
+
+	_, _, _ = newManifest.linkPeersAndChange(oldManifest, from)
+
+	deltas, err := consolidateDeltaFiles(newManifest, from, c)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to create deltas list %s", manifest)
+	}
+
+	const maxRoutines = 3
+	var deltaQueue = make(chan *File)
+	var failedChan = make(chan *File)
+	var failedList = make(chan []*File)
+	var wg sync.WaitGroup
+	wg.Add(maxRoutines)
+
+	// Start a collector for the failure cases
+	go func() {
+		var failed []*File
+		// Drains the failedChan appropriately
+		for f := range failedChan {
+			failed = append(failed, f)
+		}
+		failedList <- failed // This list is what we return in the end
+	}()
+
+	// Don't flood the system with goroutines, delta creation takes up an
+	// incredibly amount of memory, so we cannot max out on goroutines
+	for i := 0; i < maxRoutines; i++ {
+		go func(statedir string) {
+			defer wg.Done()
+
+			// Take in jobs from the queue and try to make a delta
+			for f := range deltaQueue {
+				err := createDelta(c, f, statedir, f.DeltaPeer.Version, f.DeltaPeer.Hash)
+				if err != nil {
+					failedChan <- f
+				}
+			}
+		}(statedir)
+	}
+
+	// Send jobs to the queue for delta goroutines to pick up
+	for _, file := range deltas {
+		deltaQueue <- file
+	}
+
+	// Send message that no more jobs are being sent
+	close(deltaQueue)
+	wg.Wait()
+
+	// Once wait finishes we can signal the collector that no more jobs exist
+	close(failedChan)
+
+	return <-failedList, nil
+}
+
+func consolidateDeltaFiles(manifest *Manifest, from uint32, c config) ([]*File, error) {
+	var files []*File
+
+	if manifest == nil {
+		return nil, nil
+	}
+
+	for _, file := range manifest.Files {
+		if file.Version <= from {
+			continue
+		}
+		if file.DeltaPeer == nil {
+			continue
+		}
+		if file.Type != TypeFile ||
+			file.DeltaPeer.Type != TypeFile {
+			continue
+		}
+
+		deltadir := filepath.Join(c.outputDir, fmt.Sprintf("%d", file.Version), "delta")
+		fromToString := fmt.Sprintf("%d-%d-%s-%s", file.DeltaPeer.Version, file.Version, file.DeltaPeer.Hash, file.Hash)
+
+		deltaFile := filepath.Join(deltadir, fromToString)
+
+		// Only add to list if the delta file does not already exist on the system
+		if _, err := os.Stat(deltaFile); os.IsNotExist(err) {
+			// Only create deltas for files bigger than 200 bytes
+			fullname := filepath.Join(c.imageBase, fmt.Sprintf("%d", file.Version), "full", file.Name)
+			file.Info, _ = os.Stat(fullname)
+			if file.Info.Size() > 200 {
+				files = append(files, file)
+			}
+		}
+
+	}
+
+	return files, nil
+}
+
+func createDelta(c config, file *File, statedir string, fromVersion uint32, fromHash Hashval) error {
+	// File without peers cannot have deltas
+	if file.DeltaPeer == nil {
+		return nil
+	}
+
+	// We only support deltas between two regular files for now
+	// TODO: Support directory -> file in the future, this is a complex case
+	if file.Type != TypeFile || file.DeltaPeer.Type != TypeFile {
+		return nil
+	}
+
+	newfile := filepath.Join(c.imageBase, fmt.Sprintf("%d", file.Version), "full", file.Name)
+	original := filepath.Join(c.imageBase, fmt.Sprintf("%d", fromVersion), "full", file.Name)
+
+	deltadir := filepath.Join(c.outputDir, fmt.Sprintf("%d", file.Version), "delta")
+	fromToString := fmt.Sprintf("%d-%d-%s-%s", fromVersion, file.Version, fromHash, file.Hash)
+
+	// Files to create and validate deltas with
+	deltafile := filepath.Join(deltadir, fromToString)
+	testnewfile := filepath.Join(deltadir, "."+fromToString+".testnewfile")
+
+	// Shell out to bsdiff for now...false = don't print to screen
+	if err := helpers.RunCommandSilent("bsdiff", original, newfile, deltafile); err != nil {
+		_ = os.Remove(deltafile) // Might have returned FULLDL
+		return errors.Wrapf(err, "Failed to create delta for %s -> %s", original, newfile)
+	}
+
+	// Check that the delta actually applies correctly
+	if err := helpers.RunCommandSilent("bspatch", original, testnewfile, deltafile); err != nil {
+		return errors.Wrapf(err, "Failed to apply delta %s", deltafile)
+	}
+
+	deltahash, err := Hashcalc(testnewfile)
+	if err != nil {
+		return errors.Wrap(err, "Failed to calculate hash for new file")
+	}
+
+	if !HashEquals(deltahash, file.Hash) {
+		return errors.Wrapf(err, "Delta mismatch: %s -> %s via delta: %s", original, newfile, deltafile)
+	}
+
+	_ = os.Remove(testnewfile)
+
+	// add rename code here
+
+	return nil
+}
--- a/swupd/delta_test.go
+++ b/swupd/delta_test.go
@@ -1,0 +1,26 @@
+package swupd
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+var largeString = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque tempor quam id convallis convallis. Proin vehicula nisi in augue posuere lobortis. Quisque tincidunt elit ac facilisis auctor. Praesent facilisis ex eros, nec blandit dui suscipit porta. Nunc id nunc rhoncus, condimentum nibh at, fermentum sem. Nam mollis justo ac iaculis gravida. Vestibulum convallis congue dolor, vitae rutrum ex finibus vel. Phasellus convallis sem nunc, ac laoreet risus rhoncus fermentum. Interdum et malesuada fames ac ante ipsum primis in faucibus. Nullam eget pellentesque nulla. Etiam tristique non magna quis consectetur. In ac dolor sagittis, vulputate ante tincidunt, suscipit leo."
+
+func TestCreateDeltas(t *testing.T) {
+	testDir := mustSetupTestDir(t, "deltas")
+	defer removeIfNoErrors(t, testDir)
+	mustInitStandardTest(t, testDir, "0", "10", []string{"test-bundle1", "test-bundle2"})
+	mustGenFile(t, testDir, "10", "test-bundle1", "foo", largeString)
+	mustGenFile(t, testDir, "10", "test-bundle2", "bar", largeString)
+	mustCreateManifestsStandard(t, 10, testDir)
+
+	mustInitStandardTest(t, testDir, "10", "20", []string{"test-bundle1", "test-bundle2"})
+	mustGenFile(t, testDir, "20", "test-bundle1", "foo", largeString)
+	mustGenFile(t, testDir, "20", "test-bundle2", "bar", largeString+"testingdelta")
+	mustCreateManifestsStandard(t, 20, testDir)
+	mustMkdir(t, filepath.Join(testDir, "www/20/delta"))
+
+	mustCreateDeltas(t, "Manifest.full", testDir, 10, 20)
+	mustExistDelta(t, testDir, "/bar", 10, 20)
+}
--- a/swupd/helpers_test.go
+++ b/swupd/helpers_test.go
@@ -186,6 +186,27 @@ func mustExist(t *testing.T, name string) {
 	}
 }
 
+func mustExistDelta(t *testing.T, testDir, filename string, from, to uint32) {
+	var fromFull *Manifest
+	var toFull *Manifest
+	var err error
+	if fromFull, err = ParseManifestFile(filepath.Join(testDir, "www", fmt.Sprintf("%d", from), "Manifest.full")); err != nil {
+		t.Fatalf("Failed to load from manifest to read hash from: %q", err)
+	}
+	if toFull, err = ParseManifestFile(filepath.Join(testDir, "www", fmt.Sprintf("%d", to), "Manifest.full")); err != nil {
+		t.Fatalf("Failed to load to manifest to read hash from: %q", err)
+	}
+
+	var fileNeeded = &File{Name: filename}
+	fromHash := fileNeeded.findFileNameInSlice(fromFull.Files).Hash
+	toHash := fileNeeded.findFileNameInSlice(toFull.Files).Hash
+
+	suffix := fmt.Sprintf("%d-%d-%s-%s", from, to, fromHash, toHash)
+	deltafile := filepath.Join(testDir, "www", fmt.Sprintf("%d", to), "delta", suffix)
+
+	mustExist(t, deltafile)
+}
+
 func mustNotExist(t *testing.T, name string) {
 	_, err := os.Stat(name)
 	if !os.IsNotExist(err) {
@@ -248,5 +269,18 @@ func checkManifestMatches(t *testing.T, testDir, ver, name string, res ...*regex
 		if !re.Match(b) {
 			t.Errorf("%v not found in %s/Manifest.%s", re.String(), ver, name)
 		}
+	}
+}
+
+func mustCreateDeltas(t *testing.T, manifest, statedir string, from, to uint32) {
+	failed, err := CreateDeltas(manifest, statedir, from, to)
+	if err != nil {
+		if len(failed) > 0 {
+			t.Fatalf("%s: \n%d deltas did not get created for %s: %v", err, len(failed), manifest, failed)
+			return
+		}
+		t.Fatal(err)
+	} else if len(failed) > 0 {
+		t.Fatalf("CreateDeltas succeeded but %d deltas did not get created for %s: %v", len(failed), manifest, failed)
 	}
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ before_install:
     - sudo apt-get update -qq
     - go get -u gopkg.in/alecthomas/gometalinter.v2
     - gometalinter.v2 --install
+    - wget https://github.com/clearlinux/bsdiff/releases/download/v1.0.2/bsdiff-1.0.2.tar.xz
+    - tar -xvf bsdiff-1.0.2.tar.xz
+    - pushd bsdiff-1.0.2 && ./configure --prefix=/usr --disable-tests && make -j48 && sudo make install && popd
 
 script:
     - make lint


swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+
+		deltaFile := fmt.Sprintf("%s/%d/delta/%d-%d-%s-%s",
+			c.outputDir, file.Version, file.DeltaPeer.Version,
+			file.Version, file.DeltaPeer.Hash, file.Hash)
+
+		// Only add to list if the delta file does not already exist on the
+		// system or in the list
+		if _, err := os.Stat(deltaFile); os.IsNotExist(err) {
+			if found := file.findFileNameInSlice(files); found == nil {
+				// Prepend to list...
```
tmarcu: Can this be changed to just regular append()?

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+	var c config
+	c, err := getConfig(statedir)
+	if err != nil {
+		return nil, err
+	}
+
+	var oldManifest *Manifest
+	var newManifest *Manifest
+
+	if oldManifest, err = ParseManifestFile(fmt.Sprintf("%s/%d/%s", c.outputDir, manifest.Header.Previous, manifest.Name)); err != nil {
```
matthewrsj: Prefer `filepath.Join()` to handle combining paths.

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, err
+	}
+	if newManifest, err = ParseManifestFile(fmt.Sprintf("%s/%d/%s", c.outputDir, manifest.Header.Version, manifest.Name)); err != nil {
```
matthewrsj: same - `filepath.Join()`

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+	"sync"
+
+	"github.com/clearlinux/mixer-tools/helpers"
+	"github.com/pkg/errors"
+)
+
+// CreateDeltas creates all delta files between the previous and current
+// version of the supplied manifest, returning a list of files that
+// delta creation failed for.
+func CreateDeltas(manifest *Manifest, statedir string) ([]*File, error) {
```
matthewrsj: Do we only care about the errors here, or would it be nice to have at least a count of delta files?

matthewrsj: This function only creates deltas between a manifest's version and its previous version. What if we want to create deltas from an older version of that manifest, say 5 versions back?

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+
+func consolidateDeltaFiles(manifest *Manifest, c config) ([]*File, error) {
+	var files []*File
+
+	if manifest == nil {
+		return nil, nil
+	}
+
+	for _, file := range manifest.Files {
+		if file.Version <= manifest.Header.Version ||
```
matthewrsj: As mentioned above, this only works if we only create deltas between current and previous.

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+			file.DeltaPeer == nil ||
+			(file.Type != typeFile && file.Type != typeDirectory && file.Type != typeLink) {
```
matthewrsj: We only create deltas between typeFile and typeFile, so the check here should be
```
file.Type != typeFile || file.DeltaPeer.Type != typeFile
```

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+			continue
+		}
+
+		deltaFile := fmt.Sprintf("%s/%d/delta/%d-%d-%s-%s",
```
matthewrsj: `filepath.Join()` please

swupd/delta.go
matthewrsj: Since you are doing nothing else with these files except returning them and don't rely on them to be sorted this can be a regular append, which will be much faster.

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+			c.outputDir, file.Version, file.DeltaPeer.Version,
+			file.Version, file.DeltaPeer.Hash, file.Hash)
+
+		// Only add to list if the delta file does not already exist on the
+		// system or in the list
+		if _, err := os.Stat(deltaFile); os.IsNotExist(err) {
+			if found := file.findFileNameInSlice(files); found == nil {
```
matthewrsj: I don't think this check is necessary since the only source for the `files` list is the `manifest.Files` list, which you only iterate once and will not have any duplicate file names. This is also a very expensive check since it is simply a linear search on an unsorted slice.

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil
+	}
+
+	// We only support deltas between two regular files for now
+	if file.Type != typeFile || file.DeltaPeer.Type != typeFile {
+		return nil
+	}
+
+	// By definition we cannot make deltas for files that are deleted
+	if file.Status == statusDeleted {
```
matthewrsj: Don't forget `file.Status == statusGhosted` as well since `file` is the `to` part of the delta and it doesn't make sense for a `to` to be ghosted.

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil
+	}
+
+	// TODO: Feature to support in the future, this is a complex case
+	if file.Type == typeDirectory {
```
matthewrsj: Except you already return if `file.Type != typeFile` above. Maybe move the TODO up there?

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil
+	}
+
+	newfile := fmt.Sprintf("%s/%d/full/%s", c.imageBase, file.Version, file.Name)
+	original := fmt.Sprintf("%s/%d/full/%s", c.imageBase, fromVersion, file.Name)
```
matthewrsj: `filepath.Join()` and you can generalize this to supporting more versions by tracking the version with `file.DeltaPeer.Version`

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+
+	deltafile := fmt.Sprintf("%s/%d/delta/%d-%d-%s-%s", c.outputDir, file.Version, fromVersion, file.Version, fromHash, file.Hash)
+	dotfile := fmt.Sprintf("%s/%d/delta/.%d-%d-%s-%s", c.outputDir, file.Version, fromVersion, file.Version, fromHash, file.Hash)
+	testnewfile := fmt.Sprintf("%s/%d/delta/.%d-%d-%s-%s.testnewfile", c.outputDir, file.Version, fromVersion, file.Version, fromHash, file.Hash)
+
+	if sameFile(file, file.DeltaPeer) {
```
matthewrsj: Maybe move this higher before you do all the work with strings above?

helpers/helpers.go
```diff
@@ -237,3 +238,22 @@ func Git(args ...string) {
 		os.Exit(1)
 	}
 }
+
+// RunCommand runs the given command with args and optionally prints
+func RunCommand(print bool, cmdname string, args ...string) error {
```
matthewrsj: Optional: make a `RunCommandSilent` and a `RunCommandPrint` (meh on the name) that just call an internal function to get rid of the bool. I mention this because of advice I got from @icarus-sparry and @cmarcelo on a similar test helper.

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+		return errors.Wrap(err, "Failed to calculate hash for new file")
+	}
+
+	if !HashEquals(deltahash, file.Hash) {
+		return errors.Wrapf(err, "Delta mismatch: %s -> %s via delta: %s", original, newfile, deltafile)
+	}
+
+	_ = os.Remove(testnewfile)
+
+	// add rename code here
```
matthewrsj: This may not be necessary depending on rename implementation. Rename code should be setting `file.DeltaPeer` correctly... @icarus-sparry?

swupd/delta.go
tmarcu: woops, this was supposed to be changed before pushing 👍

swupd/delta.go
tmarcu: Count of total deltas or count of successfully created deltas?

swupd/delta.go
```diff
@@ -0,0 +1,160 @@
...
+	deltafile := fmt.Sprintf("%s/%d/delta/%d-%d-%s-%s", c.outputDir, file.Version, fromVersion, file.Version, fromHash, file.Hash)
+	dotfile := fmt.Sprintf("%s/%d/delta/.%d-%d-%s-%s", c.outputDir, file.Version, fromVersion, file.Version, fromHash, file.Hash)
+	testnewfile := fmt.Sprintf("%s/%d/delta/.%d-%d-%s-%s.testnewfile", c.outputDir, file.Version, fromVersion, file.Version, fromHash, file.Hash)
+
+	if sameFile(file, file.DeltaPeer) {
+		return nil
+	}
+
+	// Shell out to bsdiff for now...false = don't print to screen
+	if err := helpers.RunCommand(false, "bsdiff", original, newfile, dotfile); err != nil {
```
phmccarty: One of the most common errors to hit here is "delta file already exists". Perhaps check for this condition before calling bsdiff? Until I fix bsdiff to not return -1 for every possible error condition (there are many), I think the existence check would be worth adding.

phmccarty: Ignore my comment... You already check for existence elsewhere.

swupd/delta.go
tmarcu: This makes these lines incredibly ugly and longer, but sure.

tmarcu: -1

helpers/helpers.go
tmarcu: I'll leave it to one of them to convince me that's more useful to do.

swupd/delta.go
tmarcu: That is indeed what is passed in.

swupd/delta.go
matthewrsj: ¯\\\_(ツ)\_/¯

swupd/delta.go
matthewrsj: I imagine the caller might want to print out or log how many deltas were created successfully after calling this step.

swupd/delta.go
icarus-sparry: It is this kind of thing which makes me think that we should change the flags to mean

1. Deleted. Can't expect this to be on the system.
1. Ghosted. If set on a deleted file then swupd client doesn't delete the file

If the deleted flag is not set then the ghosted flag is never checked.

helpers/helpers.go
icarus-sparry: The main reason is documentation. Sure your ide can tell you that the first argument for RunCommand is called 'print', but if you are reading code on github and come across 

    if err := helpers.RunCommand(false, "bsdiff", original, newfile, dotfile); err != nil {

you have to know what the parameter means. Does the false mean that cmd.Stdin should be connected to /dev/null rather than os.Stdin? In contrast

    if err := helpers.RunCommandSilent("bsdiff", original, newfile, dotfile); err != nil {

is about the same length to type (less with autocomplete), and I think is easier to understand.

swupd/delta.go
```diff
@@ -0,0 +1,164 @@
...
+	var c config
+	c, err := getConfig(statedir)
+	if err != nil {
+		return nil, err
+	}
+
+	var oldManifest *Manifest
+	var newManifest *Manifest
+
+	if oldManifest, err = ParseManifestFile(filepath.Join(c.outputDir, string(from), manifest.Name)); err != nil {
```
matthewrsj: Usually (if read from disk with `createManifestRecord()`) `manifest.Name` will not contain the `Manifest.` part of the name but will only be, for instance, `full` or `os-core` or the like. This isn't enforced so if `manifest.Name` is set to the full file name before calling this function it will be fine, but we should probably document via code comments in manifest.go what that field is expected to contain, then follow that for consistency.

swupd/delta.go
matthewrsj: So, `file` came from `manifest.Files` and can only be `<= manifest.Header.Version` right? Do you mean to compare against the old manifest version here?

swupd/delta.go
```diff
@@ -0,0 +1,164 @@
...
+			file.DeltaPeer == nil ||
+			file.Type != typeFile ||
+			file.DeltaPeer.Type != typeFile {
+			continue
+		}
+
+		deltadir := filepath.Join(c.outputDir, string(file.Version), "delta")
+		fromToString := fmt.Sprintf("%d-%d-%s-%s", file.DeltaPeer.Version, file.Version, file.DeltaPeer.Hash, file.Hash)
+
+		deltaFile := filepath.Join(deltadir, fromToString)
```
matthewrsj: Thank you 👨‍🚀

swupd/delta.go
```diff
@@ -0,0 +1,164 @@
...
+	if file.Status == statusDeleted {
+		return nil
+	}
+
+	// To files cannot be ghosted
+	if file.Status == statusGhosted {
+		return nil
+	}
+
+	if sameFile(file, file.DeltaPeer) {
```
matthewrsj: @cmarcelo does this mean a fullfile will be added to the pack in this case, since there is no delta created for this?

swupd/delta.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil
+	}
+
+	newfile := filepath.Join(c.imageBase, string(file.Version), "full"+file.Name)
```
matthewrsj: Does `string(uint32)` work? I've had to do `fmt.Sprint(uint32)` because casting directly didn't work. This wasn't caught by go build, but was caught by tests.

swupd/delta.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/clearlinux/mixer-tools/helpers"
+	"github.com/pkg/errors"
+)
+
+// CreateDeltas creates all delta files between the previous and current
+// version of the supplied manifest, returning a list of files that
+// delta creation failed for.
+func CreateDeltas(manifest *Manifest, statedir string, from, to uint32) ([]*File, error) {
+	var c config
+	c, err := getConfig(statedir)
```
icarus-sparry: Seems we will be reading the config a LOT of times.

Can we just pass the config in? Probably help with unit testing as well.

swupd/delta.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+
+	}
+
+	return files, nil
+}
+
+func createDelta(file *File, statedir string, fromVersion uint32, fromHash Hashval) error {
+	var c config
+	c, err := getConfig(statedir)
```
icarus-sparry: Oops, here is where I meant to comment we read config a lot

swupd/delta.go
icarus-sparry: Oops, wrong place.

matthewrsj: That might make sense, could pass it in to `CreateManifests` as well.

swupd/delta.go
matthewrsj: Ohhh, yes this would read config a ton.

icarus-sparry: I am also thinking that perhaps getConfig should return (*config, error) to reduce the amount of things that need to be copied around?

swupd/delta.go
tmarcu: If this is used for Manifest.Full it will only be read once, otherwise yes..it would be called a lot. I can change this to pass in regardless though.

swupd/delta.go
tmarcu: Ah, should have scrolled more, I agree.

swupd/delta.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, err
+	}
+
+	// Must be sorted before passing into linkPeersAndChange()
+	oldManifest.sortFilesName()
+	newManifest.sortFilesName()
+
+	_, _, _ = newManifest.linkPeersAndChange(oldManifest, from)
+
+	deltas, err := consolidateDeltaFiles(newManifest, from, c)
```
matthewrsj: Nitpick:

You aren't being consistent about how you are assigning to `err`. First you do
```go
c, err := getConfig(statedir)
```
so at this point err is declared and in scope. Then you do
```go
if oldManifest, err = ParseManifestFile(...
```
so you expect it to be in scope because of the missing `:`. But then here you are shadowing `err` to declare a new one with `:=`

swupd/delta_test.go
```diff
@@ -0,0 +1,26 @@
+package swupd
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+var largeString = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque tempor quam id convallis convallis. Proin vehicula nisi in augue posuere lobortis. Quisque tincidunt elit ac facilisis auctor. Praesent facilisis ex eros, nec blandit dui suscipit porta. Nunc id nunc rhoncus, condimentum nibh at, fermentum sem. Nam mollis justo ac iaculis gravida. Vestibulum convallis congue dolor, vitae rutrum ex finibus vel. Phasellus convallis sem nunc, ac laoreet risus rhoncus fermentum. Interdum et malesuada fames ac ante ipsum primis in faucibus. Nullam eget pellentesque nulla. Etiam tristique non magna quis consectetur. In ac dolor sagittis, vulputate ante tincidunt, suscipit leo."
```
matthewrsj: Another nitpick:

Can you line-wrap this? I just think it looks better in editors to have something like
```go
var largeString = `something really long that
spans multiple lines instead of one really long
line that some editors treat differently than
others`
```

tmarcu: So you can read the latin more easily? *puts cheeky comment away* :grinning:

matthewrsj: Who says I can't read latin?

swupd/delta.go
matthewrsj: This still needs to be fixed.

swupd/delta_test.go
icarus-sparry: As someone who studied latin at school for 3 years before I was allowed to drop it, I prefer `something really long that...' as it makes sense, unlike random text designed to show off typefaces. Heck even `A quick brown fox jumps over the lazy dog` and related [pangrams](https://en.wikipedia.org/wiki/Pangram) make more sense.

Of course it wasn't all wasted, it makes me appreciate [Asterix)[https://en.wikipedia.org/wiki/List_of_Asterix_volumes] more.

swupd/delta.go
```diff
@@ -0,0 +1,186 @@
...
+
+	for _, file := range deltas {
+		deltaQueue <- file
+	}
+
+	close(deltaQueue)
+	wg.Wait()
+	close(failedChan)
+
+	for f := range failedChan {
```
icarus-sparry: Hang on, you just closed the failedChan.

swupd/delta_test.go
bryteise: Hrm being able to enjoy Asterix more than I have already, I'm unable to imagine such a fine world.

swupd/delta.go
icarus-sparry: @matthewrsj Not true. The err here is the one declared at line 19. The `:=` has to declare at least one new variable, but it doesn't declare all of them to be new.  So this line creates a new variable `deltas` but reuses the existing err as it is in the same lexical block. See [playground](https://play.golang.org/p/AP9XSvOz4VX) for a poor example. See section 2.3.1 of the Go Programming Language, starting "One subtle but important point" (page 31 in my copy.

swupd/delta.go
```diff
@@ -0,0 +1,186 @@
...
+
+	const maxRoutines = 3
+	var failed []*File
+	var deltaQueue = make(chan *File)
+	var failedChan = make(chan *File)
+	var wg sync.WaitGroup
+	wg.Add(maxRoutines)
+
+	for i := 0; i < maxRoutines; i++ {
+		go func(statedir string) {
```
icarus-sparry: Without having tried it, I suspect you can remove statedir from here and line 62, Then line 57 could pick it up from line 16. There are lots of good reasons for passing in things explicitly, in particular if you think they are going to change during the lifetime of the goroutine, but I don't think this is one of them.

swupd/delta.go
matthewrsj: Ah thanks @icarus-sparry, TIL.

matthewrsj:
BTW, congrats on PR/Issue #100! 💯 🎂 🎁

helpers/helpers.go
```diff
@@ -237,3 +238,22 @@ func Git(args ...string) {
...
+	cmd := exec.Command(cmdname, args...)
+	if print {
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+	} else {
+		// Automatically connected to os.DevNull by implementation
+		cmd.Stdout = nil
+		cmd.Stderr = nil
```
cmarcelo: These are the default values for the exec.Cmd struct returned from exec.Command. So you can likely remove the else branch.

swupd/delta.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	var oldManifest *Manifest
+	var newManifest *Manifest
+
+	if oldManifest, err = ParseManifestFile(filepath.Join(c.outputDir, fmt.Sprintf("%d", from), manifest)); err != nil {
+		return nil, err
+	}
+	if newManifest, err = ParseManifestFile(filepath.Join(c.outputDir, fmt.Sprintf("%d", to), manifest)); err != nil {
+		return nil, err
```
cmarcelo: Just a heads up: during integration with packs, I'll try to make this function (or a helper function) take the Manifest structs directly, since we already have all of them parsed. I might have to reset internal state used in linkPeers... before reusing though. (Let me know if this is a bad idea) No action needed for this PR.

swupd/delta.go
```diff
@@ -0,0 +1,190 @@
...
+			continue
+		}
+		if file.DeltaPeer == nil {
+			continue
+		}
+		if file.Type != TypeFile ||
+			file.DeltaPeer.Type != TypeFile {
+			continue
+		}
+
```
cmarcelo: Should the check for Status Deleted/Ghosted and the check whether is same file happen here? Otherwise we are generating a list of files to create deltas that has files that might not need delta at all.

swupd/delta.go
```diff
@@ -0,0 +1,190 @@
...
+	// By definition we cannot make deltas for files that are deleted
+	if file.Status == StatusDeleted {
+		return nil
+	}
+
+	// To files cannot be ghosted
+	if file.Status == StatusGhosted {
+		return nil
+	}
```
cmarcelo: Given this is a helper function for creating deltas we already identified (so, not really meant to be used in isolation), ideally we shouldn't need to perform those checks again.

If you must perform those checks, move the checks to a function and call the same checks from here and consolidateDeltas.

swupd/delta.go
```diff
@@ -0,0 +1,190 @@
...
+	deltadir := filepath.Join(c.outputDir, fmt.Sprintf("%d", file.Version), "delta")
+	fromToString := fmt.Sprintf("%d-%d-%s-%s", fromVersion, file.Version, fromHash, file.Hash)
+
+	// Files to create and validate deltas with
+	deltafile := filepath.Join(deltadir, fromToString)
+	testnewfile := filepath.Join(deltadir, "."+fromToString+".testnewfile")
+
+	// Shell out to bsdiff for now...false = don't print to screen
+	if err := helpers.RunCommand(false, "bsdiff", original, newfile, deltafile); err != nil {
+		_ = os.Remove(deltafile) // Might have retuned FULLDL
```
cmarcelo: Typo "retuned". Also comment is unclear, what is FULLDL?

phmccarty: The Clear Linux BSDiff returns `1` as a hint to swupd-server that the delta it created is larger than the new file created after applying that delta. We call that hint FULLDL ("full download"), and swupd-server will make the fullfile (instead of the delta) available for download instead by adding it to the packs.

For reference, see https://github.com/clearlinux/swupd-server/blob/3cf2ade205cd4ef3189929655b72173752360f83/src/delta.c#L80

cmarcelo: Thanks for the explanation. I suggest a short version of it to be in a comment in this code.

swupd/delta.go
matthewrsj: The check shouldn't actually be needed anymore. If you look at the latest `linkPeersAndChange` code delta peers will only be set on files with the same name that are not the same file and (I believe) are not ghosted or deleted.

swupd/delta.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// We only support deltas between two regular files for now
+	// TODO: Support directory -> file in the future, this is a complex case
+	if file.Type != TypeFile || file.DeltaPeer.Type != TypeFile {
+		return nil
+	}
+
+	// By definition we cannot make deltas for files that are deleted
+	if file.Status == StatusDeleted {
```
matthewrsj: This check is no longer needed, since `linkPeersAndChange` only sets `DeltaPeer` for existing files.

swupd/delta.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil
+	}
+
+	// To files cannot be ghosted
+	if file.Status == StatusGhosted {
```
matthewrsj: This check is no longer needed, since `linkPeersAndChange` only sets `DeltaPeer` for existing files.

swupd/delta.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil
+	}
+
+	if sameFile(file, file.DeltaPeer) {
```
matthewrsj: This check is no longer needed, since `linkPeersAndChange` only sets `DeltaPeer` for changed files.

helpers/helpers.go
tmarcu: Ah I knew this from looking at another PR of yours long time ago... ty for pointing this out again!

swupd/helpers_test.go
```diff
@@ -250,3 +250,10 @@ func checkManifestMatches(t *testing.T, testDir, ver, name string, res ...*regex
 		}
 	}
 }
+
+func mustCreateDeltas(t *testing.T, manifest, statedir string, from, to uint32) {
+	failed, err := CreateDeltas(manifest, statedir, from, to)
+	if err != nil || len(failed) > 0 {
```
icarus-sparry: This seems to be mixing two cases, 'err != nil' means one thing and 'len(failed)>0'.
Probably need two different print statements for the two cases.

swupd/delta_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+func TestCreateDeltas(t *testing.T) {
+	testDir := mustSetupTestDir(t, "deltas")
+	//defer removeIfNoErrors(t, testDir)
```
matthewrsj: Need to uncomment this now.

swupd/delta.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil
+	}
+
+	// We only support deltas between two regular files for now
+	// TODO: Support directory -> file in the future, this is a complex case
+	if file.Type != TypeFile || file.DeltaPeer.Type != TypeFile {
+		return nil
+	}
+
+	newfile := filepath.Join(c.imageBase, fmt.Sprintf("%d", file.Version), "full"+file.Name)
```
matthewrsj: Ohhh, I get it now. I was having a heck of a time figuring out what `"full"+file.Name` was doing because I forgot that `file.Name` would have a leading `/`. This is precisely why `filepath.Join` is useful. Please change that last argument to `(..., "full", file.Name)`

swupd/delta_test.go
tmarcu: heh, thanks.

swupd/delta.go
```diff
@@ -0,0 +1,179 @@
...
+
+		deltadir := filepath.Join(c.outputDir, fmt.Sprintf("%d", file.Version), "delta")
+		fromToString := fmt.Sprintf("%d-%d-%s-%s", file.DeltaPeer.Version, file.Version, file.DeltaPeer.Hash, file.Hash)
+
+		deltaFile := filepath.Join(deltadir, fromToString)
+
+		// Only add to list if the delta file does not already exist on the system
+		if _, err := os.Stat(deltaFile); os.IsNotExist(err) {
+			// Only create deltas for files bigger than 200 bytes
+			fullname := filepath.Join(c.imageBase, fmt.Sprintf("%d", file.Version), "full"+file.Name)
```
matthewrsj: Sorry, one more "full"+file.Name

Issue #53, opened 2017-12-23
Feature: Bundle lint tool


kevincwells:
# New Feature
Add a new command ("mixer bundle check") to lint and validate a bundle definition file.

# Use Case
If a developer wants to write their own bundle file, or alter an existing one, it would be useful to be able to verify that it is correct.

# Functionality
- The tool will at minimum "lint" the bundle file, ensuring that it is correctly formatted.
- The tool could validate that fields are not empty.
- The tool could validate that packages/bundles mentioned in the file actually exist.
  - Bundles seems easy, packages seems harder.

# Notes
- While this is being proposed as a stand-alone call, the intent is to incorporate this linting as an automatic (but suppressible) step for any other commands that involve creating or editing bundles.
- The logic of this would need to be rewritten if the bundle definition format changes (e.g., to a TOML file), but such changes would actually significantly simplify the logic needed to lint/parse.

kevincwells:
Most of this is implemented by #177.

What's not in that PR is checking validity of the bundles and packages listed in the bundle.
Checking packages would be pretty easy - calling `getBundlePath` on each included bundle would do the trick. Not entirely sure where to look to validate packages. @matthewrsj probably knows.

Both of these can be incremental patches.

tmarcu:
This issue is fixed, we can open another for packages.

PR #102, opened 2018-01-19
builder: port chroot building to Go

### LINKED ISSUE #42: proposal: implement bundle-chroot-builder functionality inside mixer
AFAIK mixer is the only user of the Python script. Both mixer and and bundle-chroot-builder read the same configuration file, which the template value provided to users come from bundle-chroot-builder but might include some mixer specific values.

The expected result is: avoid duplication of configuration reading code, making easy to maintain the default configuration for mixer, making easy to debug mixer (one less external script in the workflow).

It is always possible to provide an extra binary later if the functionality is needed by itself without mixer.

cmarcelo:
This rewrites bundle-chroot-builder.py in Go to be part of Mixer
code. Mixer is the only user of that software, and both Mixer and
b-c-b.py the same configuration file, with overlapping fields.

Main differences from bundle-chroot-builder.py:

- New bundleset type was added, that cares about collecting as much
  information as possible from the bundles themselves. This type and
  related functions also sets us up for success when upcoming changes
  to how bundles are specified happen. There is no assumption all
  bundles are in the same directory.

- We are not using m4, instead a bundleset takes care of parsing. If
  format of individual bundle files change. The upside is that we can
  give nicer error messages, specially for the circular case.

- Read the configuration file directly (with go-ini) to peek at values
  that Mixer didn't read before. Done that to avoid conflicting with
  existing patch in-flight that parses configuration.

- Some individual steps were reordered for code clarity. E.g.: since
  we have bundleset, we can upfront generate all the *-include files.

- Fixed the output for versions file. Due to the way yum list output
  works, parsing it is not very friendly. Comments around the code
  tells the story.

- Removed the network testing step. It wasn't covering every case in
  the Python version, so I'm leaning to let the failure come from
  yum/dnf itself. I'm usually in favor of such early tests, but in
  this case the price of parsing yet another config file didn't felt
  worth.

- Removed the "yum clean all" step from the bootstrap. There isn't any
  cache at that point, and the next yum call will bootstrap the
  necessary files for yum to operate.

- Removed generation of files-* files (and the pkgmap-* files used to
  generate them). I couldn't find any tool or team making use of this
  information. Those (or their content) might be relevant in future
  changes to use a single chroot, but we should add when we need them.

- Added more detailed commentary to individual steps, collecting
  information from the developers of bcb and related software.

- The port still don't parallelize the work into multiple
  goroutines. I plan to do this in a similar way than what was done in
  CreateFullfiles, but in a separated patch.

Fixes #42.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -42,6 +42,10 @@ const Version = "3.2.1"
 // swupd-server (package swupd) when possible. This is an experimental feature.
 var UseNewSwupdServer = false
 
+// UseNewChrootBuilder controls whether to use the new implementation of
+// building the chroots. This is an experimental feature.
+var UseNewChrootBuilder = false
+
 // A Builder contains all configurable fields required to perform a full mix
 // operation, and is used to encapsulate life time data.
 type Builder struct {
@@ -557,19 +561,56 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
 		}
 	}
 
-	// If this is a mix, we need to build with the Clear version, but publish the mix version
-	chrootcmd := exec.Command(b.Buildscript, "-c", b.Buildconf, "-m", b.Mixver, b.Clearver)
-	chrootcmd.Stdout = os.Stdout
-	chrootcmd.Stderr = os.Stderr
-	err := chrootcmd.Run()
-	if err != nil {
-		helpers.PrintError(err)
-		return err
+	if UseNewChrootBuilder {
+		// TODO: Fix this output when we take bundles from multiple places.
+		fmt.Printf("Reading bundles from: %s\n", b.Bundledir)
+		files, err := ioutil.ReadDir(b.Bundledir)
+		if err != nil {
+			return err
+		}
+		var bundleFiles []string
+		for _, file := range files {
+			if file.IsDir() || strings.HasPrefix(file.Name(), ".") {
+				continue
+			}
+			bundleFiles = append(bundleFiles, filepath.Join(b.Bundledir, file.Name()))
+		}
+		var set *bundleSet
+		set, err = newBundleSet(bundleFiles)
+		if err != nil {
+			return err
+		}
+		// TODO: Merge the rest of this function into buildBundleChroots (or vice-versa).
+		err = b.buildBundleChroots(set)
+		if err != nil {
+			return err
+		}
+
+		// TODO: Move this logic to code that uses this.
+		// If LAST_VER don't exists, it means this is the first chroot we build,
+		// so initialize it to version "0".
+		lastVerPath := filepath.Join(b.Statedir, "image", "LAST_VER")
+		if _, err = os.Stat(lastVerPath); os.IsNotExist(err) {
+			err = ioutil.WriteFile(lastVerPath, []byte("0\n"), 0644)
+			if err != nil {
+				return err
+			}
+		}
+
+	} else {
+		// If this is a mix, we need to build with the Clear version, but publish the mix version
+		chrootcmd := exec.Command(b.Buildscript, "-c", b.Buildconf, "-m", b.Mixver, b.Clearver)
+		chrootcmd.Stdout = os.Stdout
+		chrootcmd.Stderr = os.Stderr
+		err := chrootcmd.Run()
+		if err != nil {
+			return err
+		}
 	}
 
 	// Generate the certificate needed for signing verification if it does not exist and insert it into the chroot
 	if signflag == false && template != nil {
-		err = helpers.GenerateCertificate(b.Cert, template, template, &privkey.PublicKey, privkey)
+		err := helpers.GenerateCertificate(b.Cert, template, template, &privkey.PublicKey, privkey)
 		if err != nil {
 			return err
 		}
--- a/builder/bundleset.go
+++ b/builder/bundleset.go
@@ -1,0 +1,188 @@
+package builder
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+)
+
+var (
+	validBundleNameRegex  = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
+	validPackageNameRegex = regexp.MustCompile(`^[A-Za-z0-9-_+.]+$`)
+)
+
+type bundle struct {
+	Name     string
+	Filename string
+
+	DirectIncludes []string
+	DirectPackages []string
+	AllPackages    []string
+}
+
+type bundleSet struct {
+	Bundles map[string]*bundle
+}
+
+// newBundleSet will take a list of files and create a bundle struct from the list of
+// files, doing all the required validation and processing.
+//
+// It currently implements parsing of the bundle files, ignoring comments and processing
+// "include()" directives the same way that m4 works.
+func newBundleSet(bundleFiles []string) (*bundleSet, error) {
+	bundles := make(map[string]*bundle, len(bundleFiles))
+
+	// First parse each bundle file in isolation.
+	for _, filename := range bundleFiles {
+		name := filepath.Base(filename)
+		b, exists := bundles[name]
+		if exists {
+			return nil, fmt.Errorf("bundle %q defined twice: in %s and %s", name, b.Filename, filename)
+		}
+		if !validBundleNameRegex.MatchString(name) {
+			return nil, fmt.Errorf("invalid bundle name %q derived from file %s", name, filename)
+		}
+
+		contents, err := ioutil.ReadFile(filename)
+		if err != nil {
+			return nil, err
+		}
+
+		includes, packages, err := parseBundle(contents)
+		if err != nil {
+			return nil, fmt.Errorf("couldn't parse bundle file %s: %s", filename, err)
+		}
+
+		b = &bundle{
+			Name:           name,
+			DirectIncludes: includes,
+			DirectPackages: packages,
+			Filename:       filename,
+		}
+		bundles[name] = b
+	}
+
+	// Then check if there are missing includes.
+	for _, b := range bundles {
+		for _, include := range b.DirectIncludes {
+			_, exists := bundles[include]
+			if !exists {
+				return nil, fmt.Errorf("bundle %q includes bundle %q which is not available", b.Name, include)
+			}
+		}
+	}
+
+	// And finally sort the bundles so that all includes appear before a bundle, then
+	// calculate AllPackages for each bundle. Cycles are identified as part of sorting
+	// the bundles.
+	sortedBundles, err := sortBundles(bundles)
+	if err != nil {
+		return nil, err
+	}
+	for _, b := range sortedBundles {
+		var allPackages []string
+		allPackages = append(allPackages, b.DirectPackages...)
+		for _, include := range b.DirectIncludes {
+			allPackages = append(allPackages, bundles[include].AllPackages...)
+		}
+		// Remove redundant packages.
+		sort.Strings(allPackages)
+		for i, p := range allPackages {
+			if i != 0 && allPackages[i-1] == p {
+				continue
+			}
+			b.AllPackages = append(b.AllPackages, p)
+		}
+	}
+
+	return &bundleSet{Bundles: bundles}, nil
+}
+
+func sortBundles(bundles map[string]*bundle) ([]*bundle, error) {
+	// Traverse all the bundles, recursing to mark its included bundles as visited before mark
+	// the bundle itself as visited. VISITING state is used to identify cycles.
+	type state int
+	const (
+		NotVisited state = iota
+		Visiting
+		Visited
+	)
+	mark := make(map[*bundle]state, len(bundles))
+	sorted := make([]*bundle, 0, len(bundles))
+	visiting := make([]string, 0, len(bundles)) // Used to produce nice error messages.
+
+	var visit func(b *bundle) error
+	visit = func(b *bundle) error {
+		switch mark[b] {
+		case Visiting:
+			return fmt.Errorf("cycle found in bundles: %s -> %s", strings.Join(visiting, " -> "), b.Name)
+		case NotVisited:
+			mark[b] = Visiting
+			visiting = append(visiting, b.Name)
+			for _, inc := range b.DirectIncludes {
+				err := visit(bundles[inc])
+				if err != nil {
+					return err
+				}
+			}
+			visiting = visiting[:len(visiting)-1]
+			mark[b] = Visited
+			sorted = append(sorted, b)
+		}
+		return nil
+	}
+
+	for _, b := range bundles {
+		err := visit(b)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return sorted, nil
+
+}
+
+func parseBundle(contents []byte) (includes []string, packages []string, err error) {
+	scanner := bufio.NewScanner(bytes.NewReader(contents))
+
+	line := 0
+	for scanner.Scan() {
+		line++
+		text := scanner.Text()
+		comment := strings.Index(text, "#")
+		if comment > -1 {
+			text = text[:comment]
+		}
+		text = strings.TrimSpace(text)
+		if len(text) == 0 {
+			continue
+		}
+		if strings.HasPrefix(text, "include(") {
+			if !strings.HasSuffix(text, ")") {
+				return nil, nil, fmt.Errorf("missing end parenthesis in line %d: %q", line, text)
+			}
+			text = text[8 : len(text)-1]
+			if !validBundleNameRegex.MatchString(text) {
+				return nil, nil, fmt.Errorf("invalid bundle name %q in line %d", text, line)
+			}
+			includes = append(includes, text)
+		} else {
+			if !validPackageNameRegex.MatchString(text) {
+				return nil, nil, fmt.Errorf("invalid package name %q in line %d", text, line)
+			}
+			packages = append(packages, text)
+		}
+	}
+
+	if scanner.Err() != nil {
+		return nil, nil, scanner.Err()
+	}
+
+	return includes, packages, nil
+}
--- a/builder/bundleset_test.go
+++ b/builder/bundleset_test.go
@@ -1,0 +1,192 @@
+package builder
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+func TestParseBundle(t *testing.T) {
+	tests := []struct {
+		Contents         []byte
+		ExpectedIncludes []string
+		ExpectedPackages []string
+		ShouldFail       bool
+	}{
+		{
+			Contents: []byte(`# Simple fake bundle
+include(a)
+include(b)
+pkg1     # Comment
+pkg2
+`),
+			ExpectedIncludes: []string{"a", "b"},
+			ExpectedPackages: []string{"pkg1", "pkg2"},
+		},
+
+		// Error cases.
+		{Contents: []byte(`include(`), ShouldFail: true},
+		{Contents: []byte(`()`), ShouldFail: true},
+		{Contents: []byte(`Include(`), ShouldFail: true},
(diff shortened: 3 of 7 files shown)


cmarcelo:
I've tested this by generating the chroots with a few bundles with --new-chroots and without, and comparing the md5sum of all the files. The only differences (all expected) were the versionstamp, versions file (this port fixes its output), and the presence of files-* files (not part of the port).

I haven't tested it in the "build clear linux itself" mode yet.

cmarcelo:
@phmccarty @tmarcu could you take a look?

cmarcelo:
@kevincwells could you take a look of how bundleset is being used? It let us pass files from different directories, so I think will play nice with the upcoming work you are doing. Instead of doing the ReadDir in the bundles directory, we'll call your functions to decide whether bundle.

Note that I've seen GetIncludedBundles, but my parsing was doing more than that (taking the packages), so I've opted out of it. I haven't looked how bundleset could benefit the code of AddBundles yet.

kevincwells:
@cmarcelo I finished up my bundle refactor and pushed it today (#107). It's designed to go as-is, generating the `mix-bundles` list on-the-fly and assuming this BCB rewrite doesn't exist.

I haven't had the time to look at your implementation yet, but I will tomorrow. I'm betting it'll be relatively easy to merge our two changes, so the `mix-bundles` dir is removed completely.

builder/bundleset.go
```diff
@@ -0,0 +1,188 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return sorted, nil
+
+}
+
+func parseBundle(contents []byte) (includes []string, packages []string, err error) {
```
kevincwells: This is very close to what I want for a `mixer bundle lint` or `mixer bundle check` command, as per #53. I would love to reuse this there.

The only addition I want for #53 is checks that the `# [TITLE]: <title>` and `# [DESCRIPTION]: <description>` fields are not empty. Maybe others too, but those for sure. I know your code doesn't care about those fields, though.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+		return nil, err
+	}
+
+	return cfg, nil
+}
+
+func (b *Builder) buildBundleChroots(set *bundleSet) error {
+	var err error
+	stateDir := b.Statedir
+	chrootDir := filepath.Join(stateDir, "image")
```
tmarcu: It'd be nice to have b.ChrootDir and whatever else is needed setup from the beginning wouldn't it? Out of scope for this PR but reading through made me think of it.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+
+	// TODO: Remove remaining references to outputDir. Let "build update" take care of
+	// bootstraping or cleaning up.
+	outputDir := filepath.Join(stateDir, "www")
+
+	if stateDir == "" || stateDir == "/" {
```
tmarcu: Why is "/" disallowed?

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+	// Bootstrap the directories.
+	err = os.MkdirAll(filepath.Join(chrootDir, "0"), 0755)
+	if err != nil {
+		return err
+	}
+	err = os.MkdirAll(filepath.Join(outputDir, "0"), 0755)
+	if err != nil {
+		return err
+	}
+	// TODO: Move this outside this function. chroot building doesn't use LAST_VER.
```
tmarcu: Why not move it out now and call it only after building finishes?

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+	var version string
+	if b.Mixver != "" {
+		fmt.Printf("Creating chroots for version %s based on Clear Linux %s\n", b.Mixver, b.Clearver)
+		version = b.Mixver
+	} else {
+		fmt.Printf("Creating chroots for version %s\n", b.Clearver)
+		version = b.Clearver
+	}
+	// TODO: This validation should happen when reading the configuration.
+	if version == "" {
```
tmarcu: Could add this in the else case above and remove from here.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+		return errors.Errorf("no Mixver or Clearver set, unable to proceed")
+	}
+
+	// TODO: Check if this is redundant with the same cleanup in builder.go!
```
tmarcu: Yes, remove :)

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+	chrootVersionDir := filepath.Join(chrootDir, version)
+	fmt.Printf("Removing pre-existing %s before starting\n", chrootVersionDir)
+	err = os.RemoveAll(chrootVersionDir)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Preparing new %s\n", chrootVersionDir)
+	// TODO: Fix this output when we take bundles from multiple places.
+	fmt.Printf("  based on bundles from: %s\n", b.Bundledir)
```
tmarcu: This should read from .mixer/mixbundles right @kevincwells ? (once the refactor PR is in)

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+	fmt.Printf("  and yum config: %s\n", b.Yumconf)
+
+	err = os.MkdirAll(chrootVersionDir, 0755)
+	if err != nil {
+		return err
+	}
+	for name, bundle := range set.Bundles {
+		// TODO: Should we embed this information in groups.ini? (Maybe rename it to bundles.ini)
```
tmarcu: Yes, the groups.ini should be removed in the future and replaced with a bundles toml similar, which hosts all the information.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+	}
+
+	err = fixOSRelease(filepath.Join(osCoreDir, "usr/lib/os-release"), version)
+	if err != nil {
+		return errors.Wrap(err, "couldn't fix os-release file")
+	}
+
+	fmt.Println("Creating chroots for bundles")
+	// TODO: Use goroutines.
+	for _, bundle := range set.Bundles {
```
tmarcu: You can likely just throw each bundle install into a goroutine without any complicated waitgroup limiting, there's not much processing here just IO.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+func fixOSRelease(filename, version string) error {
+	f, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = f.Close()
+	}()
+
+	// TODO: If this is a mix, NAME and ID should probably change too. Create a section in
```
tmarcu: I think this is a good opportunity to update the os-release file to be more descriptive if it's on a mix, updating the name & adding a UpstreamVersion field perhaps.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+	if err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(filepath.Join(chrootVersionDir, "packages-"+bundle.Name), packages.Bytes(), 0644)
+}
+
+// run executes args storing its output in memory. If the command succeeds returns that output, if
+// it fails, return err that contains both the out and err streams from the execution.
+func run(args ...string) (*bytes.Buffer, error) {
```
tmarcu: Put this in helpers.go where RunCommand and RunCommandSilent are? Or update one/both of them to do what you need here? Don't want 3 functions doing essentially the same thing in different files.

builder/bundleset_test.go
```diff
@@ -0,0 +1,195 @@
...
+				"d": 13,
+			},
+		},
+
+		{
+			"all packages don't have duplicates",
+			FilesMap{
+				"a": Lines("A A A A"),
+				"b": Lines("include(a) A A A A"),
+				"c": Lines("include(b) A A A A include(a)"),
```
matthewrsj: are we calling this valid? I guess I don't see why not...
```
include(b)
A
A
A
A
include(a)
```

builder/bundleset_test.go
```diff
@@ -0,0 +1,195 @@
...
+				t.Fatalf("couldn't create temporary file for test case: %s", err)
+			}
+			bundleFiles = append(bundleFiles, bundleFile)
+		}
+
+		expectedError := (len(tt.ExpectedAllPackageCounts) == 0)
+
+		var set *bundleSet
+		set, err = newBundleSet(bundleFiles)
+		if err != nil {
```
matthewrsj: You could simplify this as I mentioned in an above comment.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return cfg, nil
+}
+
+func (b *Builder) buildBundleChroots(set *bundleSet) error {
+	var err error
+	stateDir := b.Statedir
```
matthewrsj: Any reason you create a new variable here? I don't think it is very expensive to just access the struct member every time you want to use it and stateDir only saves you two characters over b.Statedir.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+	// TODO: Move this outside this function. chroot building doesn't use LAST_VER.
+	lastVerPath := filepath.Join(chrootDir, "LAST_VER")
+	if _, err = os.Stat(lastVerPath); os.IsNotExist(err) {
+		err = ioutil.WriteFile(lastVerPath, []byte("0\n"), 0644)
+		if err != nil {
+			return err
+		}
+	}
+
+	// TODO: To not touch config code that is in flux at the moment, reparsing it here to grab
```
matthewrsj: s/To/Do/ ?

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+		return err
+	}
+
+	if _, ok := set.Bundles[cfg.UpdateBundle]; !ok {
+		return fmt.Errorf("couldn't find bundle %q specified in configuration as the update bundle", cfg.UpdateBundle)
+	}
+
+	// Write INI files. These are used to communicate to the next step of mixing (build update).
+	var serverINI bytes.Buffer
+	fmt.Fprintf(&serverINI, `[Server]
```
matthewrsj: This is a good use case for a template.

builder/chroots.go
matthewrsj: This should definitely be de-duplicated.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+		return err
+	}
+
+	yumCmd := []string{
+		"yum",
+		"--config=" + b.Yumconf,
+		"-y",
+		"--releasever=" + b.Clearver,
+	}
+	// When in Fedora, call dnf instead of yum.
```
matthewrsj: I know this was in the original version, but we should consider just deprecating yum. Clear Linux is headed in that direction already and dnf is available.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+		return err
+	}
+
+	// Writing special files identifying the version in os-core.
+	err = ioutil.WriteFile(filepath.Join(clearDir, "version"), []byte(version), 0644)
+	if err != nil {
+		return err
+	}
+	// TODO: This seems to be the only thing that makes two consecutive chroots of the same
+	// version to be different. Use SOURCE_DATE_EPOCH if available?
```
matthewrsj: SOURCE_DATE_EPOCH would be a good change as it is used in Clear Linux package builds. This was also discussed as a change for swupd-server.

builder/chroots.go
```diff
@@ -0,0 +1,637 @@
...
+
+	fmt.Println("Creating chroots for bundles")
+	// TODO: Use goroutines.
+	for _, bundle := range set.Bundles {
+		if bundle.Name == "os-core" {
+			continue
+		}
+
+		fmt.Printf("Creating %s chroot\n", bundle.Name)
+		_, err = run("cp", "-a", "--preserve=all", osCoreDir, filepath.Join(chrootVersionDir, bundle.Name))
```
matthewrsj: Is there not a way to do this without shelling out to `cp`?

matthewrsj: quick research shows you would have to implement your own recursive copy... never mind.

builder/chroots.go
matthewrsj: Maybe a better question for @bryteise

builder/builder.go
```diff
@@ -42,6 +42,10 @@ const Version = "3.2.1"
 // swupd-server (package swupd) when possible. This is an experimental feature.
 var UseNewSwupdServer = false
 
+// UseNewChroots controls whether to use the new implementation of
+// building the chroots. This is an experimental feature.
+var UseNewChroots = false
```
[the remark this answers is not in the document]
cmarcelo: Done.

builder/chroots.go
cmarcelo: OK.

builder/chroots.go
cmarcelo: This was just some paranoid version of me wanting to prevent some invalid input. Removed.

cmarcelo: This is the kind of sanity check I would want configuration reading to deal with, as marked in another TODO.

builder/chroots.go
cmarcelo: Moved outside, but added a TODO to consider moving it to the bundle update logic.

Not doing that other move now to contain the amount of change (don't want to touch "build update" in this patch).

builder/chroots.go
cmarcelo: Done.

builder/chroots.go
cmarcelo: Done.

builder/chroots.go
cmarcelo: Removed this print and added a print to the place we read the bundles directory. This will make the fix suggested in TODO more straightforward later.

I think the idea when Kevin changes get in is that this would not output the transient bundles directory, but some information about the original location of the bundles ("using these and these local bundles, and these and these remove bundles").

builder/chroots.go
cmarcelo: OK.

builder/chroots.go
cmarcelo: I rather do the parallel thing in a separate patch on top if it is OK.

I suspect that even if the processes will be I/O bound, we might get some slowness from a lot of different I/O competing (may reduce the locality). Maybe doesn't make a difference, but think it is worth testing.

builder/chroots.go
cmarcelo: I agree, but think we can do that in a separate change. Created #113 and assigned to me.

builder/chroots.go
cmarcelo: Done. Make total sense, they just weren't in the codebase when the PR was made. :)

Created RunCommandOutput, which does what the run here does, and implemented RunCommandSilent using that. Added two tests for RunCommandOutput checking the output and error message have what we expect.

builder/builder.go
```diff
@@ -487,19 +491,43 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
+			return err
+		}
+	} else {
+		// If this is a mix, we need to build with the Clear version, but publish the mix version
+		chrootcmd := exec.Command(b.Buildscript, "-c", b.Buildconf, "-m", b.Mixver, b.Clearver)
+		chrootcmd.Stdout = os.Stdout
+		chrootcmd.Stderr = os.Stderr
+		err := chrootcmd.Run()
+		if err != nil {
+			helpers.PrintError(err)
```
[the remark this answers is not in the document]
cmarcelo: Done.

builder/chroots.go
tmarcu: That's fair, let's look at it sooner than later then.

builder/chroots.go
tmarcu: ok +1

builder/bundleset.go
```diff
@@ -0,0 +1,188 @@
...
+var (
+	validBundleNameRegex  = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
+	validPackageNameRegex = regexp.MustCompile(`^[A-Za-z0-9-_+.]+$`)
+)
+
+type bundle struct {
+	Name     string
+	Filename string
+
+	DirectIncludes []string
```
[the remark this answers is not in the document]
cmarcelo: I was thinking about someone reading code that says "bundle.Includes", and there could be ambiguity if this were all the includes or just the direct ones. Since this notion of Direct/All already appears for package I just reused.

builder/bundleset_test.go
```diff
@@ -0,0 +1,195 @@
...
+pkg2
+`),
+			ExpectedIncludes: []string{"a", "b"},
+			ExpectedPackages: []string{"pkg1", "pkg2"},
+		},
+
+		// Error cases.
+		{Contents: []byte(`include(`), ExpectedError: true},
+		{Contents: []byte(`()`), ExpectedError: true},
+		{Contents: []byte(`incl` + `ud(`), ExpectedError: true}, // Hide the misspelling from linter.
```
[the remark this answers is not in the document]
cmarcelo: Done.

builder/bundleset_test.go
tmarcu: Includes could be throw in anywhere, we just like to keep them organized.

builder/bundleset_test.go
```diff
@@ -0,0 +1,195 @@
...
+		{Contents: []byte(`include())`), ExpectedError: true},
+		{Contents: []byte(`include(abc))`), ExpectedError: true},
+	}
+
+	for _, tt := range tests {
+		includes, packages, err := parseBundle(tt.Contents)
+		if err != nil {
```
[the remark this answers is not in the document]
cmarcelo: Done.

Renamed ExpectedError to ShouldFail to make it more evident it is a boolean and added a local variable failed just to make the if condition more readable.

builder/bundleset_test.go
cmarcelo: We could make it invalid later, but at the moment, we play by m4 rules (the program that parses the bundles).

builder/chroots.go
tmarcu: Myself and others have been using dnf without problem for a while AFAIK, let's poke the list and see what people think?

matthewrsj: +1 @tmarcu

builder/bundleset_test.go
```diff
@@ -0,0 +1,195 @@
...
+				continue
+			}
+			t.Errorf("unexpected error parsing bundle: %s\nCONTENTS:\n%s", err, tt.Contents)
+			continue
+		}
+		if tt.ExpectedError {
+			t.Errorf("unexpected success when parsing bundle\nCONTENTS:\n%s\nPARSED INCLUDES: %s\nPARSED PACKAGES:\n%s", tt.Contents, includes, packages)
+			continue
+		}
+		sameIncludes := reflect.DeepEqual(includes, tt.ExpectedIncludes)
```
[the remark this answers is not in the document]
cmarcelo: Done.

builder/bundleset_test.go
cmarcelo: Done, also renaming to shouldFail.

builder/chroots.go
cmarcelo: Done.

builder/chroots.go
cmarcelo: Tone.

builder/chroots.go
cmarcelo: Maybe, but since we are exploring how to use information directly from builder.conf, I have the impression it will go away, so won't take the trouble. If it does stay, we templatize it.

Created https://github.com/clearlinux/mixer-tools/issues/114 to keep track of server.ini's fate.

builder/chroots.go
cmarcelo: Sounds good, but for the sake of this PR, I'll stick to original behavior on this. We later can test and nuke yum usage.

cmarcelo: Recorded in issue https://github.com/clearlinux/mixer-tools/issues/115.

cmarcelo:
@tmarcu @matthewrsj  new version is up.

> Will run this on the larger server for a bigger test.

So, did your tests work?

tmarcu:
@cmarcelo the boot prompt in my booted VM says it worked 👍

Issue #57, opened 2017-12-23
Feature: Bundle List command


kevincwells:
# New Feature
Add a command (`mixer bundle list`) that reports what bundles are in the mix, in the `local-bundles` directory (see #55), or available from Upstream (CLR).

# Notes
- Should support reporting both what top-level bundles are in the mix (i.e., the contents of the Mix Bundles List), as well as the total list of bundles that will end up in the mix by expanding the list recursively following bundle includes.

kevincwells:
Implemented in #137

Issue #55, opened 2017-12-23
Proposal: Local & Upstream Bundles + Mix Bundle List


kevincwells:
# Note
This is a condensed form of a more detailed proposal on this topic. Please contact @kevincwells for information.

# Current State/Motivation
Currently, the way you add bundles to your mix is to copy the bundle definition file from a given version of CLR into the `mix-bundles` directory. That means the _existence_ of bundle definition files in your mix-bundles directory is the _specification_ of what bundles you want to include in your mix. Meanwhile, the mix-bundles directory is, in fact, input to other tools (`bundle-chroot-builder.py`).

This causes ambiguity and problems, especially when handling bundles that include other bundles. For example let’s say the developer cares about including Bundle_A in their mix, and that bundle itself includes Bundle_B. That means `mix-bundles` will contain both Bundle_A and Bundle_B, but there is no way to know that Bundle_B is only there as an include. So later, if Bundle_A gets modified upstream to no longer have that include, the developer is “stuck” with Bundle_B, because it is in their `mix-bundles` directory, and we cannot tell the difference.

# Proposal
Allow the user to include some bundles as-is from upstream (CLR) without maintaining a local copy.

- Change how to specify what bundles are included in a mix to be a Mix Bundles List containing all the bundles that must be in the mix. Other bundles will be automatically added by Mixer to fulfill bundle includes.
- Create a `local-bundles` directory to contain user-created and edited bundles.
  - Think of this as analogous to the `local-rpms` directory.
- Any bundles in the Mix Bundles List that are _not_ in the `local-bundles` directory are treated as "Upstream Bundles" and are included from upstream (CLR) as-is.
- Populate the `mix-bundles` directory (the input to `bundle-chroot-builder`) on the fly when chroots are built. Bundles included by other bundles are handled automatically, and local bundles always take precedence over their upstream counterparts.

cmarcelo:
Minor: would change "Change the specification of what bundles are included in a mix to be a Mix Bundles List containing only top-level bundles (i.e., not those that are included)"

to something like:

"Change how to specify what bundles are included in a mix to be a Mix Bundles List containing all the bundles that must be in the mix. Other bundles will be automatically added by Mixer to fulfill bundle includes."

with the intent of clarifying what bundles are guaranteed to be part of the mix (those listed).

kevincwells:
@cmarcelo I like that wording. Thanks -- I'll update it.

kevincwells:
Implemented in #137

PR #107, opened 2018-01-23
[WIP] Bundle refactor

kevincwells:
# Overview
This commit implements #55 and #57.

# Changes
## Major changes in this patch:
- Introduces the **Mix Bundles List**, a newline-separated file in the mixer working directory that lists the bundles that must be in the mix. Other bundles (included by those in the list) will be added automatically.
- Introduces the `local-bundles` directory that stores any bundles that the user created or that have been edited from upstream. Bundles in the `local-bundles` directory take precedence over upstream bundles of the same name.
- Generates the `mix-bundles` directory (used as input to the Bundle Chroot Builder) automatically on-the-fly when chroots are built. The Mix Bundles List is recursed to find all bundles needed for the mix, and the bundle definition files are copied from either the local or upstream bundles.
  - Note: Changes introduced in #102 will allow this `mix-bundles` generation to be completely eliminated. Instead, the functions used to populate this directory can be used to directly inform the new BCP code where to look for the bundles.
- Modifies the `mixer bundle add` command to now edit the Mix Bundles List. Bundles are verified to exist in either the local or upstream bundles. The user is informed whether the added bundles came from local or upstream.
- Adds a new `mixer bundle list` command that prints out the bundles in the Mix Bundle List, *all* bundles  in the mix (including those recursively included), all bundles available in local bundles, or all bundles available in upstream bundles.

## Minor changes in this patch:
- Files that generally have to do with configuration, such as the `builder.conf` and `.clearversion` file, now live inside the mixer working directory (`.mixer`) by default.
- As the `mix-bundles` directory is now created on-the-fly, it is no longer tracked as a git repository. Instead, the entire mixer working directory (`.mixer`) is tracked, with temporary, tool-generated files ignored. This allows one to track and correlate the Mix Bundle List, mix version, and upstream version as they are changed.
- The above-mentioned git revision tracking is now optional, and is only set up if `--git` is passed to `mixer init`. `mixer bundle add` still supports `--git`, which mirrors this behavior.
- The `--all` flags for `mixer init` and `mixer bundle add` have been split up into `--all-local` and `--all-upstream`. This allows a user to add or start with all local or upstream bundles, or both.
- A `mixer-tools/internal/stringset` package is added that introduces a new `stringset.Set` type. This is used throughout the Mix Bundle List code to efficiently handle bundle duplication (especially while recursing included bundles)  and list sorting.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -561,6 +561,19 @@ func (b *Builder) getMixBundlesListFull() ([]string, error) {
 	return bundleSet.Sort(), nil
 }
 
+func (b *Builder) getDirBundleList(dir string) ([]string, error) {
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to read bundles dir: %s", dir)
+	}
+
+	var bundles []string
+	for _, file := range files {
+		bundles = append(bundles, file.Name())
+	}
+	return bundles, nil
+}
+
 // AddBundles adds the specified bundles to the Mix Bundles List. Values are
 // verified as valid, and duplicate values are removed. The resulting Mix
 // Bundles List will be in sorted order.
@@ -640,6 +653,43 @@ func (b *Builder) AddBundles(bundles []string, alllocal bool, allupstream bool, 
 			return err
 		}
 	}
+	return nil
+}
+
+// ListBundles prints out the bundles in the mix, all bundles in the mix
+// including recursive includes, all bundles available in local bundles, or
+// all bundles available in upstream bundles
+func (b *Builder) ListBundles(full bool, local bool, upstream bool) error {
+	var bundles []string
+	var err error
+
+	switch {
+	case local:
+		bundles, err = b.getDirBundleList(b.Lbundledir)
+		if err != nil {
+			return errors.New("Error retreiving local bundle list")
+		}
+	case upstream:
+		bundles, err = b.getDirBundleList(b.getUpstreamBundlesPath(b.Clearver))
+		if err != nil {
+			return errors.New("Error retreiving local bundle list")
+		}
+	case full:
+		bundles, err = b.getMixBundlesListFull()
+		if err != nil {
+			return errors.New("Error retreiving mix bundle list")
+		}
+	default:
+		bundles, err = b.getMixBundlesList()
+		if err != nil {
+			return errors.New("Error retreiving mix bundle list")
+		}
+	}
+
+	for _, bundle := range bundles {
+		fmt.Println(bundle)
+	}
+
 	return nil
 }
 
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -31,6 +31,7 @@ import (
 	"strings"
 
 	"github.com/clearlinux/mixer-tools/helpers"
+	"github.com/clearlinux/mixer-tools/internal/stringset"
 	"github.com/clearlinux/mixer-tools/swupd"
 	"github.com/pkg/errors"
 )
@@ -50,6 +51,7 @@ type Builder struct {
 	Mixerdir    string
 
 	Bundledir   string
+	Lbundledir  string
 	Cert        string
 	Clearver    string
 	Format      string
@@ -128,14 +130,17 @@ func (b *Builder) CreateDefaultConfig(localrpms bool) error {
 	data := strings.Replace(string(raw), "/home/clr/mix", pwd, -1)
 
 	// Patch default directory for version files
-	// TODO: Clean this up once it's not needed
+	// TODO: Clean this up once it's not needed (i.e., once config is generated from scratch)
 	data = strings.Replace(data, "VERSIONS_PATH="+pwd+"\n", "VERSIONS_PATH="+filepath.Join(pwd, b.Mixerdir)+"\n", -1)
+	data = strings.Replace(data, "BUNDLE_DIR="+pwd+"/mix-bundles\n", "BUNDLE_DIR="+filepath.Join(pwd, b.Mixerdir, "mix-bundles")+"\n", -1)
+
+	// Add [Mixer] section
+	data += "\n[Mixer]\n"
+	data += "LOCAL_BUNDLE_DIR=" + pwd + "/local-bundles\n"
 
 	if localrpms {
-		// Add [Mixer] section
-		data += "\n[Mixer]\n"
-		data += "RPMDIR=" + pwd + "/rpms\n"
-		data += "REPODIR=" + pwd + "/local\n"
+		data += "LOCAL_RPM_DIR=" + pwd + "/local-rpms\n"
+		data += "LOCAL_REPO_DIR=" + pwd + "/local-yum\n"
 	}
 
 	if err = ioutil.WriteFile(builderconf, []byte(data), 0666); err != nil {
@@ -145,18 +150,29 @@ func (b *Builder) CreateDefaultConfig(localrpms bool) error {
 }
 
 // createRpmDirs creates the RPM directories
-func (b *Builder) createRpmDirs() error {
-	// Make the folder to store rpms if they don't already exist
+func (b *Builder) initDirs() error {
+	// Create mixer work dir
+	if err := os.MkdirAll(b.Mixerdir, 0777); err != nil {
+		return errors.Wrap(err, "Failed to create mixer working directory")
+	}
+
+	// Create folder to store local rpms if defined but doesn't already exist
 	if b.RPMdir != "" {
 		if err := os.MkdirAll(b.RPMdir, 0777); err != nil {
-			return err
+			return errors.Wrap(err, "Failed to create local rpms directory")
 		}
 	}
 
+	// Create folder for local yum repo if defined but doesn't already exist
 	if b.Repodir != "" {
 		if err := os.MkdirAll(b.Repodir, 0777); err != nil {
-			return err
+			return errors.Wrap(err, "Failed to create local rpms directory")
 		}
+	}
+
+	// Create folder for local bundle files
+	if err := os.MkdirAll(b.Lbundledir, 0777); err != nil {
+		return errors.Wrap(err, "Failed to create local bundles directory")
 	}
 
 	return nil
@@ -164,12 +180,13 @@ func (b *Builder) createRpmDirs() error {
 
 // InitMix will initialise a new swupd-client consumable "mix" with the given
 // based Clear Linux version and specified mix version.
-func (b *Builder) InitMix(clearver string, mixver string, all bool, upstreamurl string) error {
-	if err := os.MkdirAll(b.Mixerdir, 0777); err != nil {
-		return errors.Wrap(err, "Failed to create mixer working directory")
+func (b *Builder) InitMix(clearver string, mixver string, alllocal bool, allupstream bool, upstreamurl string, git bool) error {
+	// Set up local dirs
+	if err := b.initDirs(); err != nil {
+		return err
 	}
 
-	// Set up metadata
+	// Set up mix metadata
 	if err := ioutil.WriteFile(filepath.Join(b.Versiondir, "clearurl"), []byte(upstreamurl), 0644); err != nil {
 		return err
 	}
@@ -185,38 +202,36 @@ func (b *Builder) InitMix(clearver string, mixver string, all bool, upstreamurl 
 	}
 	b.Mixver = mixver
 
+	// Initialize the Mix Bundles List
+	if _, err := os.Stat(filepath.Join(b.Versiondir, "mixbundles")); os.IsNotExist(err) {
+		// Add default bundles (or all)
+		defaultBundles := []string{"os-core", "os-core-update", "bootloader", "kernel-native"}
+		if err := b.AddBundles(defaultBundles, alllocal, allupstream, false); err != nil {
+			return err
+		}
+	}
+
 	// Get upstream bundles
 	if err := b.getUpstreamBundles(clearver, true); err != nil {
 		return err
 	}
 
-	// Set up rpms and local dirs
-	if err := b.createRpmDirs(); err != nil {
-		return errors.Wrap(err, "Failed to create local RPM dirs")
-	}
-
-	// Set up mix-bundles dir
-	if _, err := os.Stat(b.Bundledir); os.IsNotExist(err) {
-		if err = os.MkdirAll(b.Bundledir, 0777); err != nil {
-			return errors.Wrap(err, "Failed to create mix-bundles dir")
+	if git {
+		// Generate .gitignore
+		gitignore := "upstream-bundles/\nmix-bundles/"
+		if err := ioutil.WriteFile(filepath.Join(b.Mixerdir, ".gitignore"), []byte(gitignore), 0644); err != nil {
+			return errors.Wrap(err, "Failed to create .gitignore file")
 		}
 
-		// Add default bundles (or all)
-		// Note: os-core-update already includes os-core, and kernel-native already includes bootloader
-		defaultBundles := []string{"os-core-update", "kernel-native"}
-		if _, err := b.AddBundles(defaultBundles, false, all, false); err != nil {
+		// Init repo and add initial commit
+		if err := helpers.Git("-C", b.Mixerdir, "init"); err != nil {
 			return err
 		}
-
-		// Init the mix-bundles git repo
-		if err := helpers.Git("-C", b.Bundledir, "init"); err != nil {
-			return err
-		}
-		if err := helpers.Git("-C", b.Bundledir, "add", "."); err != nil {
+		if err := helpers.Git("-C", b.Mixerdir, "add", "."); err != nil {
 			return err
 		}
 		commitMsg := fmt.Sprintf("Initial mix version %s from upstream version %s", b.Mixver, b.Clearver)
-		if err := helpers.Git("-C", b.Bundledir, "commit", "-m", commitMsg); err != nil {
+		if err := helpers.Git("-C", b.Mixerdir, "commit", "-m", commitMsg); err != nil {
 			return err
 		}
 	}
@@ -265,12 +280,13 @@ func (b *Builder) ReadBuilderConf() error {
 		required bool
 	}{
 		{`^BUNDLE_DIR\s*=\s*`, &b.Bundledir, true},
+		{`^LOCAL_BUNDLE_DIR\s*=\s*`, &b.Lbundledir, true},
 		{`^CERT\s*=\s*`, &b.Cert, true},
 		{`^CLEARVER\s*=\s*`, &b.Clearver, false},
 		{`^FORMAT\s*=\s*`, &b.Format, true},
 		{`^MIXVER\s*=\s*`, &b.Mixver, false},
-		{`^REPODIR\s*=\s*`, &b.Repodir, false},
-		{`^RPMDIR\s*=\s*`, &b.RPMdir, false},
+		{`^LOCAL_REPO_DIR\s*=\s*`, &b.Repodir, false},
+		{`^LOCAL_RPM_DIR\s*=\s*`, &b.RPMdir, false},
 		{`^SERVER_STATE_DIR\s*=\s*`, &b.Statedir, true},
 		{`^VERSIONS_PATH\s*=\s*`, &b.Versiondir, true},
 		{`^YUM_CONF\s*=\s*`, &b.Yumconf, true},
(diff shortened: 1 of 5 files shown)


kevincwells:
@tmarcu  -- I rebased this on top of master now that #86 is merged, and rebased on top of @rchiossi's new #101.  As a result, I no longer have the cherry-picked commit, and can cleanly rebase once #101 gets merged.

kevincwells:
Fully rebased.

builder/builder.go
```diff
@@ -561,6 +561,19 @@ func (b *Builder) getMixBundlesListFull() ([]string, error) {
 	return bundleSet.Sort(), nil
 }
 
+func (b *Builder) getDirBundleList(dir string) ([]string, error) {
+	files, err := ioutil.ReadDir(dir)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to read bundles dir: %s", dir)
+	}
+
+	var bundles []string
```
icarus-sparry: Since we know how long the list of bundles is going to be, we can avoid a certain amount of copying.

    var bundles = make([]string, len(files))
    for i := range files {
              bundles[i] = files[i].Name()
    }

builder/builder.go
```diff
@@ -643,6 +656,43 @@ func (b *Builder) AddBundles(bundles []string, alllocal bool, allupstream bool,
...
 }
 
+// ListBundles prints out the bundles in the mix, all bundles in the mix
+// including recursive includes, all bundles available in local bundles, or
+// all bundles available in upstream bundles
+func (b *Builder) ListBundles(full bool, local bool, upstream bool) error {
+	var bundles []string
+	var err error
+
+	switch {
```
icarus-sparry: This feels wrong. If a bundle can be at most full, local or upstream, then we should have a type saying so, rather than a set of bool. If on the other-hand it can be both full and upstream then we need a comment to say why being upstream is more important than being full.

builder/builder.go
```diff
@@ -100,7 +102,12 @@ func (b *Builder) CreateDefaultConfig(localrpms bool) error {
 		return err
 	}
 
-	builderconf := filepath.Join(pwd, "builder.conf")
+	// Create mixer work dir (for builder.conf) if needed
+	if err = os.MkdirAll(filepath.Join(pwd, ".mixer"), 0777); err != nil {
```
icarus-sparry: Should ".mixer" here not be b.MixerDir

builder/builder.go
```diff
@@ -117,8 +124,15 @@ func (b *Builder) CreateDefaultConfig(localrpms bool) error {
...
 	}
 
+	// Patch all default path prefixes to PWD
 	data := strings.Replace(string(raw), "/home/clr/mix", pwd, -1)
+
+	// Patch default directory for version files
+	// TODO: Clean this up once it's not needed
+	data = strings.Replace(data, "VERSIONS_PATH="+pwd+"\n", "VERSIONS_PATH="+filepath.Join(pwd, b.Mixerdir)+"\n", -1)
+
 	if localrpms {
```
icarus-sparry: Is it worth thinking about changing this to use the template system? Probably not for this PR, but future work.

internal/stringset/stringset.go
```diff
@@ -0,0 +1,72 @@
...
+// Delete removes one or more strings from the set
+func (s *Set) Delete(values ...string) {
+	for _, v := range values {
+		delete(*s, v)
+	}
+}
+
+// Contains checks if the set contains a string
+func (s *Set) Contains(value string) bool {
+	_, c := (*s)[value]
```
icarus-sparry: Traditional name for 'c' is 'ok'.

internal/stringset/stringset.go
```diff
@@ -0,0 +1,72 @@
...
+	return c
+}
+
+// Values returns the strings in the set as a slice of strings
+func (s *Set) Values() []string {
+	keys := make([]string, 0, len(*s))
+	for k := range *s {
+		keys = append(keys, k)
```
icarus-sparry: I would have used assignment rather than append. I guess it depends on if the runtime & compiler are smart enough to optimize this code better or not.

internal/stringset/stringset.go
```diff
@@ -0,0 +1,72 @@
+// Copyright © 2018 Intel Corporation
```
icarus-sparry: This package induces one huge WTF. WTF is this for, nothing else seems to use it. no tests...

internal/stringset/stringset.go
```diff
@@ -0,0 +1,72 @@
...
+// While the functionality of this package could easily be expanded, it would
+// likely soon be better to use a more full-featured, public set package.
+package stringset
+
+import (
+	"sort"
+)
+
+// Set represents a set collection of strings.
+type Set map[string]struct{}
```
icarus-sparry: This is good, but perhaps a comment to explain why a mapping to an empty struct is the way to do things (i.e. minimal runtime space) would help.

internal/stringset/stringset.go
```diff
@@ -0,0 +1,72 @@
...
+func (s *Set) Add(values ...string) {
+	for _, v := range values {
+		(*s)[v] = struct{}{}
+	}
+}
+
+// Delete removes one or more strings from the set
+func (s *Set) Delete(values ...string) {
+	for _, v := range values {
+		delete(*s, v)
```
icarus-sparry: No error to delete something twice?

internal/stringset/stringset.go
```diff
@@ -0,0 +1,72 @@
...
+	s.Add(values...)
+	return s
+}
+
+// Add adds one or more strings to the set
+func (s *Set) Add(values ...string) {
+	for _, v := range values {
+		(*s)[v] = struct{}{}
+	}
+}
```
icarus-sparry: Since New returns s, perhaps Add and Delete should do so as well?

builder/builder.go
```diff
@@ -185,38 +202,36 @@ func (b *Builder) InitMix(clearver string, mixver string, all bool, upstreamurl
 	}
 	b.Mixver = mixver
 
+	// Initialize the Mix Bundles List
+	if _, err := os.Stat(filepath.Join(b.Versiondir, "mixbundles")); os.IsNotExist(err) {
+		// Add default bundles (or all)
+		defaultBundles := []string{"os-core", "os-core-update", "bootloader", "kernel-native"}
+		if err := b.AddBundles(defaultBundles, alllocal, allupstream, false); err != nil {
```
tmarcu: allLocal feels easier to read

builder/builder.go
kevincwells: Yes, nice catch. I wrote the entire `mixer bundle list` functionality in about 10 minutes, so I am guessing there are other optimizations I didn't do as well.

Will fix in a future push.

builder/builder.go
kevincwells: Yes, it should. And... I thought it was. I think this got lost in my pretty extensive rebasing yesterday.

builder/builder.go
kevincwells: I think you're misunderstanding the conditions. A _bundles_ isn't "full" or "upstream", the _list_ is. You're either getting a "full list" or an "upstream list".

If you look at the flag's documentation, the command either lists the items in the mix bundles list, the _full_ set of bundles that will be included (by recursively unpacking the includes in the mix bundles list), all of the available local bundles, or all of the available upstream bundles.

You're right that the precedence of how the command behaves (i.e., passing `--upstream` trumps `--full`) needs to be documented.

internal/stringset/stringset.go
kevincwells: `New` is a factory function; `Add` and `Delete` are methods on the Set object itself. That feels like a pretty normal convention to me.

internal/stringset/stringset.go
kevincwells: I considered Contains checks for both addition and deletion, but there wasn't a use case for that yet. Do you think it's important?

internal/stringset/stringset.go
kevincwells: Gotcha. Will update.

internal/stringset/stringset.go
kevincwells: True -- since I have the full array in memory already from the `make` above, assignment here would be easy. I would need to introduce an index variable, though, since `range` on a map doesn't return one on its own.

builder/builder.go
kevincwells: What, you don't like three lowercase l's in a row?!

cmarcelo:
Comments on the description:

> Files that generally have to do with configuration, such as the builder.conf and .clearversion file, now live inside the mixer working directory (.mixer) by default.

I think this shouldn't be a hidden directory, I would just call it "mixer" or "mix" without the dot. I think it will contain data that is going to be edited by the user and something the user might want to put on a git repository.

The layout that I'm thinking about of such a git repository is:

```
caio's my repository
    .gitignore
    mixer/                     # mix-bundle-list, builder.conf, .clearversion (which also shouldn't be a hidden file!)
    local-rpms/           # possibly empty, possibly I'd add a Makefile outside to download those
    local-bundles/
        caio         # a bundle description
        os-core    # my own version of os-core
    local-specs/            # in the future, we might teach mixer how to deal with specs directly....
    Makefile, or whatever.bash            # stuff I might add for the sake of my CI to verify stuff
```

Note that I still want to git track local bundle descriptions and it does make sense put them in the same repository that will hold the specification of clear version and mix version.

Also note that I'm fine adding transient content to a hidden folder (e.g. the merged mix bundles directory).

internal/stringset/stringset.go
kevincwells: Great question, and you very well might come back with a great alternative implementation.

For working with bundle lists, I need a data structure to store the bundle names read in from the Mix Bundle List, to add bundles to as I recurse that list, etc.

One of the requirements for the bundle list processing is the ability to eliminate duplicate bundles.  I originally was going to just use a slice to contain the bundle names (simple, easy, built-in), but the uniqueness check then requires traversing the list each time, creating an O(n^2) operation just for insertion.

A "collection of unique items" is obviously a set. Python and Ruby both have a built-in set type, and Java has plenty of variants in their standard library. I was surprised to find Go didn't bother. The closest I found to an "official" explanation was, essentially, "If you want a set of something in Go, it's easy enough to write it yourself using a `map[something]bool` or `map[something]struct{}` for a more space-efficient variant.`

"I have to implement my own Set?!" I thought. "Surely there's a good library out there for this". And there are -- I found several -- but all contained a heck of a lot more than I needed (set intersection and union, multithread mutex locking, etc), so I gave up and wrote the 30ish (if you take out comments and whitespace) lines of code myself.

cmarcelo:
After writing down that layout, I'm even thinking we don't need a mixer directory and could use the root directory of the repository to hold builder.conf and friends.

internal/stringset/stringset.go
kevincwells: As for why I wrote this as a generic string set, rather than a more-specific "bundle set" for a new `Bundle` data type, well -- at the time I needed it, I was only working with bundle names, which are strings.  Perhaps it'd be more useful to have a "bundle set" feature in the future. I could have called it a "bundle set" and still just made it strings, but I figured if I'm going to write a set that is simply of strings, might as well admit to that in case it's useful elsewhere for non-bundle-related strings.

As for why I made it a standalone internal package, well, that came from discussions with @cmarcelo in a few places about how best to organize generic code, and reading a bunch of Go blogs and discussions about package organization and how best to expose utility functionality.

The fact no one else uses this yet is breaking with @cmarcelo's suggestion of keeping things next to where they're used until it truly is generic. But it's in keeping with a broader opinion of "organize your utility functionality into meaningful packages with consolidated behavior".

internal/stringset/stringset.go
```diff
@@ -0,0 +1,72 @@
...
+// limitations under the License.
+
+// Package stringset implements a very basic set collection of strings. Items
+// are unique, and basic set operations of addition, deletion, and existence
+// testing are supported. Additionally, supports exporting the elements as a
+// slice, optionally with the values sorted in ascending order.
+//
+// While the functionality of this package could easily be expanded, it would
+// likely soon be better to use a more full-featured, public set package.
+package stringset
```
cmarcelo: I feel like most of functionality you could get by making a couple of the functions in this package to work on map[string]bool and drop the new type. So my inclination would be delay from creating the abstraction.

```
    x := make(map[string]bool)

    # Add an element.
    x["a"] = true
    
    # Remove an element.
    delete(x, "a")
    
    # Is it in the list?
    if x["a"] { fmt.Println("it is!") }

    # The two functions I'm thinking of.
    v := GetKeys(x)       # --> gives me []string
    v := SortedKeys(x)    # --> gives me []string, sorted.
```

Sure using struct{} instead of bool can give you less memory usage, but for the stuff we are talking about here, I don't think it matters. If we are looking at providing our own non-map implementation, or make the Set thread-safe, that deserves a package, but that doesn't seem the case.

mixer/cmd/root.go
```diff
@@ -85,16 +85,17 @@ var rootCmdFlags = struct {
 }{}
 
 type initCmdFlags struct {
-	all         bool
+	alllocal    bool
+	allupstream bool
 	clearver    int
 	mixver      int
+	localrpms   bool
```
cmarcelo: Optional: stick to the Go naming conventions, we'll fix the other names as we touch the code.

allLocal
allUpstream
localRPMs

builder/builder.go
```diff
@@ -47,8 +48,10 @@ var UseNewSwupdServer = false
 type Builder struct {
 	Buildscript string
 	Buildconf   string
+	Mixerdir    string
 
 	Bundledir   string
+	Lbundledir  string
```
cmarcelo: Optional: LocalBundleDir, MixerDir. (We'll eventually fix the other names in this struct)

builder/builder.go
icarus-sparry: It is entirely possible I am misunderstanding, I am looking at the code rather than the big picture. As I said it "feels wrong". So let me rephrase the question - what are the valid states of the 3 bool values? The 3 bools look to just be to make it trivial to use Cobra, without any warning to the user that they are misunderstanding what they are asking for.

internal/stringset/stringset.go
icarus-sparry: My point is that the 30ish lines you have written are nothing to do with this PR - "Bundle refactor". Please remove it.

Yes, go does require you to implement 'set' on top of map. map gives you the O(1) lookup time and other good properties that you typically want.

It may be that what you find up wanting in the future is a pointer to bundle, rather than a bundle.

internal/stringset/stringset.go
icarus-sparry: Since there is no code using this, it is hard to say one way or another!.

builder/builder.go
cmarcelo: Optional: maybe the feeling is an indication that the interface could be improved. Two possible venues here:

(a) you have 4 mutual exclusive cases, so make this visible in the code. Make your list function take an enum or give 4 different functions to the callsite. Callsite will handle the boolean drama matching.

(b) Rework the CLI.
- mixer bundle list: show all bundles in the bundle list, their dependencies (marked in some way) AND the local bundles NOT included (marked in some way).
- mixer bundle list --upstream: show the all upstream bundles, with those used/included by the bundle list marked.

And maybe in that case Builder would expose 2 different functions, instead one with a boolean.

internal/stringset/stringset.go
cmarcelo: I feel like this knot could be untied if I've extracted the bundleset package out of the BCB patch (this is easy and wouldn't break anything), and add whatever extra things you need (you talked about also parsing some of the fields in comments). @kevincwells  Would that be helpful?

internal/stringset/stringset.go
icarus-sparry: Actually I like that he is using an empty struct, it shows that this is written in go rather than ported from say python. @cmarcelo you are perfectly correct that the memory usage for this case is trivial, which is part of the reason I didn't comment on your use of string->bool maps elsewhere, and it does make the `if x["a"] { fmt.Println("it is!") }` case much nicer to read than ` if _,ok := x["a"]; ok { fmt.Println("it is!") }`

However the *reason* for it being an empty struct (to save space) rather than the more obvious `bool` should be explained in the file if we are going to keep it in.

builder/builder.go
```diff
@@ -643,6 +656,43 @@ func (b *Builder) AddBundles(bundles []string, alllocal bool, allupstream bool,
...
+// all bundles available in upstream bundles
+func (b *Builder) ListBundles(full bool, local bool, upstream bool) error {
+	var bundles []string
+	var err error
+
+	switch {
+	case local:
+		bundles, err = b.getDirBundleList(b.Lbundledir)
+		if err != nil {
+			return errors.New("Error retreiving local bundle list")
```
tmarcu: s/retreiving/retrieving/

builder/builder.go
```diff
@@ -643,6 +656,43 @@ func (b *Builder) AddBundles(bundles []string, alllocal bool, allupstream bool,
...
+		}
+	case upstream:
+		bundles, err = b.getDirBundleList(b.getUpstreamBundlesPath(b.Clearver))
+		if err != nil {
+			return errors.New("Error retreiving local bundle list")
```
tmarcu: s/retreiving/retrieving/

builder/builder.go
```diff
@@ -643,6 +656,43 @@ func (b *Builder) AddBundles(bundles []string, alllocal bool, allupstream bool,
...
+		}
+	case full:
+		bundles, err = b.getMixBundlesListFull()
+		if err != nil {
+			return errors.New("Error retreiving mix bundle list")
```
tmarcu: s/retreiving/retrieving/

builder/builder.go
```diff
@@ -643,6 +656,43 @@ func (b *Builder) AddBundles(bundles []string, alllocal bool, allupstream bool,
...
+		}
+	default:
+		bundles, err = b.getMixBundlesList()
+		if err != nil {
+			return errors.New("Error retreiving mix bundle list")
```
tmarcu: s/retreiving/retrieving/

mixer/cmd/bundles.go
```diff
@@ -52,17 +54,44 @@ var addBundlesCmd = &cobra.Command{
 		if err != nil {
 			fail(err)
 		}
-		err = b.AddBundles(bundles, bundleFlags.alllocal, bundleFlags.allupstream, bundleFlags.git)
+		err = b.AddBundles(bundles, bundleAddFlags.alllocal, bundleAddFlags.allupstream, bundleAddFlags.git)
```
tmarcu: lll

builder/builder.go
```diff
@@ -643,6 +656,43 @@ func (b *Builder) AddBundles(bundles []string, alllocal bool, allupstream bool,
...
+		bundles, err = b.getDirBundleList(b.getUpstreamBundlesPath(b.Clearver))
+		if err != nil {
+			return errors.New("Error retreiving local bundle list")
+		}
+	case full:
+		bundles, err = b.getMixBundlesListFull()
+		if err != nil {
+			return errors.New("Error retreiving mix bundle list")
+		}
+	default:
```
tmarcu: This should not be the default. The default is "I want to know what bundles I'm about to spend X amount of time building," which should include every single bundle that is going to get built (includes, deps, etc). Unless that is what this is already doing and I misunderstood the code :)

builder/builder.go
```diff
@@ -145,31 +150,43 @@ func (b *Builder) CreateDefaultConfig(localrpms bool) error {
 }
 
 // createRpmDirs creates the RPM directories
```
tmarcu: Need to update comment.

builder/builder.go
```diff
@@ -145,31 +150,43 @@ func (b *Builder) CreateDefaultConfig(localrpms bool) error {
...
-			return err
+			return errors.Wrap(err, "Failed to create local rpms directory")
 		}
 	}
 
+	// Create folder for local yum repo if defined but doesn't already exist
 	if b.Repodir != "" {
 		if err := os.MkdirAll(b.Repodir, 0777); err != nil {
-			return err
+			return errors.Wrap(err, "Failed to create local rpms directory")
```
tmarcu: copy pasta from above, should be local yum repo directory :)

builder/builder.go
```diff
@@ -145,31 +150,43 @@ func (b *Builder) CreateDefaultConfig(localrpms bool) error {
...
 		}
 	}
 
+	// Create folder for local bundle files
+	if err := os.MkdirAll(b.Lbundledir, 0777); err != nil {
```
tmarcu: LocalBundleDir would be more appropriate, as mentioned by someone else already.

builder/builder.go
```diff
@@ -426,93 +442,205 @@ func (b *Builder) getUpstreamBundles(ver string, prune bool) error {
...
-					return bundleAddCount, errors.Errorf("Cannot parse bundle %s from CLR version %s", bundle, b.Clearver)
-				} else if len(ib) > 0 {
-					includes = append(includes, ib...)
+	}
+	return includes, nil
+}
+
+// buildMixBundleListFull adds a list of bundles to a bundle set, recursively
+// adding any bundles included by those in the list.
+func (b *Builder) buildMixBundleSetFull(bundlelist []string, bundleSet stringset.Set) (stringset.Set, error) {
```
tmarcu: :+1:

builder/builder.go
```diff
@@ -426,93 +442,205 @@ func (b *Builder) getUpstreamBundles(ver string, prune bool) error {
...
+
+	return bundleSet.Sort(), nil
+}
+
+// AddBundles adds the specified bundles to the Mix Bundles List. Values are
+// verified as valid, and duplicate values are removed. The resulting Mix
+// Bundles List will be in sorted order.
+//
+// bundles: array slice of bundle names
+// all: include all CLR bundles. Overrides bundles.
```
tmarcu: These seem redundant but ymmv

builder/builder.go
```diff
@@ -426,93 +442,205 @@ func (b *Builder) getUpstreamBundles(ver string, prune bool) error {
...
+		clrbundledir := b.getUpstreamBundlesPath(b.Clearver)
+		files, err := ioutil.ReadDir(clrbundledir)
 		if err != nil {
-			return bundleAddCount, err
+			return errors.Wrapf(err, "Failed to read upstream bundles dir: %s", clrbundledir)
+		}
+
+		for _, file := range files {
+			bundle := file.Name()
+			bundles = append(bundles, bundle)
```
tmarcu: why not just 
`bundles = append(bundles, file.Name())` instead of creating a new var each time?

builder/builder.go
```diff
@@ -522,10 +650,39 @@ func (b *Builder) UpdateMixVer() error {
 	return ioutil.WriteFile(filepath.Join(b.Versiondir, "mixversion"), []byte(strconv.Itoa(mixver+10)), 0644)
 }
 
+func (b *Builder) buildMixBundleDir() error {
+	if err := os.MkdirAll(b.Bundledir, 0777); err != nil {
+		return errors.Errorf("Failed to create dir: %s", b.Bundledir)
+	}
+	// TODO: clear it out if it exists
```
tmarcu: What are the consequences of not clearling it out right now? Can we just include that in this PR?

builder/builder.go
```diff
@@ -522,10 +650,39 @@ func (b *Builder) UpdateMixVer() error {
...
+	if err != nil {
+		return err
+	}
+
+	for _, bundle := range bundles {
+		path, err := b.getBundlePath(bundle)
+		if err != nil {
+			return err
+		}
+		if err = helpers.CopyFile(filepath.Join(b.Bundledir, bundle), path); err != nil {
```
tmarcu: I would add a comment here saying you always **want** to overwrite here, which is why you're not using CopyFileNoOverwrite in this case. Is that a correct assumption?

internal/stringset/stringset.go
tmarcu: Not sure how I feel about this addition. I see how you use it in your code and it's fine, but seems like a slightly unnecessary abstraction. I defer to @icarus-sparry and @cmarcelo  here...

tmarcu:
I also share @cmarcelo sentiments on making .mixer not a hidden directory, and also would be OK with just having everything live in the toplevel workspace too (like it currently is setup). It would make things a bit more transparent IMO.

internal/stringset/stringset.go
kevincwells: @cmarcelo I'm fine with changing it to this implementation.

As for the bool vs empty struct... I chose the empty struct when I was abstracting it behind a set interface, because of the space advantage. But from a readability and maintainability standpoint, it's awful.  I would much rather _use_ the bool.

YMMV, but for me priorities go "serious space efficiency requirements" > "readability/maintainability" > "trivial space efficiency".  IMHO, "evidence something was written originally in Go rather than ported from Python" seems like a very odd thing to desire to me, and doesn't fall on my priority list at all.

mixer/cmd/root.go
kevincwells: I greatly prefer that, I was just matching the rest of the code.

builder/builder.go
kevincwells: **sighs and repeats "i before e, except after c, ....**

Sorry.

builder/builder.go
kevincwells: copypasta

builder/builder.go
kevincwells: Boooo, this got lost in my rebase. Thanks for catching!

builder/builder.go
kevincwells: Nah, you understood correctly. Valid point.

I fall on the other side of the fence. There has _never been_ a Mix Bundle List before, so I know your mindset is definitely firmly in "what bundles are going in my mix" and not "what bundles do I _want_ in my mix". And it's that second thought that I'm trying to push things toward, from a UX perspective. That's large part of the motivation for the Mix Bundle List -- stating only the bundles you care about, and letting the tool figure out what dependencies it needs to pull in to meet that.  It's _that_ list that all of the `mixer bundle add/delete/edit/whatever` commands are intended to deal with.

For debugging or validation purposes, it's invaluable to be able to see (without trudging through leftover build artifacts) what all bundles will be fully in the list, but IMHO that's not what users should _normally_ be thinking about.

When someone runs `mixer bundle add editors`, for example, it's _only_ the `editors` bundle that is added to their Mix Bundle List. So if the user then does `mixer bundle list`, having it also spit back out `python-basic` and `python3-basic` feels really confusing. If they then were to go "Oh weird, I don't want python! `mixer bundle remove python-basic`, **nothing would happen**, because python-basic was never in the list, only `editors`.

builder/builder.go
kevincwells: Oh whoops! Hahah, completely forgot to go back and implement that. Definitely intended to put that in this PR.

builder/builder.go
kevincwells: I absolutely meant to clear the dir above, so if the code explicitly did that, would you still want the comment? It's a moot point kinda at that point.

builder/builder.go
tmarcu: I disagree, you are about *all* your inputs, not just what you think you added. There have been multiple emails and issues opened where people try to add a single bundle (sysadmin-basic, or editors) which pulls in python, python3, etc, which are massive, and to their dismay their image/container is now GB in size vs just a few hundred MB as they thought. It's been a pain point for us for a while. As someone who is building an OS I want to know what things are going into my build and why. 
Yes, before your update to recurse includes and bring them in, it would crash and say "you're missing these bundles too," and you would have to add them manually. However, you're still aware of exactly what's being added and why, and finding out what your next build will have means typing  "ls mix-bundles". 
I don't see a reason to hide this from people at all, it creates ambiguity and more commands to run. The ideal  scenario IMO is a local-bundles folder, with users' custom bundle defs, the clr-bundles-VER folder with upstream bundles, and the mixbundles.list which specifies ALL bundles going into a given build.  The list can be more detailed specifying which bundles come from where etc etc, but at the least it gives an easily query-able data point for the builder input. Or you generate the mix-bundles directory each time the mixbundles.list is updated and data can be read from there. There has to be some persistent data of what will actually  be built, and if a user then wants to know "which ones only came from me, or only from upstream" they can pass the appropriate flags then.

kevincwells:
Replaced by #137

PR #109, opened 2018-01-24
builder: propagate error to caller

rchiossi:
This patch removes the last occurrences where an error would be handled
by builder itself and cause the program to Exit. Now the error is always
propagated to the caller.


rchiossi:
Rebased on new master

tmarcu:
Aw man I did not think you were around and wanted to merge it, so I rebased it myself as well.. x) Apologies.

PR #110, opened 2018-01-25
Add BAT tests for UI testing

matthewrsj:
Add BAT tests for command-line interface testing. Use --new-swupd option to run with new integrated swupd implementation.

TODO: make them run faster pls
TODO: add `make batcheck` to .travis.yml

Collaboration between @matthewrsj @icarus-sparry and @tmarcu (mostly this guy, he originally wrote the tests)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,11 @@ go:
 
 go_import_path: github.com/clearlinux/mixer-tools
 
+services:
+    - docker
+
 before_install:
-    - sudo apt-get update -qq
-    - go get -u gopkg.in/alecthomas/gometalinter.v2
-    - gometalinter.v2 --install
-    - wget https://github.com/clearlinux/bsdiff/releases/download/v1.0.2/bsdiff-1.0.2.tar.xz
-    - tar -xvf bsdiff-1.0.2.tar.xz
-    - pushd bsdiff-1.0.2 && ./configure --prefix=/usr --disable-tests && make -j48 && sudo make install && popd
+    - docker build -t testdock .
 
 script:
-    - make lint
-    - make check
+    - docker run testdock
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM clearlinux:latest
+
+ENV GOPATH /home/gopath
+ENV PATH="/home/gopath/bin:${PATH}"
+COPY . / /home/gopath/src/github.com/clearlinux/mixer-tools/
+RUN swupd bundle-add mixer go-basic c-basic os-core-update-dev && \
+    git config --global user.email "travis@example.com" && \
+    git config --global user.name "Travis CI" && \
+    clrtrust generate && \
+    go get -u gopkg.in/alecthomas/gometalinter.v2 && \
+    gometalinter.v2 --install
+
+ENTRYPOINT ["/bin/sh", "-c", "cd /home/gopath/src/github.com/clearlinux/mixer-tools && make && make install && make lint; make check"]
+
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@
 #
 # For historical purposes, it also works in a development environment
 # when the repository is already inside a GOPATH.
+include Makefile.bats
 
 .NOTPARALLEL:
 
--- a/Makefile.bats
+++ b/Makefile.bats
@@ -1,0 +1,18 @@
+SUBMAKEFILES := $(shell find bat/tests/ -name Makefile)
+DIRS2RUNMAKECHECK := $(addprefix checkdir-,${SUBMAKEFILES})
+DIRS2RUNMAKECLEAN := $(addprefix clean-,${SUBMAKEFILES})
+
+batcheck: ${DIRS2RUNMAKECHECK}
+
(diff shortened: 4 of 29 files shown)


matthewrsj:
Note: I get some weird permission denied failures when running `make batclean` from the root of the repo for some files.

tmarcu:
:boom: :rocket:

tmarcu:
@kevincwells 
`not ok 2 Create version 20 with unique custom bundle
# (from function `mixer-init-versions' in file ../../lib/mixerlib.bash, line 35,
#  in test file run.bats, line 25)
#   `mixer-init-versions $CLRVER 20' failed
# ERROR: buildconf missing entry for variable: LOCAL_BUNDLE_DIR`
Just need to add this var to the bat test builder.conf?

bat/lib/locallib.bash
```diff
@@ -0,0 +1,251 @@
+# NOTE: source this file from a BATS test case file
```
phmccarty: This file is unused, so you should remove it.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN mkdir -p /home/gopath/src/github.com/clearlinux/mixer-tools
+COPY . / /home/gopath/src/github.com/clearlinux/mixer-tools/
+RUN swupd bundle-add mixer go-basic c-basic dev-utils os-core-update-dev
+RUN git config --global user.email "travis@example.com"
+RUN git config --global user.name "Travis CI"
+RUN clrtrust generate
+RUN go get -u gopkg.in/alecthomas/gometalinter.v2
+RUN gometalinter.v2 --install
+
+ENTRYPOINT ["/bin/sh", "-c", "cd /home/gopath/src/github.com/clearlinux/mixer-tools && make && make install && make lint && make check && make batcheck"]
```
matthewrsj: We talked about having these run separately so even if `make lint` fails we can see the output of the other tests. I bet developers don't want to wait 20 minutes every time to see the **next** failure.

icarus-sparry: By `separately` do you mean changing the `&&` to `;`?

kevincwells:
... Erm, neither of my above comments are showing up in-place for me now.

.travis.yml
```diff
@@ -6,14 +6,11 @@ go:
...
+    - docker
+
 before_install:
-    - sudo apt-get update -qq
-    - go get -u gopkg.in/alecthomas/gometalinter.v2
-    - gometalinter.v2 --install
-    - wget https://github.com/clearlinux/bsdiff/releases/download/v1.0.2/bsdiff-1.0.2.tar.xz
-    - tar -xvf bsdiff-1.0.2.tar.xz
-    - pushd bsdiff-1.0.2 && ./configure --prefix=/usr --disable-tests && make -j48 && sudo make install && popd
+    - docker build -t clearlinux .
```
kevincwells: FYI - This overwrites/masks the standard `clearlinux:latest` image from upstream. Might be fine on Travis, but not something people will want to build locally for testing.

kevincwells: > @cmarcelo: Maybe we should just pick a different name?

Yeah, easy fix. Could even just do a different tag, but a different name is safer.

Dockerfile
kevincwells: Some thoughts:

This isn't _wrong_, but normally you'd put something like this into a script that gets copied over at build and run as the single entrypoint command.

Since all of these commands are executed in the same directory, normally you would use `WORKDIR` to configure this, and forego the `cd`. You'd just want to put it at the end before the `ENTRYPOINT`, though I don't think any of the earlier commands would actually fail if run in that directory, so you could in theory set it anywhere in the `Dockerfile`.

Also, you can override `ENTRYPOINT` at container run time with `--entrypoint`, so can also pretty easily call these steps independently with the rest of the `Dockerfile` as-is.

Another option for breaking this into steps would be to have the scripting outside the Dockerfile -- have the container run in the background, then execute each of the steps inside it from outside with `docker exec <container_id> <cmd>`

tmarcu: Yeah I thought of that as well, but I wanted this container to be treated as an executable essentially that is just run, and won't be taking arguments. I don't see this implementation happening for the full docker setup when mixer will only run out of a container, but this works nicely for travis testing IMO.

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+ENV GOPATH /home/gopath
+ENV PATH="/home/gopath/bin:${PATH}"
+RUN mkdir -p /home/gopath/src/github.com/clearlinux/mixer-tools
+COPY . / /home/gopath/src/github.com/clearlinux/mixer-tools/
+RUN swupd bundle-add mixer go-basic c-basic dev-utils os-core-update-dev
+RUN git config --global user.email "travis@example.com"
+RUN git config --global user.name "Travis CI"
+RUN clrtrust generate
+RUN go get -u gopkg.in/alecthomas/gometalinter.v2
+RUN gometalinter.v2 --install
```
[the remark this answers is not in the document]
tmarcu: Didn't know about copy! I will change the RUN command as well, I didn't have a strong feeling on it since it never persists afaik, but it's an easy changeup.

kevincwells:
:+1: Dockerfile changes

PR #111, opened 2018-01-26
swupd: add tests for packs

cmarcelo:
- Add tests for the logic of figuring out which bundles need delta
  packs between two versions.

- Add tests for creating zero packs, with data from different sources,
  as well as incomplete data (so fallback logic kicks in), validating
  the contents of the generated pack.

- Add testFileSystem helper struct, to make filesystem-like operations
  in a test directory. By using methods of the struct, we reduce a bit
  the noise of passing t and dir around.

### DIFF
--- a/swupd/packs_test.go
+++ b/swupd/packs_test.go
@@ -1,0 +1,403 @@
+package swupd
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"reflect"
+	"sort"
+	"strings"
+	"testing"
+)
+
+func TestFindBundlesToPack(t *testing.T) {
+	type M map[string]uint32
+	tests := []struct {
+		Name       string
+		FromV      uint32
+		From       M
+		ToV        uint32
+		To         M
+		Expected   []BundleToPack
+		ShouldFail bool
+	}{
+		{
+			Name: "simple case",
+
+			FromV: 10,
+			From:  M{"os-core": 10},
+
+			ToV: 20,
+			To:  M{"os-core": 20},
+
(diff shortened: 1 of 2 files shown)


cmarcelo:
@matthewrsj will add non-zero packs tests together with the delta integration.

swupd/packs_test.go
```diff
@@ -0,0 +1,401 @@
...
+			}
+		}
+		toM := &Manifest{}
+		toM.Header.Version = tt.ToV
+		for name, v := range tt.To {
+			addBundle(toM, name, v)
+		}
+
+		bundleMap, err := FindBundlesToPack(fromM, toM)
+		if err != nil {
```
matthewrsj: I've suggested this elsewhere, you can (optionally) simplify this logic by doing something like
```go
if (err != nil) != tt.ShouldFail {
    if tt.ShouldFail {
        t.Fatalf(...)
    } else {
        t.Fatalf(...)
    }
    continue
}
```

cmarcelo:
> Just an optional suggestion. It would be very nice to use those new helpers in the CreateManifests tests.

Sure, I thought the same, but as the tests were working already, I pushed to later. I'd be happy to take care of it once delta integration is done.

swupd/packs_test.go
cmarcelo: I haven't looked at the BCB review, but I will try that approach there.

cmarcelo: It did look better on the BCB, so I'll try the same thing here.

cmarcelo: Done.

PR #112, opened 2018-01-26
swupd-extract: initial import

cmarcelo:
swupd-extract is a program that reads a swupd update repository and
extracts the files for bundles of a specific version. It does a
similar job as "swupd verify --install --no-scripts --no-boot-update".

There is code for maintaining the client state, but at the moment it
only care about zero packs. Verification of the Manifest.MoM signature
is done using "openssl" external program.

It has some niceties when we are extracting Clear Linux content

    // List the latest clear linux bundles.
    $ sudo swupd-extract clear

    // Extracts os-core into "output/" directory.
    $ sudo swupd-extract clear os-core

    // Extracts editors and all its included bundles into mydir.
    $ sudo swupd-extract --output mydir clear/20540 editors

but instead of clear, a proper URL with version can be used instead,
i.e. it is easy to use it to extract mixes content. It also works fine
with local paths. The included help enumerates the rest of the features.

There is some duplicate code for compressedTarReader which I intend to
make into an internal package further on (as well as consolidating
other related archive code).

The main motivation was to allow users of distros other than Clear
Linux to use mkosi to generate Clear Linux images, but the program can
also be useful for validation.

### DIFF
--- a/swupd-extract/main.go
+++ b/swupd-extract/main.go
@@ -1,0 +1,516 @@
+package main
+
+// TODO: Flag to ignore included bundles (only extract the requested). Still need to parse
+// the included to figure out the directories metadata. Skip the packs for includes, just
+// download fullfile.
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+	"syscall"
+
+	"github.com/clearlinux/mixer-tools/swupd"
+)
+
+func usage() {
+	fmt.Printf(`swupd-extract reads a swupd update repository and write it to disk
+
+Usage:
+
+    swupd-extract https://.../version        [bundles...]
+    swupd-extract /local/path/to/.../version [bundles...]
+
+The program downloads the content of a specific version in a swupd
+repository and extract the content of the specified bundles (groups of
+content). If no bundle is specified the program will list the bundles
+available. Bundles specified will automatically trigger the extraction
+of bundles that they include.
+
+The program extracts the content to a directory called "output" or a
+directory set with the -output flag. Intermediate data is saved in a
+directory sibling to the output directory called "swupd-state" or a
+directory set with the -state flag.
+
+The state directory can hold intermediate data for multiple versions
+from the same repository.
+
+There is a shorthand for extracting content from Clear Linux. For this
+case the version can be omitted so that the latest version is used:
+
+    swupd-extract clear/20520 [bundles...]
+    swupd-extract clear       [bundles...]
+
+If not available locally, the certificate for Clear Linux is
+automatically downloaded and verified.
+
+Flags:
+`)
+	flag.PrintDefaults()
+	os.Exit(1)
+}
+
+const (
+	clearLinuxLatestURL         = "https://download.clearlinux.org/latest"
+	clearLinuxBaseContent       = "https://cdn.download.clearlinux.org/update"
+	clearLinuxCertificateURL    = "https://download.clearlinux.org/current/Swupd_Root.pem"
+	clearLinuxCertificateSHA256 = "ff06fc76ec5148040acb4fcb2bc8105cc72f1963b55de0daf3a4ed664c6fe72c"
+)
+
+func main() {
+	log.SetFlags(0)
+
+	var (
+		outputDir   string
+		stateDir    string
+		cert        string
+		noCache     bool
+		noOverwrite bool
+	)
+
+	flag.StringVar(&outputDir, "output", "output", "where to extract the files")
+	flag.StringVar(&stateDir, "state", "", "directory to store intermediate files")
+	flag.StringVar(&cert, "cert", "", "certificate used to verify content")
+	flag.BoolVar(&noCache, "no-cache", false, "don't use cached files, force downloads")
+	flag.BoolVar(&noOverwrite, "no-overwrite", false, "don't overwrite output files")
+	flag.Parse()
+
+	if os.Getuid() != 0 {
+		log.Fatal("This program needs to run as root to write files with proper permissions.")
+	}
+
+	if len(flag.Args()) == 0 {
+		usage()
+		return
+	}
+
+	// Normalize content argument to not have ending slashes.
+	content := flag.Arg(0)
+	for content[len(content)-1] == '/' {
+		content = content[:len(content)-1]
+	}
+
+	if strings.HasPrefix(content, "clear/") || strings.HasPrefix(content, "clearlinux/") {
+		parts := strings.SplitN(content, "/", 2)
+		content = clearLinuxBaseContent + "/" + parts[1]
+	} else if content == "clear" || content == "clearlinux" {
+		// Query latest version from Clear Linux.
+		res, err := http.Get(clearLinuxLatestURL)
+		if err != nil {
+			log.Fatalf("ERROR: no version passed and couldn't query latest version of Clear Linux: %s", err)
+		}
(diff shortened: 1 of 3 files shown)


swupd-extract/main.go
```diff
@@ -0,0 +1,516 @@
...
+		"-content", content,
+		"-CAfile", cert,
+		"-purpose", "crlsign",
+	)
+	var buf bytes.Buffer
+	cmd.Stdout = &buf
+	cmd.Stderr = &buf
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("signature verification failed: %s\nOUTPUT:\n%s", err, buf.Bytes())
```
icarus-sparry: Maybe worth outputting the command line used when it fails?

swupd-extract/state.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+	err = download(cs.baseContent+"/"+joined, localFile)
+	if err != nil {
+		return "", err
+	}
+	return localFile, nil
+}
+
+func (cs *clientState) Path(elem ...string) string {
+	return filepath.Join(cs.dir, filepath.Join(elem...))
```
icarus-sparry: join of join?

swupd-extract/state.go
```diff
@@ -0,0 +1,367 @@
...
+		return nil, fmt.Errorf("invalid arguments to GetBundleManifest: MoM is not a bundle")
+	}
+	filename, err := cs.GetFile(version, "Manifest."+name)
+	if err != nil {
+		return nil, err
+	}
+	hash, err := swupd.GetHashForFile(filename)
+	if err != nil {
+		_ = os.Rename(filename, filename+".invalid")
+		return nil, fmt.Errorf("couldn't calculate hash for %s: %s", filename, err)
```
icarus-sparry: Some indication that the bad file is left in $filename.invalid ,at least of os.Rename doesn't report an error?

swupd-extract/state.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+	defer func() {
+		if err != nil {
+			_ = os.Remove(tempPath)
+		}
+	}()
+	fmt.Printf("- downloading %s\n", u)
+	res, err := http.Get(u)
+	if err != nil {
+		_ = f.Close()
```
icarus-sparry: Message if http.Get returns an error? You give one if the res.StatusCode is not 200...

swupd-extract/main.go
cmarcelo: Done.

swupd-extract/state.go
cmarcelo: I can't do Join(cs.dir, elem...).

swupd-extract/state.go
cmarcelo: The Get error already contains the URL and the reason of failure.

swupd-extract/state.go
cmarcelo: Makes sense to warn the user.

I'll see if I can also change this to download the file elsewhere and only move it to the final place when happy. That at least diminishes the impact of os.Rename failing.

cmarcelo: > I'll see if I can also change this to download the file elsewhere and only move it to the final place when happy. That at least diminishes the impact of os.Rename failing.

Nevermind my comment. For this particular case, we are downloading to replicate what is externally AND always verifying the file. I'll remove the .invalid rename, as it's the wrong thing to do if we are dealing with local paths (where the file from the local path repository is used directly).

cmarcelo: Removed the os.Rename. If the file is bad, we'll keep telling the user about it.

swupd-extract/state.go
cmarcelo: I've considered also `filepath.Join(append([]string{cs.dir}, elem...)...)` but looks uglier. Maybe I'm missing a simpler solution here?

swupd-extract/main.go
```diff
@@ -0,0 +1,516 @@
+package main
+
+// TODO: Flag to ignore included bundles (only extract the requested). Still need to parse
```
tmarcu: While this ~might~ have some uses, I think you should always add the includes if you're attempting to create a full, usable bundle, or it won't work just installing editors for example. Includes are a space saving abstraction, so I think of includes more as Requires(thing) which reduces our bundle def size and manifest sizes. I would possibly be ok if they were providing the other deps themselves, is that the case this would cover? 
And for skipping packs, probably OK to do.

cmarcelo: > you should always add the includes if you're attempting to create a full, usable bundle

That is the current behavior of the program :)

> I would possibly be ok if they were providing the other deps themselves, is that the case this would cover?

Not really what I was thinking. This program is about extracting content from an existing swupd update repo, it felt like a natural extension if you are using it to perform some kind of validation or inspection or the update repo.

The mindset was more: if you have unzip program, you want to add a feature to just extract a single file out of it.

I didn't have a real use case for this, so just added a note there.

tmarcu: Ah ok, thanks for clarifying! I can see that being a useful feature for checking manifest/pack/etc correctness.

swupd-extract/state.go
tmarcu: I think this is fine... you can split into 
`elems := filepath.Join(elem...)`
``return filepath.Join(cd.dir, elems)` and waste another line of space, I can read it fine either way.

tmarcu:
@cmarcelo @matthewrsj @phmccarty @icarus-sparry @everyonelse, since this is part of server(mixer) and not client, would it make more sense to name this mixer-extract? Or do you want this to be tied specifically to the swupd command itself? I'd expect any swupd related things to  now mean *client*, and *mixer* to represent server work

cmarcelo:
> since this is part of server(mixer) and not client, would it make more sense to name this mixer-extract?

This program reads "swupd update repos", and its not directly related to how mixer work, e.g. I can create such update repositories without mixer.

> I'd expect any swupd related things to now mean client, and mixer to represent server work

I disagree with that direction in naming. If we have tools that work specifically with swupd update repos, it should be fine to call them as such. For example, our package for handling update repos is called "swupd".

matthewrsj:
> I disagree with that direction in naming. If we have tools that work specifically with swupd update repos, it should be fine to call them as such. For example, our package for handling update repos is called "swupd".

That doesn't need to be the case though. Mixer is already tied to swupd-client throughout. Everywhere else that is just an assumption. The `swupd` library could arguably be renamed to something else (`content-gen`? no please no, but something). It was only named `swupd` to tie it to `swupd-server` as a rewrite.

cmarcelo:
I don't see a reason for collapsing the file/repo format name with the tool we use to generating it, unless we are moving towards only being able to generate it from that tool (so it is an "internal format").

For more context: as part of the mkosi and other working, I'm considering generating swupd update repo in ways different than mixer do (no RPMs or yum/dnf involved) for particular use cases.

icarus-sparry:
@cmarcelo Perhaps you can take some time in the next meeting to explain what you think your plans are, so we don't feel blindsided when pull requests like this appear out of the blue?

tmarcu:
> For more context: as part of the mkosi and other working, I'm considering generating swupd update repo in ways different than mixer do (no RPMs or yum/dnf involved) for particular use cases.

In that case there really shouldn't be a swupd in the name since it was not created by anything swupd related :P I don't personally care if it stays swupd or not, but we should start keeping in mind what things we still call 'swupd' and why, since we are getting a clean-slate in a way.

cmarcelo:
> In that case there really shouldn't be a swupd in the name since it was not created by anything swupd related :P

It would be using the swupd package we've been working and generating "swupd update repo" format. (Manifests, fullfiles).

My understanding (which seems at least incomplete from this conversation) is that the "repository format" would still be called "swupd". If that's not the case, I'll rename "swupd-extract" to a suitable name (and whatever else I come up related to that format).

PR #116, opened 2018-01-27
swupd: don't pack hashes that are present in From version

cmarcelo:


swupd/packs.go
```diff
@@ -60,15 +60,21 @@ type PackInfo struct {
...
-	if fromVersion >= m.Header.Version {
-		return nil, fmt.Errorf("fromVersion (%d) smaller than toVersion (%d)", fromVersion, m.Header.Version)
+	toVersion := toManifest.Header.Version
+
+	var fromVersion uint32
+	if fromManifest != nil {
+		fromVersion = fromManifest.Header.Version
+		if fromVersion >= toVersion {
+			return nil, fmt.Errorf("fromVersion (%d) smaller than toVersion (%d)", fromVersion, toVersion)
+		}
```
matthewrsj: You should also return if `fromManifest == nil`, that seems like an error case to me.

swupd/packs.go
```diff
@@ -115,15 +135,15 @@ func WritePack(w io.Writer, m *Manifest, fromVersion uint32, outputDir, chrootDi
...
 	}
 
 	done := make(map[Hashval]bool)
-	for i, f := range m.Files {
+	for i, f := range toManifest.Files {
 		entry := &info.Entries[i]
 		entry.File = f
 
-		if f.Version <= fromVersion {
+		if f.Version <= fromVersion || hashesInFrom[f.Hash] {
```
matthewrsj: `fromVersion` will not be set if `fromManifest == nil`

swupd/packs.go
cmarcelo: "fromManifest == nil" implies we are creating a zeropack, so it is a valid usage. Added a sentence in the comments.

Also added a check for toManifest not being nil early on.

swupd/packs.go
cmarcelo: For zero packs, fromVersion is zero, and that works for this condition because the f.Version will always be greater than zero.

matthewrsj: oh, right, because zero values. +1, thanks for the comment.

PR #117, opened 2018-01-27
WIP: Using deltas in the packs

cmarcelo:
DO NOT MERGE.

This is on top of PR #116, so just look at the last commit. Pushed PR to show the direction I'm taking with the delta integration:

- Create a function to find the deltas, it contains logic that was in linkPeersAndChange and consolidateDeltaFiles. I think that this led to a clearer logic, but has problems (see below).

- I'm experimenting with returning a list, basically because I wanted to see how it would work. This still can be done with putting things on DeltaPeers, but I've felt this was a bit cleaner, because with a Delta struct I have a place to put the error that delta had as well as cache the full path to the delta. Tradeoffs.

- In theory packing already works and there's even a test. But we need to put more test cases.

The problem I have: I don't know where Rename logic will hook up, and whether it will be compatible with the direction I'm taking.

@icarus-sparry what's the approach you are taking to return the result of the rename logic, is it by mutating the existing Files in the Manifests being evaluated, or by building up a separate list?

### DIFF
--- a/swupd/delta.go
+++ b/swupd/delta.go
@@ -10,10 +10,31 @@ import (
 	"github.com/pkg/errors"
 )
 
-// CreateDeltas creates all delta files between the previous and current
-// version of the supplied manifest, returning a list of files that
-// delta creation failed for.
-func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
+const (
+	// From swupd-server's include/swupd.h:
+	//
+	//     Approximately the smallest size of a pair of input files which differ by a
+	//     single bit that bsdiff can produce a more compact deltafile. Files smaller
+	//     than this are always marked as different. See the magic 200 value in the
+	//     bsdiff/src/diff.c code.
+	//
+	minimumSizeToMakeDeltaInBytes = 200
+)
+
+// Delta represents a delta file between two other files. If Error is present, it
+// indicates that the delta couldn't be created.
+type Delta struct {
+	Path  string
+	Error error
+	from  *File
+	to    *File
+}
+
+// CreateDeltas creates all delta files between the previous and current version of the
+// supplied manifest. Returns a list of deltas (which contains information about
+// individual delta errors). Returns error (and no deltas) if it can't assemble the delta
+// list.
+func CreateDeltas(manifest, statedir string, from, to uint32) ([]Delta, error) {
 	var c config
 
 	c, err := getConfig(statedir)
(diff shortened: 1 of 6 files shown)


icarus-sparry:
@cmarcelo I am mutating the files in the new manifest list. Towards the end of linkPeersAndChange we have 2 lists, `removed` and `added`, which I look through to find thing which might be renames, and if so I set the DeltaPeer to point to each other.

swupd/delta.go
```diff
@@ -42,152 +43,123 @@ func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
...
+			// Check file size to decide whether it should have a delta.
+			newPath := filepath.Join(c.imageBase, fmt.Sprint(nf.Version), "full", nf.Name)
+			fi, err := os.Stat(newPath)
 			if err != nil {
 				return nil, errors.Wrapf(err, "error accessing %s to decide whether it can have a delta or not")
 			}
-			if file.Info.Size() < minimumSizeToMakeDeltaInBytes {
-				continue
+			if fi.Size() >= minimumSizeToMakeDeltaInBytes {
+				result = append(result, Delta{From: of, To: nf})
```
icarus-sparry: If we are taking this approach, we don't need DeltaPeer in the File structure.

Whilst I feel that the DeltaPeer doesn't really belong in the File structure, I wonder if this Delta structure is just a little too remote. Really the Delta is a property of two manifests, and we bodge it by saying that files belong to a manifest and the delta information for those files is to the previous manifest.

swupd/delta.go
```diff
@@ -42,152 +43,123 @@ func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
...
+		newIndex++
+		oldIndex++
 	}
 
-	deltahash, err := Hashcalc(testnewfile)
-	if err != nil {
-		return errors.Wrap(err, "Failed to calculate hash for new file")
-	}
+	// TODO: Figure out relationship between deltas and renames. Ideally we want to
+	// call rename detection from here and return more deltas.
```
icarus-sparry: My thought was that the linkPeerAndChange would be doing this walk of the two manifests. If finds all the files which have the same name, produces two lists offiles whose name differs (added and removed), and the rename detection walks these two lists to find things which look as if they are a good match.

See the rename-wip branch for some code which is a couple of weeks old.

cmarcelo: Tks. I’ll take a look!

cmarcelo:
> You can rebase this now that #116 is merged.

Done.

swupd/delta.go
```diff
@@ -31,149 +43,123 @@ func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
...
 	}
-
-	return files, nil
-}
-
-func createDelta(c config, file *File, statedir string, fromVersion uint32, fromHash Hashval) error {
-	// File without peers cannot have deltas
-	if file.DeltaPeer == nil {
-		return nil
+	testHash, err := Hashcalc(testPath)
```
matthewrsj: If the above check passed we know testPath exists on the filesystem. The test file should be removed even in the error cases so `defer os.Remove(testPath)` before this line. I don't see a benefit of keeping it around in the error cases.

swupd/delta.go
```diff
@@ -31,149 +43,123 @@ func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
...
+	return nil
+}
 
-	// Check that the delta actually applies correctly
-	if err := helpers.RunCommandSilent("bspatch", original, testnewfile, deltafile); err != nil {
-		return errors.Wrapf(err, "Failed to apply delta %s", deltafile)
-	}
+// Delta represents a delta file between two other files. If Error is present, it
+// indicates that the delta couldn't be created.
+type Delta struct {
```
matthewrsj: Can you move this definition to the top of the file? You use it in the `createDeltasFromManifest` and `createDelta` functions but I had to scroll all the way down here to see the definition.

swupd/delta.go
```diff
@@ -31,149 +43,123 @@ func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
...
-	if err != nil {
-		return errors.Wrap(err, "Failed to calculate hash for new file")
-	}
+	path string
+}
 
-	if !HashEquals(deltahash, file.Hash) {
-		return errors.Wrapf(err, "Delta mismatch: %s -> %s via delta: %s", original, newfile, deltafile)
+func findDeltas(c *config, oldManifest, newManifest *Manifest) ([]Delta, error) {
+	// TODO: Pre-allocate with the maximum size possible (wasteful) or trust the reallocs?
```
matthewrsj: I'd say trust the reallocs. This is already very memory intensive so we probably don't want to assume worst case here and pre-allocate to the max size.

swupd/delta.go
```diff
@@ -31,149 +43,123 @@ func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
...
+	var result []Delta
+	newIndex := 0
+	oldIndex := 0
+	for newIndex < len(newManifest.Files) && oldIndex < len(oldManifest.Files) {
+		nf := newManifest.Files[newIndex]
+		of := oldManifest.Files[oldIndex]
+
+		if nf.Name < of.Name {
```
matthewrsj: Personal preference and therefore optional: This looks better to me as
```go
if {
...
} else if {
...
} else {
...
}
```
because the conditionals are all related to each other. Two `if`s followed by what is essentially a default case makes it easier to assume you can add code between the `if` statements which can (will?) lead to bugs.

A `switch...case` would also work.

swupd/packs.go
```diff
@@ -435,20 +495,38 @@ func CreatePack(name string, fromVersion, toVersion uint32, outputDir, chrootDir
...
+			return nil, err
+		}
+
+		deltas, err = createDeltasFromManifests(&c, fromM, toM)
+		if err != nil {
+			return nil, err
+		}
+
+		// TODO: remove this.
+		for _, d := range deltas {
```
matthewrsj: I assume this *will* be removed for the final version of this PR?

swupd/delta.go
cmarcelo: If Hashcalc fails, we don't know for sure that the file exists.

I agree that we should remove it regardless. This was just the logic from the previous existing code, I'll change it.

swupd/delta.go
cmarcelo: OK.

swupd/delta.go
cmarcelo: I'll try that approach (switch). Thanks.

swupd/packs.go
cmarcelo: Sure. This PR was more in a WIP kind of stage. I'll remove this when I update it.

swupd/delta.go
cmarcelo: OK, to avoid conflicts with your work, I'll try to keep the "DeltaPeer" setting around and after that build up my list -- a bit different than what the WIP here was doing, but not so much.

swupd/delta.go
cmarcelo: I'll take a middle ground here between the WIP and the current state to make this change less intrusive:

- Keep using DeltaPeer (so zero impact for renaming work you are doing); I'll figure out how to handle avoiding the Version changes in linkedPeers when calculating deltas, as well as resetting the DeltaPeers.
- Make Delta not exported, and use it inside WritePack (so WritePack will call the creation of deltas), this reduces its scope a bit. I could even go as far as say these are not deltas but entries in the delta creation operation. The creation of deltas kind already do 1/3 of that by giving me the failed deltas (as strings and not structured information). The file information is not essential but we use it to be able to pick a fallback.

cmarcelo:
I'm working on an updated version of this.

cmarcelo:
@matthewrsj could you take another look?

Tomorrow I'll try to consolidate the two link functions and add a few more tests, but the rest of the code should be good to go.

swupd/manifest.go
```diff
@@ -467,6 +468,77 @@ func (m *Manifest) newDeleted(df *File) {
 	m.Files = append(m.Files, df)
 }
 
+// TODO: Consolidate this function with linkPeersAndChange.
+
+// TODO: Double check minVersion handling in linkPeersAndChange: if old=20, minversion=30 and
+// new=40, we are always setting to 40, but what if 30 exists?
```
cmarcelo: Double checked. Is not an issue because when setting versions (calling linkPeersAndChange, we are calling with the directly previous Manifest, so if we can't use that version, we use the new manifest one).

I'll remove this TODO.

swupd/packs.go
```diff
@@ -142,6 +174,27 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
...
+	hasDelta := make(map[Hashval]bool)
+	for i := range deltas {
+		d := &deltas[i]
+		if d.Error != nil {
+			info.Warnings = append(info.Warnings, err.Error())
+			continue
+		}
+		var fallback bool
+		fallback, err = copyFromDelta(tw, d)
+		if fallback {
```
matthewrsj: Took me a while to get the logic of this function with `fallback` here and `fallback` below but here using `hasDelta[d.to.Hash]` if `fallback == false`. It would be helpful to add a few comments to the code explaining that if the copy fails you will try another method below (in the next loop).

swupd/packs_test.go
```diff
@@ -254,6 +254,50 @@ func TestCreatePackZeroPacks(t *testing.T) {
...
+		format = 1
+		minVer = 0
+	)
+
+	//
+	// In version 10, create a bundle with files of different sizes.
+	//
+	emptyContents := ""
+	smallContents := "small"
+	largeContents := strings.Repeat("large", 1000)
```
matthewrsj: `strings.Repeat`... Ah, handy.

matthewrsj: @tmarcu better than `lorem ipsum...` 😆

swupd/packs_test.go
```diff
@@ -254,6 +254,50 @@ func TestCreatePackZeroPacks(t *testing.T) {
...
+	//
+	mustInitStandardTest(t, fs.Dir, "10", "20", []string{"contents"})
+	fs.write("image/20/contents/small1", smallContents)
+	fs.write("image/20/contents/small2", smallContents)
+	fs.write("image/20/contents/large1", strings.ToUpper(largeContents[:1])+largeContents[1:])
+	fs.write("image/20/contents/large2", largeContents[:1]+strings.ToUpper(largeContents[1:]))
+	mustCreateManifests(t, 20, minVer, format, fs.Dir)
+
+	// TODO: Move directory creation somewhere else.
+	fs.mkdir("www/20/delta")
```
matthewrsj: Delta creation should definitely do this.

cmarcelo: I did some elaborated way to perform only the amount of mkdirs necessary, but then dropped and now I'm creating the directory when we create the new version, that way we can just assume all versions have a delta directory. Would that work?

matthewrsj: Yes that's fine. Output directory structure should look the same every time imo. That way people can do things like `ls <path-to>/delta | wc -l` for a simple way to find the number of delta files without the command failing.

swupd/delta.go
```diff
@@ -31,149 +52,101 @@ func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
...
-		// Drains the failedChan appropriately
-		for f := range failedChan {
-			failed = append(failed, f)
-		}
-		failedList <- failed // This list is what we return in the end
-	}()
-
 	// Don't flood the system with goroutines, delta creation takes up an
-	// incredibly amount of memory, so we cannot max out on goroutines
+	// incredibly amount of memory, so we cannot max out on goroutines.
```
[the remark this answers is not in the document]
cmarcelo: Rewrote the phrase to be smaller.

swupd/packs.go
cmarcelo: Reorganized the code to show the relationship between fallback and err in this case, added a comment here and in the other place we use fallback logic.

cmarcelo:
Closing to reopen a non-WIP one.

PR #118, opened 2018-01-29
builder: fix generation of versions file when using dnf

cmarcelo:
Add one more entry to be skipped. This fixes the scenario of using dnf
by changing yum to be a symlink to it.


tmarcu:
+1

PR #119, opened 2018-01-29
mixer: accept --packager flag when building chroots

cmarcelo:
This lets override the default choice of packager for performing build
chroots. Makes easy to test dnf with --packager=dnf. The flag is
ignored when not using --new-chroots.
    
Also prints the actual command line being used.

### DIFF
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -35,6 +35,9 @@ type buildCmdFlags struct {
 	noPublish  bool
 	keepChroot bool
 	template   string
+
+	// Name of the program to use instead of yum (or dnf in Fedora >= 22).
+	packager string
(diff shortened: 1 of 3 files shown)


mixer/cmd/build.go
```diff
@@ -35,6 +35,10 @@ type buildCmdFlags struct {
 	noPublish  bool
 	keepChroot bool
 	template   string
+
+	// TODO: This deserves to be in the config too.
```
matthewrsj: Why add this to the config when `--packager` will only be used until we switch to dnf entirely?

cmarcelo: This is less valuable if the switch is imminent. Will remove it then.

PR #120, opened 2018-01-30
** DO NOT MERGE ** : Use Viper to parse config file

rchiossi:
This is just a preview of how it will look like to have Viper as our config parser for mixer. This does not replace go-ini in swupd code due to some reasons described below. This PR also doesn't include the files from `dep ensure`

One of the main things I want to point out with this preview is how the defaults are initialized. If you look at the builder only, it works just fine. When you start looking at values that are only used by swupd, things start to get a bit ugly. Right now in the main program, every time swupd reads a config value it so happens to be after a builder object has been initialized, but there is no guarantee that this will always be the case. For instance, the tests for swupd never use a builder object. This is one of the main reasons why I believe it would be healthy to separate config parsing from the builder object.

Some notable changes with the move:
- The use of Environment Variables also changes a bit since now nested values are properly parsed. E.g. `CONTENTURL="..."` becomes `SWUPD_CONTENTURL="..."`
- Since viper determines the type of config based on its extension, `builder.conf` needs to be renamed `builder.toml`
- There is no check to see if all required values are present in the config file since there is always a default value that will be used as fallback

On a side note, I found an issue in the Viper API to save a config file without overwriting an existing one. I sent a PR to Viper, I'm waiting for review.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -26,13 +26,13 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 
 	"github.com/clearlinux/mixer-tools/helpers"
 	"github.com/clearlinux/mixer-tools/swupd"
 	"github.com/pkg/errors"
+	"github.com/spf13/viper"
 )
 
 // Version of Mixer. Also used by the Makefile for releases.
@@ -87,16 +87,47 @@ func NewFromConfig(conf string) (*Builder, error) {
 	if err := b.LoadBuilderConf(conf); err != nil {
 		return nil, err
 	}
-	if err := b.ReadBuilderConf(); err != nil {
-		return nil, err
-	}
 	if err := b.ReadVersions(); err != nil {
 		return nil, err
 	}
 	return b, nil
 }
 
-// CreateDefaultConfig creates a default builder.conf using the active
+// LoadDefaults set default values for the properties in builder.toml
+func (b *Builder) LoadDefaults() error {
+	pwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	viper.SetDefault("Builder.SERVER_STATE_DIR", filepath.Join(pwd, "update"))
+	viper.SetDefault("Builder.BUNDLE_DIR", filepath.Join(pwd, "mix-bundles"))
+	viper.SetDefault("Builder.YUM_CONF", filepath.Join(pwd, ".yum-mix.conf"))
+	viper.SetDefault("Builder.CERT", filepath.Join(pwd, "Swupd_Root.pem"))
+	viper.SetDefault("Builder.VERSIONS_PATH", pwd)
+
+	viper.SetDefault("swupd.BUNDLE", "os-core-update")
+	viper.SetDefault("swupd.CONTENTURL", "<URL where the content will be hosted>")
+	viper.SetDefault("swupd.VERSIONURL", "<URL where the version of the mix will be hosted>")
+	viper.SetDefault("swupd.FORMAT", "1")
+
+	viper.SetDefault("Server.debuginfo_banned", "true")
+	viper.SetDefault("Server.debuginfo_lib", "/usr/lib/debug")
+	viper.SetDefault("Server.debuginfo_src", "/usr/src/debug")
+
+	viper.SetDefault("Mixer.RPMDIR", "")
+	viper.SetDefault("Mixer.REPODIR", "")
+
+	viper.SetConfigName("builder")
+	viper.AddConfigPath(pwd)
+
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
+
+	return nil
+}
+
+// CreateDefaultConfig creates a default builder.toml using the active
 // directory as base path for the variables values.
 func (b *Builder) CreateDefaultConfig(localrpms bool) error {
 	pwd, err := os.Getwd()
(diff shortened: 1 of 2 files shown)


rchiossi:
@cmarcelo Any opinion on the "keep config inside builder.go or in a separate thing" issue above?

cmarcelo:
> Right now in the main program, every time swupd reads a config value it so happens to be after a builder object has been initialized, but there is no guarantee that this will always be the case. 

Note that swupd is a bit at fault here IMHO. We should pass any configuration to the functions, so that getConfig will eventually be gone. So when calling functions in swupd package we either pass arguments directly our a swupd.Config struct of sorts, that we can fill after configuration happens.

swupd package should never need to use Viper types directly.

cmarcelo:
> The use of Environment Variables also changes a bit since now nested values are properly parsed. E.g. CONTENTURL="..." becomes SWUPD_CONTENTURL="..."

Could we disable use of environment variables at least for now? (I'm understanding that you can set something via conf file, flags, and also environemnet variables)

It seems having command line and configuration file will cover pretty well our needs and would not cause restrictions to the key names.

cmarcelo:
> There is no check to see if all required values are present in the config file since there is always a default value that will be used as fallback

This is kind of bad since we get this check for command line and for existing ini parsing, but I guess we can work around it.

builder/builder.go
```diff
@@ -87,51 +87,62 @@ func NewFromConfig(conf string) (*Builder, error) {
...
+
+	viper.SetDefault("Mixer.RPMDIR", "")
+	viper.SetDefault("Mixer.REPODIR", "")
+
+	viper.SetConfigName("builder")
+	viper.AddConfigPath(pwd)
 
-	fmt.Println("Creating new builder.conf configuration file...")
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
```
cmarcelo: Probably remove these two?

cmarcelo:
> I sent a PR to Viper, I'm waiting for review

Nice!

cmarcelo:
> This is one of the main reasons why I believe it would be healthy to separate config parsing from the builder object.

This is fine, my point is that you don't need a package for that (if need arises, we can create). Just create a type Config struct and keep moving.

builder/builder.go
```diff
@@ -154,85 +165,27 @@ func (b *Builder) createRpmDirs() error {
...
-	if err != nil {
-		return err
-	}
-
-	// If builderconf is set via cmd line, use that one
 	if len(builderconf) > 0 {
-		b.Buildconf = builderconf
-		return nil
+		viper.AddConfigPath(filepath.Dir(builderconf))
+		viper.SetConfigName(filepath.Base(builderconf))
```
cmarcelo: viper.SetConfigFile?

cmarcelo:
To be honest, I'm not totally sold on changing to toml. But assuming that's the direction, I think this is a good start. Please send a new PR with the vendored stuff and etc.

Note: I thought there was an INI parser as part of Viper, but couldn't find in the docs. Do they have one?

rchiossi:
> Question: could we disable use of environment variables at least for now? (I'm understanding that you can set something via conf file, flags, and also environemnet variables)
It seems having command line and configuration file will cover pretty well our needs and would not cause restrictions to the key names.

Maybe my comment was a bit misleading. In the config file, CONTENTURL is still called CONTENTURL. But if you want to pass it as an envvar, then you would use SWUPD_CONTENTURL in the command line only. Keeping it enabled has no influence in the rest of the code, so I don't se a reason do to it.

rchiossi:
>> There is no check to see if all required values are present in the config file since there is always a default value that will be used as fallback
>This is kind of bad since we get this check for command line and for existing ini parsing, but I guess we can work around it.

I'm not sure I agree that this is a bad thing. The reason the check existed in the first place was to make sure a valid value existed. Given we now always have a valid value, the checkis not needed anymore. Also , it allow simpler config files to exist since you can omit values that will stay default.

builder/builder.go
rchiossi: Indeed. Nice catch.

cmarcelo:
> Keeping it enabled has no influence in the rest of the code, so I don't se a reason do to it.

Do we have any reason to keep this enabled (allowing environment variables to override values in the configuration). It seems to me right now it might make things more complicated -- i.e. we'll have to document that somewhere. I'd shelve it for now and revisit later.

cmarcelo:
> I'm not sure I agree that this is a bad thing. The reason the check existed in the first place was to make sure a valid value existed. Given we now always have a valid value, the checkis not needed anymore. Also , it allow simpler config files to exist since you can omit values that will stay default.

It makes awkward to distinguish unset from default values (you need to workaround to get it). It is good that we don't need that feature now (because we have good default values everywhere), but it is bad that the library doesn't solve that one for us. (which is surprising given pflags/cobra handle that case)

In case it wasn't clear: this is not a deal breaker. Keep walking :-)

kevincwells:
@cmarcelo 
>To be honest, I'm not totally sold on changing to toml. 

Do you have specific concerns with using TOML? Technically it is, right now, valid TOML.

>Note: I thought there was an INI parser as part of Viper, but couldn't find in the docs. Do they have one?

It doesn't, because generally speaking TOML is a superset of INI*, so it can parse INI files by treating them as TOML

\* Technically there is no official spec for INI, which is part of why people hate it. Some implementations support `;` as comments, others `#`, and many don't support comments at all. Some support comma-separated values as lists, while most don't support lists at all. And none support nested sections. TOML is, essentially, "what INI was supposed to be".

cmarcelo:
> It doesn't, because generally speaking TOML is a superset of INI*, so it can parse INI files by treating them as TOML

That was the information I've forgot, one of you already explained that to me.

PR #128, opened 2018-01-31
Rename wip

icarus-sparry:
Time for a code review.

### DIFF
--- a/swupd/rename.go
+++ b/swupd/rename.go
@@ -1,0 +1,187 @@
+// Copyright 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package swupd
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// BSDIFFSIZE is smallest size of file to consider using bsdiff
+const BSDIFFSIZE = 200
+
+func renameDetection(manifest *Manifest, added []*File, removed []*File, c config) {
+	if len(added) == 0 || len(removed) == 0 {
+		return // nothing to rename
+	}
+	added = trimRenamed(added) // Make copies of input slices, tidy up whilst we are here
+	removed = trimRenamed(removed)
+	if err := fixupStatFields(removed, manifest, &c); err != nil {
+		panic(err)
+	}
+	// Handle pure renames first, don't need to worry about size. Should we skip zero size?
+	// just add call to trimSmall if so
+	// Sort by Hash. No need to have tiebreaker on name as the Manifest links by hash
+	// so 2 identical files being renamed, e.g. python2.6/foo and python3.7/bar being
+	// renamed to python2.7/foo and python3.7/bar have no concept which is the source for the
+	// python2.7/foo. This decision might need to change if the way renames are shown in
+	// the manifest changes.
+	sort.Slice(added, func(i, j int) bool {
+		return added[i].Hash < added[j].Hash
+	})
+	sort.Slice(removed, func(i, j int) bool {
+		return removed[i].Hash < removed[j].Hash
+	})
+	for ax, rx := 0, 0; ax < len(added) && rx < len(removed); {
+		af := added[ax]
+		rf := removed[rx]
+		switch {
+		case af.Hash < rf.Hash:
+			ax++
+		case af.Hash > rf.Hash:
+			rx++
+		default: // Equal hash, so link
+			linkRenamePair(af, rf)
+			ax++
+			rx++
+		}
+	}
+	// Link things with the same names skipping digits
+	// First remove small files (not worth sending diff) and files which have
+	// exact match
+	added = trimSmall(trimRenamed(added), BSDIFFSIZE)     // Make it explicit we are doing two steps
+	removed = trimSmall(trimRenamed(removed), BSDIFFSIZE) // TODO. make it one pass.
+	//generate the pairs of *File and short name
+	pa := makePairedNames(added)
+	pr := makePairedNames(removed)
+	// Merge where short names match
+	for ax, rx := 0, 0; ax < len(pa) && rx < len(pr); {
+		af := pa[ax]
+		rf := pr[rx]
+		switch {
+		case af.partialName < rf.partialName:
+			ax++
+		case af.partialName > rf.partialName:
+			rx++
+		default: // Equal truncated name
+			linkRenamePair(af.f, rf.f)
+			ax++
+			rx++
+		}
+	}
+}
+
+// linkRenamePair links two files together
+func linkRenamePair(renameTo, renameFrom *File) {
+	renameTo.DeltaPeer = renameFrom
+	renameFrom.DeltaPeer = renameTo
+	renameTo.Rename = true
+	renameFrom.Rename = true
+	renameFrom.Type = TypeUnset
+}
+
+// trimRenamed returns an slice which has had files with DeltaPeers purged
+func trimRenamed(a []*File) []*File {
+	r := make([]*File, 0, len(a))
+	for _, f := range a {
+		// Do not worry about ghosted files here, assume that they will
+		// exist on the target system, and so can be used as the source for
+		// renames. The classic ghosted files are the OS kernels and modules,
+		// and it is helpful to be able to ship just the differences for them.
+		// Worst case is that the files do not exist on the target system, in which
+		// case the swupd-client will fall back to doing a full download.
+		if f.DeltaPeer == nil {
+			r = append(r, f)
+		}
+	}
+	return r
+}
+
+// trimsmall returns an slice which has had files with small files removed
+func trimSmall(a []*File, minsize int64) []*File {
+	r := make([]*File, 0, len(a))
+	for _, f := range a {
+		if f.Info == nil {
+			panic("Missing f.Info for " + f.Name)
+		}
+		if f.Info.Size() > minsize {
+			r = append(r, f)
+		}
+	}
+	return r
+}
+
+// pairednames holds a *File and another "name", which is used for various
+// alterations of the filename for matching purposes. In particular removing
+// digits so a new release of a shared library is matched to previous ones
+type pairedNames struct {
+	f           *File
+	partialName string // Filename with digits removed
+}
+
+// stripDigits is used with strings.Map to remove digits from filename
+func stripDigits(r rune) rune {
+	if r >= '0' && r <= '9' {
+		return -1 // strings.Map removes negative values
+	}
+	return r
+}
+
+// makePairedNames returns `a` sorted by name, disregarding digits.
+// The secondary sort key is the original name. This is mainly intended for
+// cases like with python we have the same filename under both python3.6 and python2.7
+// it would look weird to rename the old python2.7/file1 to python3.6/file2 and at
+// the same time rename the old python3.6/file1 to python2.7/file2.
+func makePairedNames(list []*File) []pairedNames {
+	pairs := make([]pairedNames, len(list))
+	for i, f := range list {
+		pairs[i].f = f
+		pairs[i].partialName = strings.Map(stripDigits, f.Name)
+	}
+	sort.Slice(pairs, func(a, b int) bool {
+		if pairs[a].partialName == pairs[b].partialName { // Same stripped name, sort on original name
+			// mainly for python2.x vs 3.x
+			return pairs[a].f.Name < pairs[b].f.Name
+		}
+		return pairs[a].partialName < pairs[b].partialName
+	})
+	return pairs
+}
+
+// fixupstatfields adds the missing stat fields
+// construct the path to the old chroot by Joining the c.imageBase
+// file.Previous, bundle name, and file.Name fields
+// Note this is horrible
+func fixupStatFields(needed []*File, m *Manifest, c *config) error {
+	var bundleChroot string
+	for i := range needed {
+		if needed[i].Info != nil {
+			continue
+		}
+		if bundleChroot == "" {
+			bundleChroot = filepath.Join(c.imageBase, fmt.Sprint(m.Header.Previous), "full")
+		}
+		path := filepath.Join(bundleChroot, needed[i].Name)
+		fi, err := os.Stat(path)
+		if err != nil {
+			return err
+		}
+		needed[i].Info = fi
+	}
+	return nil
+}
--- a/swupd/rename_test.go
+++ b/swupd/rename_test.go
@@ -1,0 +1,22 @@
+package swupd
+
+import (
+	"os"
+)
+
+// Need to set up a type to hold the FileInfo
+// as we are not holding them on disk
+type mockinfo struct {
+	os.FileInfo // Embedded but only used to get things like Name() method
+	size        int64
+}
+
+// Here is the method to override the FileInfo.Size() for mockinfo
+func (m mockinfo) Size() int64 {
+	return m.size
+}
+
+// Constructor for os.FileInfo using mockinfo
+func sizer(s int64) os.FileInfo {
+	return mockinfo{size: s}
+}
--- a/swupd/rename_test.go
+++ b/swupd/rename_test.go
@@ -2,6 +2,8 @@ package swupd
 
 import (
 	"os"
+	"syscall"
+	"testing"
 )
 
 // Need to set up a type to hold the FileInfo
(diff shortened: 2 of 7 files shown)


swupd/rename.go
```diff
@@ -0,0 +1,183 @@
...
+// it would look weird to rename the old python2.7/file1 to python3.6/file2 and at
+// the same time rename the old python3.6/file1 to python2.7/file2.
+func makepairs(a []*File) []pairednames {
+	pair := make([]pairednames, len(a))
+	for i, f := range a {
+		pair[i].f = f
+		pair[i].n = strings.Map(stripdigits, f.Name)
+	}
+	sort.Slice(pair, func(i, j int) bool {
+		if pair[1].n == pair[j].n { // Same striped name, sort on original name
```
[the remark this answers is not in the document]
matthewrsj: why `pair[1]`? Should it be `pair[i]`?

swupd/rename.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	return r
+}
+
+// makepairs returns a sorted by name, disregarding digits
+// secondary sort key is the original name. This is mainly inteneded for
+// cases like with python we have the same filename under both python3.6 and python2.7
+// it would look weird to rename the old python2.7/file1 to python3.6/file2 and at
+// the same time rename the old python3.6/file1 to python2.7/file2.
+func makepairs(a []*File) []pairednames {
```
matthewrsj: I'm not really understanding the `makepairs` name here since it doesn't pair any filenames here. Do you instead mean `makepares` and `parednames` or something since you are kind of paring the filenames?

swupd/rename.go
```diff
@@ -0,0 +1,183 @@
...
+			continue
+		}
+		if bundleChroot == "" {
+			bundleChroot = filepath.Join(c.imageBase, fmt.Sprint(m.Header.Previous), m.Name)
+		}
+		path := filepath.Join(bundleChroot, needed[i].Name)
+		fi, err := os.Stat(path)
+		if err != nil {
+			panic(err)
+			return err
```
matthewrsj: you `panic` and `return` here, then `panic` in the caller as well.

swupd/rename_test.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		f.f.Hash = internHash(hv)
+		f.f.Info = sizer(f.s) // A FileInfo that returns our desired size
+		sn[f.n] = f.f
+	}
+	return sn
+}
+
+// filelist takes a list of wanted shortnames, looks them up in sn, and returns
+// the a list of pointers to copies of the Files. Because they are copies they can
```
matthewrsj: remove `the` or `a` at the beginning of this line.

swupd/rename_test.go
```diff
@@ -0,0 +1,156 @@
...
+func markdelete(a []*File) []*File {
+	for i := range a {
+		a[i].Status = StatusDeleted
+	}
+	return a
+}
+
+// checkLinked tests forward and backward pairing of DeltaPeers
+func checkLinked(t *testing.T, a, b *File, description string) bool {
+	if a.DeltaPeer != b {
```
matthewrsj: You check if b is `nil` below but not if `a` is nil here?

swupd/rename_test.go
```diff
@@ -0,0 +1,156 @@
...
+	// 	fmt.Printf("%s %v\n", f, sn[f])
+	// }
+	tests := []struct {
+		description string
+		remove      []string // shortnames of File to remove
+		add         []string // shortnames of File to add
+		partner     []int    // Indexes of add that should match remove
+	}{
+		{description: "Rename of file, same contents",
+			remove: []string{"I1", "I2"}, add: []string{"I2"}, partner: []int{0, -1}},
```
matthewrsj: I1 and I2 are removed but I2 is also added? When could this be possible?....... Oh, shortnames. Also forgot that `sn` is being generated.

I'm leaving this so you can see my thought process, no action needed.

swupd/rename_test.go
```diff
@@ -0,0 +1,156 @@
...
+			remove: []string{"S1", "L3", "L1"}, add: []string{"L2", "L4"},
+			partner: []int{-1, 1, 0}},
+		{description: "Rename files pairwise longer reversed should link",
+			remove: []string{"S1", "L3", "S2", "L1"}, add: []string{"L4", "L2"},
+			partner: []int{-1, 0, -1, 1}},
+	}
+	for _, tc := range tests {
+		add := filelist(t, sn, tc.add)
+		remove := markdelete(filelist(t, sn, tc.remove))
+		// fmt.Println(add)
```
matthewrsj: debug print.

swupd/rename_test.go
```diff
@@ -0,0 +1,156 @@
...
+	if b != nil && b.DeltaPeer != a {
+		t.Errorf("%s Incorrect Linkage reverse %v -> %v, got %v", description, b, a, b.DeltaPeer)
+		return false
+	}
+	return true
+}
+
+func TestRename(t *testing.T) {
+	sn := genTestArray(t)
+	// for f := range sn {
```
matthewrsj: remove

swupd/rename.go
```diff
@@ -0,0 +1,183 @@
...
+	b.DeltaPeer = a
+	a.Rename = true
+	b.Rename = true
+}
+
+// trimrenamed returns an slice which has had files with DeltaPeers purged
+func trimrenamed(a []*File) []*File {
+	r := make([]*File, 0, len(a))
+	for _, f := range a {
+		if f.Status == StatusGhosted {
```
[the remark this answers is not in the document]
icarus-sparry: Interesting. Perhaps I need to think a little more about what a ghosted file can be.

We certainly need to add a test case for this.

My reasoning is that a ghosted file shouldn't be a candidate for a renamefrom, as the lifetime of the file is not managed by swupd-client. In particular a ghosted file could be deleted a minute after it is installed. I agree that a `swupd verify --fix` would find it missing, and reinstall it, which could be annoying. An example I can think of is kernel modules being removed from /usr/lib/modules when they have been added to an initrd, which is stored on another partition.

Since you added the ghosted code, do you think that there is a requirement for the ghosted files to exist until there is an update? Put it another way, do you think 'ghosted' means that the files are already dead, or just that we don't have to kill them?

Of course we shouldn't panic here, just skip the file.

swupd/rename.go
icarus-sparry: You can tell I was getting frustrated with the lack of Info! This panic removed. I need to remove the other one soon.

swupd/rename.go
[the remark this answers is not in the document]
matthewrsj: The whole point of the ghosted file is to enable renames for kernel files. I think currently we actually deprecate old ghosted files immediately after every build, which is the wrong answer (I will add a unit test to check this and fix if necessary). In that case the client would attempt a retry but fail and proceed as normal - we do know that is the current state of the client.

Ghosted only means we don't have to kill them on the client.

swupd/rename.go
```diff
@@ -0,0 +1,187 @@
...
+			// TODO. Do we allow ghosted files to be source files for renames?
+			// I think not, as we don't manage their lifetimes, so they could
+			// be deleted. The counter argument is that rename is only an optimization
```
matthewrsj: The counter argument is essentially my argument. The whole point of ghosted files is to allow renames with them

icarus-sparry: OK, I will allow ghosted files as sources for renames. It makes the code shorter, even if it makes the code longer.

matthewrsj:
> I will add a unit test to check this and fix if necessary

I'll add it with this PR as a base.

cmarcelo:
> ...but the rename code wants to know the size of the files to see if it makes sense to use a deleted
file as the source of a rename.

Questions: does it ever make sense to use a deleted file as a source of rename? Also, isn't checking the Status flag enough?

matthewrsj:
> does it ever make sense to use a deleted file as a source of rename?

Deleted files are the only source of renames so it always makes sense to use a deleted file as the source, unless it is too small.

> Also, isn't checking the Status flag enough?

Not if the file is under 200 bytes. Experimentation with bsdiff in the old swupd-server showed us that bsdiff does not perform well with files under that threshold, sometimes even resulting in segfaults I believe.

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+		af := added[ax]
+		rf := removed[rx]
+		switch {
+		case af.Hash < rf.Hash:
+			ax++
+		case af.Hash > rf.Hash:
+			rx++
+		default: // Equal hash, so link
+			linkrenamepair(af, rf)
+			ax++
```
tmarcu: This means 'added index' I assume?

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+			rx++
+		}
+	}
+	// Link things with the same names skippingdigits
```
tmarcu: s/skippingdigits/skipping digits ?

swupd/rename.go
matthewrsj: yeah

icarus-sparry: I said in one of my code reviews that using `i` and `j` wasn't very helpful as the two index values, as there was no connect in my mind between `i` and one of the added and removed slices, and therefore I needed to look very hard to see if the correct one was being incremented. Go programs traditionally use short variable names, and `i` is fine for a single index, but for this loop where we are stepping down two lists looking for matches I feel we need something better than `i` and `j`. Hence `ax` and `rx` for the added and removed index.

swupd/rename.go
icarus-sparry: Will add the missing space in "skippingdigits".

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+// linkrenamepair links two files together
+func linkrenamepair(a, b *File) {
+	a.DeltaPeer = b
+	b.DeltaPeer = a
+	a.Rename = true
+	b.Rename = true
+}
+
+// trimrenamed returns an slice which has had files with DeltaPeers purged
+func trimrenamed(a []*File) []*File {
```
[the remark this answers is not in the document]
icarus-sparry: Yes. My C background is showing! Please continue to call these out.

swupd/rename.go
tmarcu: Es ok, I figured it out pretty quickly. I just have PTSD from a different codebase using completely terrible names, so when I see 1-2 letter vars I look into it too hard :hammer:  Was mainly clarifying for myself/anyone else.

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// BSDIFFSIZE is smallest size of file to consider using bsdiff
+const BSDIFFSIZE = 200
```
cmarcelo: Optional: I've added a similar constant (but with a more detailed name and background story) in the delta integration PR. If that PR lands first, reuse it.

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+
+func renameDetection(manifest *Manifest, added []*File, removed []*File, c config) {
+	if len(added) == 0 || len(removed) == 0 {
+		return // nothing to rename
+	}
+	added = trimrenamed(added) // Make copies of input slices, tidy up whilst we are here
+	removed = trimrenamed(removed)
```
cmarcelo: I don't get the tidying up here. Both lists shouldn't have any DeltaPeer at this point. Maybe you just need a function to copy the slices (or copy them inline).

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+			linkrenamepair(af, rf)
+			ax++
+			rx++
+		}
+	}
+	// Link things with the same names skippingdigits
+	// First remove small files (not worth sending diff) and files which have
+	// exact match
+	added = trimsmall(trimrenamed(added), BSDIFFSIZE)     // Make it explicit we are doing two steps
+	removed = trimsmall(trimrenamed(removed), BSDIFFSIZE) // Can trivially make it one pass.
```
cmarcelo: Agree with your comment. Do that in a single pass, otherwise you allocate space twice for both trims.

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+func makePairedNames(list []*File) []pairednames {
+	pairs := make([]pairednames, len(list))
+	for i, f := range list {
+		pairs[i].f = f
+		pairs[i].n = strings.Map(stripdigits, f.Name)
+	}
+	sort.Slice(pairs, func(a, b int) bool {
+		if pairs[a].n == pairs[b].n { // Same stripped name, sort on original name
+			// mainly for python2.x vs 3.x
+			return pairs[a].f.Name < pairs[b].f.Name
```
cmarcelo: Wondering if would be sufficient to use sort.SliceStable in this case. (If is just the Python case looks so)

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+			rx++
+		}
+	}
+	// Link things with the same names skippingdigits
+	// First remove small files (not worth sending diff) and files which have
+	// exact match
+	added = trimsmall(trimrenamed(added), BSDIFFSIZE)     // Make it explicit we are doing two steps
+	removed = trimsmall(trimrenamed(removed), BSDIFFSIZE) // Can trivially make it one pass.
+	//generate the pairs of *File and short name
+	pa := makePairedNames(added)
```
cmarcelo: Optional: the section from here to the end of the function is a strong candidate to have some unit test cases, so I would extract it into a separate function and have some concrete tests. I know there are wider integration tests, but for more "hairy" cases of this logic, a test without the rest of the luggage might come handy.

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	return r
+}
+
+// pairednames holds a *File and another "name", which is used for various
+// alterations of the filename for matching purposes. In particular removing
+// digits so a new release of a shared library is matched to previous ones
+type pairednames struct {
+	f *File
+	n string // Filename with digits removed
```
cmarcelo: Optional: would call it 'short' or 'shortName'. With all things equal, 'n' sounds like more a number than a reference to the filename without the digits.

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	return r
+}
+
+// trimsmall returns an slice which has had files with small files removed
+func trimsmall(a []*File, minsize int64) []*File {
+	r := make([]*File, 0, len(a))
+	for _, f := range a {
+		if f.Info == nil {
+			panic("Missing f.Info for " + f.Name)
```
cmarcelo: Return error, and let the caller return the error up.

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+// construct the path to the old chroot by Joining the c.imageBase
+// file.Previous, bundle name, and file.Name fields
+func fixupstatfields(needed []*File, m *Manifest, c *config) error {
+	var bundleChroot string
+	for i := range needed {
+		if needed[i].Info != nil {
+			continue
+		}
+		if bundleChroot == "" {
+			bundleChroot = filepath.Join(c.imageBase, fmt.Sprint(m.Header.Previous), m.Name)
```
cmarcelo: It seems you can set bundleChroot (unconditionally) outside the loop before it starts.

swupd/rename.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+	}
+}
+
+// linkrenamepair links two files together
+func linkrenamepair(a, b *File) {
+	a.DeltaPeer = b
+	b.DeltaPeer = a
+	a.Rename = true
+	b.Rename = true
```
matthewrsj: Need to set `b.Type = TypeUnset` to match current swupd-server behavior. Probably rename `a` and `b` to `renameTo` and `renameFrom` or similar to avoid confusion as well.

swupd/rename_test.go
```diff
@@ -0,0 +1,151 @@
...
+			remove: []string{"S1", "L1"}, add: []string{"L2"}, partner: []int{-1, 0}},
+		{description: "Rename of long files pairwise, should link",
+			remove: []string{"S1", "L1", "L3"}, add: []string{"L2", "L4"},
+			partner: []int{-1, 0, 1}},
+		{description: "Rename of long files pairwise reversed should link",
+			remove: []string{"S1", "L3", "L1"}, add: []string{"L2", "L4"},
+			partner: []int{-1, 1, 0}},
+		{description: "Rename files pairwise longer reversed should link",
+			remove: []string{"S1", "L3", "S2", "L1"}, add: []string{"L4", "L2"},
+			partner: []int{-1, 0, -1, 1}},
```
cmarcelo: Optional: it is easy to get confused with partner indices. I would consider making the expected result look more like "L3->L4 L1->L2" or ["L3->L4", "L1->L2"]. It is slightly more work in the test mechanics but I think pays off later.

swupd/rename_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+// generateTestArray creates some File structures with shortnames and
+// returns a map of them.
+func generateTestArray(t *testing.T) map[string]File {
+	// Set up an array of File with short names
+	const (
+		_rwx = syscall.S_IFREG + 0644
+	)
+	files := []struct {
+		n string // short name for the tests
```
cmarcelo: This ended up being confusing, because there's also the notion of "shortname" above that is the name without digits.

Optional: I think the indirection here doesn't add much. I would explore one of those two venues.

(a) Instead of having generateTestArray, have a "createTestFile" (with a better name), that takes the parameters and creates a *File (or something else) with the properties you want. Then in the top of the test you create those assigning variable names that could act as your short name. The arrays would use those variables instead of strings.
(b) Use the real filenames in the test. The testcase entries will be multiline, but I think its fine.

swupd/rename.go
icarus-sparry: No, this is a logic error. The information needs to be passed in. This is like assert() in C, it is a "shouldn't happen" case.

swupd/rename.go
icarus-sparry: OK.

swupd/rename.go
icarus-sparry: Copy inline would work, I was just reusing something I had around....

swupd/rename.go
tmarcu: The panic happens here because the bundle-chroot does not exist anymore when this happens, and only "full" is available. This should always use "full" instead of m.Name since not everyone will build with --keep-chroots flag on, so there won't be a dir to stat() anymore. :raised_hands:

swupd/rename_test.go
icarus-sparry: I agree it is easy to get confused with partner indicies, and yes, I did think about expressing the renames differently. I will try implementing it and see how it goes.

swupd/rename_test.go
icarus-sparry: My intention with the generateTestArray is to provide a pool of names that tests can draw on, rather than having each test create their own. I look at some of the manifest tests for example and it seems like 90% of the test is setup. It probably needs more tests to be written to make the advantage clear.

Perhaps I should set the field which has the name without digits to be partialName, which fits in better with another use case for the field as well.

Initially I did use the full names in the tests, but they made the tests long enough that my eyes started glazing over.  When you have several similar but different names I found the L1, L2 type short names much easier to reason about.

swupd/rename.go
matthewrsj: This appears to be resolved. No more panics for me.

icarus-sparry: Just did what @tmarcu told me to do!

Just pushing a rebase to fix the delta.go conflict

matthewrsj:
The last commit looks wrong.

icarus-sparry:
You are welcome to tell me why.

matthewrsj:
Because it is the commit message from my swupd index commit but the content of the commit from my rename tests pr

matthewrsj:
And now that I'm looking at it the "Add CreateManifests tests..." Commit does not actually contain the tests.

PR #130, opened 2018-02-01
swupd: Add file index for swupd search

matthewrsj:
Add a file->bundle index file to go in a manufactured
os-core-update-index bundle for faster swupd search operations on
client. Instead of downloading all manifests in the current version the
client can just bundle-add this bundle and use the file as its search
helper.

Because this file is part of a bundle and regular update it will be a
candidate for renames and delta updating. Only present files and
symlinks, not directories or deleted files, are added to this file. The
file contains a tab-separated list of filename to bundlename mappings,
sorted first by filename and secondarily by bundle name.

This manifest has to be processed after all other manifests have been
processed because the "subtracted" file lists have to be complete. This
means we have to post-munge the MoM and full manifests with the new
index file and new index bundle.

An exception in bundle includes reading is also necessary because the
first time includes are read this bundle will not exist. This means that
bundles that include this index bundle will not perform manifest
subtraction with that bundle. In this specific case it is okay because
the only file in the manifest lives at a path that is present in every
bundle (/usr/share/clear/...). If this changes in the future the impact
will be small because this bundle will only ever provide one file.

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -268,8 +268,15 @@ func CreateManifests(version uint32, minVersion uint32, format uint, statedir st
 		},
 	}
 
+	var newFull *Manifest
 	// write manifests then add them to the MoM
 	for _, bMan := range newManifests {
+		// handle the full manifest last
+		if bMan.Name == "full" {
+			newFull = bMan
+			continue
+		}
+
 		// sort by version then by filename, previously to this sort these bundles
 		// were sorted by file name only to make processing easier
 		bMan.sortFilesVersionName()
@@ -278,17 +285,8 @@ func CreateManifests(version uint32, minVersion uint32, format uint, statedir st
 			return nil, err
 		}
 
-		// don't need to add full to the MoM
-		if bMan.Name == "full" {
-			continue
-		}
-
-		var fi os.FileInfo
-		if fi, err = os.Lstat(manPath); err != nil {
-			return nil, err
-		}
-
-		if err = newMoM.createManifestRecord(verOutput, manPath, fi, version); err != nil {
+		// add bundle to Manifest.MoM
+		if err = newMoM.createManifestRecord(verOutput, manPath, version); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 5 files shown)


cmarcelo:
Clever idea to have a separate bundle with that information. I'll test and take a look at this tomorrow.

One random thought for the future: we could have another (optional) file like the one you are adding for the bundle -> packages/versions.

matthewrsj:
@cmarcelo:
> One random thought for the future: we could have another (optional) file like the one you are adding for the bundle -> packages/versions.

Yes I suppose, but we would need to persist the package information from the chroot building phase.

swupd/create_manifests.go
```diff
@@ -295,6 +293,35 @@ func CreateManifests(version uint32, minVersion uint32, format uint, statedir st
 		}
 	}
 
+	allManifests := newManifests
+	// add all manifests in MoM, including old ones, to the list to index for swupd search
+	for _, m := range newMoM.Files {
+		if m.Version < version {
```
tmarcu: Your comment above doesn't make sense, this only adds old manifests to the list...?

swupd/manifest.go
```diff
@@ -622,3 +636,176 @@ func maximizeFull(mf *Manifest, bundles []*Manifest) {
...
+}
+
+type bundleIndex struct {
+	fname string
+	bname string
+}
+
+func constructIndex(c *config, ui *UpdateInfo, f2b []*bundleIndex) error {
+	// sort by filename then by bundle
+	sort.Slice(f2b[:], func(i, j int) bool {
```
tmarcu: Nice.

swupd/create_manifests.go
matthewrsj: `allManifests := newManifests` above, then I append all old ones. I'll clarify the comment.

matthewrsj: Done

matthewrsj:
Clarified a comment and decided to sort by bundle size instead of bundle name (after file name), so that smaller bundles can more easily show up first in search.

PR #131, opened 2018-02-01
swupd: add \n to warning printing

cmarcelo:

### DIFF
--- a/swupd/filesystem.go
+++ b/swupd/filesystem.go
@@ -91,7 +91,7 @@ func (m *Manifest) createManifestRecord(rootPath string, path string, fi os.File
 		if strings.Contains(err.Error(), "hash calculation error") {
 			return err
 		}
-		fmt.Fprint(os.Stderr, err)
+		fmt.Fprintf(os.Stderr, "Warning: %s\n", err)
 	}
 
 	// this is a file to skip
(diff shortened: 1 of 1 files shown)


cmarcelo:
Haven't looked whether these should be printing or not, just ensuring they don't get in the way in testing output.

matthewrsj:
This is poor error handling to begin with. I'd rather see a solution that fixes the `strings.Contains()` nastiness as well, which will probably make this `\n` fix obsolete.

My fault, I wrote this error handling. I remember you and @icarus-sparry had advice to fix it but I haven't gotten around to it.

cmarcelo:
Should non hash calculation errors be ignored? Do you want a list of warnings in the end?

matthewrsj:
Yes. The other errors that can occur on the file level are blacklisted filenames and unsupported file types. In those cases we should report them, but not fail the build. The blacklisted filenames happen (always? almost always?) often.

cmarcelo:
> In those cases we should report them, but not fail the build

At the moment CreateManifests return a MoM and an error. Should it also return some kind of list of warnings (like we did for packs and fullfiles)?

matthewrsj:
@cmarcelo yes I think that's a good idea.

tmarcu:
Bump?

cmarcelo:
I'm happy merging this (to avoid messing with go test output), and deal with improvements in a different PR.

PR #132, opened 2018-02-01
swupd: Add deltas to the packs

cmarcelo:
Make pack creation call delta generation, so the delta packs have
deltas inside instead of only fullfiles.

To achieve this, the delta generation was modified in following ways:

- Return a list of all the deltas, not only the failed ones, this is
  useful when packing. Otherwise a combination of peeking at DeltaPeer
  and the failed list was needed.

- Simplify the goroutine logic: since we have a "home" for all the
  error values (in Delta struct), we can just feed all of the deltas
  to be processed and wait for the goroutines to finish.

- Instead of calling linkPeerAndChange, call a variant that does not
  change the versions, that allows the manifest to be used later on
  for the rest of packing.

It is still a TODO to consolidate those two linking functions, or
decide it is not worth.

### DIFF
--- a/swupd/delta.go
+++ b/swupd/delta.go
@@ -3,17 +3,40 @@ package swupd
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sync"
+	"syscall"
 
 	"github.com/clearlinux/mixer-tools/helpers"
 	"github.com/pkg/errors"
 )
 
-// CreateDeltas creates all delta files between the previous and current
-// version of the supplied manifest, returning a list of files that
-// delta creation failed for.
-func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
+const (
+	// From swupd-server's include/swupd.h:
+	//
+	//     Approximately the smallest size of a pair of input files which differ by a
+	//     single bit that bsdiff can produce a more compact deltafile. Files smaller
+	//     than this are always marked as different. See the magic 200 value in the
+	//     bsdiff/src/diff.c code.
+	//
+	minimumSizeToMakeDeltaInBytes = 200
+)
+
+// Delta represents a delta file between two other files. If Error is present, it
+// indicates that the delta couldn't be created.
+type Delta struct {
+	Path  string
+	Error error
+	from  *File
+	to    *File
+}
+
+// CreateDeltas creates all delta files between the previous and current version of the
+// supplied manifest. Returns a list of deltas (which contains information about
+// individual delta errors). Returns error (and no deltas) if it can't assemble the delta
+// list.
+func CreateDeltas(manifest, statedir string, from, to uint32) ([]Delta, error) {
 	var c config
 
 	c, err := getConfig(statedir)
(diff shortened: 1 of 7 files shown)


swupd/manifest.go
```diff
@@ -467,6 +468,74 @@ func (m *Manifest) newDeleted(df *File) {
...
+
+			if nf.Hash == of.Hash {
+				continue
+			}
+
+			if !nf.Present() || !nf.Present() {
+				continue
+			}
+
+			if nf.Type != of.Type || nf.Type != TypeFile {
```
matthewrsj: Optional: `nf.Type != TypeFile` is much more likely to be `true` than `nf.Type != of.Type` since type changes are rare. If you move the second conditional before the first you will be doing much less comparisons for the full update.

cmarcelo: Done.

cmarcelo:
Pushed a new version right now with the following major changes:

- Use a different error message when we get FULLDL return code from bsdiff. This helps when trying to understand why a delta was not considered.
- Remove TODO for consolidating link peers functions. Rename the new function to linkDeltaPeersForPack to make its role clear.
- Change the TODO inside that function to describe how we currently think the logic for handling renames in the new manifest should be.

Added another patch that removes the "if its in from, don't ship it" logic, and the corresponding test.

swupd/packs.go
```diff
@@ -142,15 +169,43 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
...
+		d := &deltas[i]
+		if d.Error != nil {
+			info.Warnings = append(info.Warnings, d.Error.Error())
+			continue
+		}
+		var fallback bool
+		fallback, err = copyFromDelta(tw, d)
+		if err != nil {
+			// If copy from delta fails before writing to the pack, we can
+			// fallback to use the fullfile later.
```
matthewrsj: Thank you for adding this comment

cmarcelo:
Just pushed a new version without an outdated text in the commit message of the first patch. No real change.

swupd/delta.go
```diff
@@ -31,149 +54,122 @@ func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
...
 	if err != nil {
-		return nil, errors.Wrapf(err, "Failed to create deltas list %s", manifest)
+		return nil, errors.Wrapf(err, "Failed to create deltas list %s", newManifest.Name)
+	}
+
+	if len(deltas) == 0 {
+		return []Delta{}, nil
 	}
 
+	// TODO: Unify these maxRoutines constants, get a better default (maybe NumCPUs)
```
tmarcu: In the C code we have typically used 3 or NumCPUs*3, and my (very minimal) testing showed that it was a consistently good limit. I'd be interested in your perf analysis later on to see if this still holds true here.

swupd/delta.go
```diff
@@ -31,149 +54,122 @@ func CreateDeltas(manifest, statedir string, from, to uint32) ([]*File, error) {
...
 			defer wg.Done()
-
-			// Take in jobs from the queue and try to make a delta
-			for f := range deltaQueue {
-				err := createDelta(c, f, statedir, f.DeltaPeer.Version, f.DeltaPeer.Hash)
-				if err != nil {
-					failedChan <- f
-				}
+			for delta := range deltaQueue {
+				delta.Error = createDelta(c, delta)
```
tmarcu: +1, cool alternative.

swupd/delta.go
cmarcelo: Just making the user has slightly more control of this (even if our default remains 3) will already be useful.

PR #135, opened 2018-02-02
builder: Parallelize chroot creation

tmarcu:
This is very IO bound and will barely hit the CPUs even if we do not limit the amount of routines starting up.

### DIFF
--- a/builder/chroots.go
+++ b/builder/chroots.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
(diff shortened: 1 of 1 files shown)


cmarcelo:
I'd like to test that the competition on I/O will not cause trouble. My suspicion is if we spawn N different processes doing I/O for different places, contention on our shared resource (disk) might lead to worse results.

tmarcu:
This was pushed too early, it needs to return the err from the first routine that fails heh.

PR #134, opened 2018-02-02
Rename tests

matthewrsj:
These tests need to pass for renames. This is based on #128. The only relevant commit is the most recent one.

PR #136, opened 2018-02-02
swupd: improve test infrastructure and import tests from swupd-server

cmarcelo:
Add testSwupd, in the same spirit of testFileSystem, to make tests less noisy.

Convert two tests to use it, to give some context for reviewing it.

Also import new tests from swupd-server.

Related to #138.

### DIFF
--- a/swupd/create_manifests_test.go
+++ b/swupd/create_manifests_test.go
@@ -229,68 +229,6 @@ func TestCreateManifestDeletes(t *testing.T) {
 	checkManifestContains(t, testDir, "20", "test-bundle", deletedLine)
 }
 
-func TestCreateManifestIncludeVersionBump(t *testing.T) {
-	testDir := mustSetupTestDir(t, "includeverbump")
-	defer removeIfNoErrors(t, testDir)
-	bundles := []string{"test-bundle", "included", "included2", "included-nested"}
-	mustInitStandardTest(t, testDir, "0", "10", bundles)
-	mustGenFile(t, testDir, "10", "test-bundle", "foo", "foo")
-	// no includes file for first update
-	mustCreateManifestsStandard(t, 10, testDir)
-
-	expected := []string{"includes:\tos-core\n", "10\t/foo\n"}
-	checkManifestContains(t, testDir, "10", "test-bundle", expected...)
-
-	mustInitStandardTest(t, testDir, "10", "20", bundles)
-	// same file as last update, now manifest creation must be triggered by new includes
-	mustGenFile(t, testDir, "20", "test-bundle", "foo", "foo")
-	mustGenFile(t, testDir, "20", "included", "bar", "bar")
-	mustGenFile(t, testDir, "20", "included2", "baz", "baz")
-	mustInitIncludesFile(t, testDir, "20", "test-bundle", []string{"included", "included2"})
-	mustCreateManifestsStandard(t, 20, testDir)
-
-	cases := []struct {
-		exp   []string
-		bname string
-	}{
-		{[]string{
-			"includes:\tos-core\n",
-			"includes:\tincluded\n",
-			"includes:\tincluded2\n",
-		},
-			"test-bundle"},
-		{[]string{"20\t/bar\n"}, "included"},
-		{[]string{"20\t/baz\n"}, "included2"},
-	}
-	for _, tc := range cases {
-		checkManifestContains(t, testDir, "20", tc.bname, tc.exp...)
-	}
-
-	mustInitStandardTest(t, testDir, "20", "30", bundles)
-	// again, same files
-	mustGenFile(t, testDir, "30", "test-bundle", "foo", "foo")
-	mustGenFile(t, testDir, "30", "included", "bar", "bar")
-	mustGenFile(t, testDir, "30", "included2", "baz", "baz")
-	mustGenFile(t, testDir, "30", "included-nested", "foobarbaz", "foobarbaz")
-	mustInitIncludesFile(t, testDir, "30", "test-bundle", []string{"included", "included2"})
-	mustInitIncludesFile(t, testDir, "30", "included", []string{"included-nested"})
-	mustCreateManifestsStandard(t, 30, testDir)
-
-	cases = []struct {
-		exp   []string
-		bname string
-	}{
-		{[]string{"includes:\tincluded-nested\n", "20\t/bar\n"}, "included"},
-		{[]string{"30\t/foobarbaz\n"}, "included-nested"},
-	}
-	for _, tc := range cases {
-		checkManifestContains(t, testDir, "30", tc.bname, tc.exp...)
-	}
-
-	mustNotExist(t, filepath.Join(testDir, "www/30/Manifest.test-bundle"))
-	mustNotExist(t, filepath.Join(testDir, "www/30/Manifest.included2"))
-}
-
 func TestCreateManifestsState(t *testing.T) {
 	testDir := mustSetupTestDir(t, "state")
 	defer removeIfNoErrors(t, testDir)
--- a/swupd/create_manifests_test.go
+++ b/swupd/create_manifests_test.go
@@ -144,15 +144,28 @@ func TestCreateManifestDebuginfo(t *testing.T) {
 }
 
 func TestCreateManifestFormatNoDecrement(t *testing.T) {
-	testDir := mustSetupTestDir(t, "format-no-decrement")
-	defer removeIfNoErrors(t, testDir)
-	mustInitStandardTest(t, testDir, "0", "10", []string{})
-	mustGenFile(t, testDir, "10", "os-core", "baz", "bazcontent")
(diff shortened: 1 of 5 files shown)


cmarcelo:
Imported another test.

swupd/helpers_test.go
```diff
@@ -361,9 +360,165 @@ func (fs *testFileSystem) cp(src, dst string) {
...
+
+	Bundles    []string
+	MinVersion uint32
+	Format     uint
+}
+
+func newTestSwupd(t *testing.T, prefix string) *testSwupd {
+	fs := newTestFileSystem(t, prefix)
+	defer func() {
+		if t.Failed() {
```
matthewrsj: Do you mean `if !t.Failed()`? This looks like it is cleaning up only if the test failed.

matthewrsj:
commit 1: one comment
commit 2: +1

matthewrsj:
commit 3: did you see https://github.com/clearlinux/mixer-tools/blob/master/swupd/create_manifests_test.go#L231 ?

swupd/packs_test.go
```diff
@@ -377,6 +378,29 @@ func mustValidateZeroPack(t *testing.T, manifestPath, packPath string) {
...
+		if hdr.Typeflag != tar.TypeDir {
+			t.Fatalf("invalid pack: %s is of type %c instead of %c (directory)", name, hdr.Typeflag, tar.TypeDir)
+		}
+		var expectedMode int64 = 0700
+		if hdr.Mode != expectedMode {
+			t.Fatalf("invalid pack: wrong permissions %s for %s, expected %s", os.FileMode(hdr.Mode), name, os.FileMode(expectedMode))
+		}
+	}
+
+	mustHaveDir("delta/")
```
matthewrsj: These rely on order right? If I'm right, can you put a comment in?

swupd/packs_test.go
```diff
@@ -385,8 +409,12 @@ func mustValidateZeroPack(t *testing.T, manifestPath, packPath string) {
 		if err != nil {
 			t.Fatalf("error reading pack: %s", err)
 		}
-		if hdr.Name == "staged/" || !strings.HasPrefix(hdr.Name, "staged/") {
-			continue
+		if hdr.Name == "staged/" || hdr.Name == "delta/" {
```
matthewrsj: Another good place to put the comment that you've already encountered these.

swupd/packs_test.go
```diff
@@ -385,8 +409,12 @@ func mustValidateZeroPack(t *testing.T, manifestPath, packPath string) {
...
+			t.Fatalf("multiple entries of %s directory", hdr.Name)
+		}
+
+		if !strings.HasPrefix(hdr.Name, "staged/") {
```
matthewrsj: And a comment that delta/ shouldn't have anything in it for a zero pack?

matthewrsj:
commit 4: great full run test, couple comment requests.

matthewrsj:
commit 5: that test wasn't really testing much of anything before. Oversight. Good update. +1

swupd/create_manifests_test.go
```diff
@@ -567,3 +567,61 @@ func mustParseAllManifests(t *testing.T, version uint32, outputDir string) map[s
...
 	return result
 }
+
+// Imported from swupd-server/test/functional/subtract-delete.
+func TestSubtractDelete(t *testing.T) {
+	ts := newTestSwupd(t, "subtract-delete-")
+	defer ts.cleanup()
+
+	ts.Bundles = []string{"os-core", "test-bundle"}
+	ts.Format = 3
```
matthewrsj: You don't seem to change this value again. Why 3? I recall it was 3 in the old server because Clear Linux was on format 3 when the tests were originally written, but I don't think it is relevant here.

swupd/helpers_test.go
cmarcelo: The code is correct, but due to similarity how defer cleanup is used elsewhere might be confusing. If when creating a testSwupd we have a failure, cleanup any existing directory. Otherwise let it live because the caller will call cleanup themselves.

I'll add a comment there.

cmarcelo: Note that currently fs.cleanup won't really cleanup (because it was a failure) but will still show the helpful message with the path of the temporary directory.

cmarcelo:
> commit 3: did you see https://github.com/clearlinux/mixer-tools/blob/master/swupd/create_manifests_test.go#L231 ?

I've missed that one, otherwise I'd written the extra tests in there, sorry about that.

@matthewrsj  If you like the direction test helpers in this PR took, I'll amend the commit to remove the existing version of the test. Otherwise I'll just remove this commit and later post a PR adding the extra checks to that test.

swupd/create_manifests_test.go
cmarcelo: Yes, I was just keeping the test doing as much as possible the same as before -- I guessed it was an historical reason or something. Having some tests using different formats is a win, but I think we have those around already.

I'll remove format 3 from tests that don't really depend on that.

matthewrsj:
>  If you like the direction test helpers in this PR took, I'll amend the commit to remove the existing version of the test. Otherwise I'll just remove this commit and later post a PR adding the extra checks to that test.

Can you actually just replace the old test as part of this commit or do you think that is out-of-scope? The new test uses the new helpers, which I like.

matthewrsj:
/me realizes that's what you suggested doing @cmarcelo, +1 please do.

cmarcelo:
> commit 6: please remove the if false { blockers. Your assumptions are correct and these tests should be failing (#134).

Problem of having tests failing is make the test suite less useful, I think the TODOs and skipping the test is fine in this case, since this PR is not pushing the fix for the issue at the same time. When the fix land, it can get rid of TODOs and skips.

cmarcelo:
> Your assumptions are correct and these tests should be failing (#134).

I'll update the TODOs to reflect that knowledge.

matthewrsj:
> Problem of having tests failing is make the test suite less useful, I think the TODOs and skipping the test is fine in this case, since this PR is not pushing the fix for the issue at the same time. When the fix land, it can get rid of TODOs and skips.

I disagree. The whole point of tests is so that they fail. It is easy to forget about TODOs and skips once something is merged. If that means we merge failing tests now at least we know what is failing and what needs to be fixed. I see no usefulness in tests that **should** be failing but are passing instead.

cmarcelo:
> I disagree. The whole point of tests is so that they fail. It is easy to forget about TODOs and skips once something is merged. If that means we merge failing tests now at least we know what is failing and what needs to be fixed. I see no usefulness in tests that should be failing but are passing instead.

If you merge a failing test, the CI output becomes less useful, because every PR and every test execution from now on will be marked as fail. Unless you verify each time: "hey, it's failing because the known failures A, B, C, so its fine", you might let some other failure to creep in. That manual verification is the opposite of what we want to achieve with the CI marking the PR as "checked".

Are those skipped tests useful? They are a better version of the "// TODO: Hey, we should be persisting deletes here.", because they come with the exact code that is expected to be working.

matthewrsj:
> If you merge a failing test, the CI output becomes less useful, because every PR and every test execution from now on will be marked as fail. Unless you verify each time: "hey, it's failing because the known failures A, B, C, so its fine", you might let some other failure to creep in. That manual verification is the opposite of what we want to achieve with the CI marking the PR as "checked".

So verify every time. I think for minor tests what you say makes sense. This is a breaking issue though. The updates **do not** work over several versions until it is fixed. It should be advertised that the bug exists. Painful, but better than relying on memory that the TODO exists.

> Are those skipped tests useful? They are a better version of the "// TODO: Hey, we should be persisting deletes here.", because they come with the exact code that is expected to be working.

Sure, but I have an actual test in #134 to test for this

This is all philosophical. I have a fix for the bug that I'll be pushing to a new PR in the next few minutes.

cmarcelo:
> This is all philosophical. I have a fix for the bug that I'll be pushing to a new PR in the next few minutes.

Great! So we have a non-problem :-)

cmarcelo:
Updated PR addressing comments.

swupd/helpers_test.go
matthewrsj: Nice, thanks.

cmarcelo:
> Would have liked a couple more comments around how the zero pack scanning is done (expects alphabetical order, delta directory should not have any children), but that was optional, so +1.

Argh. Read but missed those when doing the second pass.

swupd/packs_test.go
cmarcelo: Done.

swupd/packs_test.go
cmarcelo: Done. Also mentioned I think we can relax this in the future. For now I think the validation function is fine as is.

swupd/packs_test.go
cmarcelo: The error message talks about "multiple %s found", so I think we are covered with that plus the other two comments added.

matthewrsj:
+1 again, thanks for adding the comments.

PR #142, opened 2018-02-06
Import no-deltas test from swupd-server

cmarcelo:
Import a test and add a simple one.

Fix a difference in behavior with swupd-server: a target might have both a fullfile and a delta to it. Added a TODO note that we may want to drop packing the delta in this case.

Improve information about packs -- now we report what delta was pack for a specific file.

### DIFF
--- a/swupd/delta_test.go
+++ b/swupd/delta_test.go
@@ -1,7 +1,10 @@
 package swupd
 
 import (
+	"fmt"
+	"io/ioutil"
 	"path/filepath"
+	"strings"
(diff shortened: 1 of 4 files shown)


cmarcelo:
Repushed to fix typo in test noticed by Tudor. Thanks!

tmarcu:
+1 but i'd like @matthewrsj to look at the import no-delta patch as well.

matthewrsj:
Yep, +1

PR #149, opened 2018-02-07
Rename, cleanup, and remove .mixer directory

kevincwells:
>**NOTE**: Per @cmarcelo's request, I pulled this out of #137 so that it can be merged sooner, and so the discussion around deprecation can be had here. 
It is almost exactly the same as the commit in #137, but because it's now on its own (instead of on top of the commit now in #148), there was a small conflict in the section in BuildChroots that #148 also touches. This will likely need to be rebased on top of #148 once it merges.

This commit has three main side effects:
1) The .clearversion, .mixversion, and .clearurl files are no longer
hidden files.

They were likely originally hidden to declutter the workspace, but
they are (for now) files the user needs to edit directly, and thus
should not be hidden.

2) The clearversion and clearurl files has been renamed
upstreamversion and upstreamurl, as part of a larger goal to
accomodate derivatives-of-derivatives, for whom upstream is not
mainline CLR. (The internal fields on the Builder object have been
renamed as well.)

3) Per feedback, the .mixer directory has been removed.

Two of the goals the directory was introduced to solve will instead
be met in different ways:

i) Hiding temporary, sausage-making files. The mix-bundles/
directory will soon cease to exist entirely, once the BCB rewrite
moves out from behind UseNewChrootBuilder. upstream-bundles/ also
contains temporary, tool-generated files, but people have expressed
interest (or at least tolerance) of these not being hidden away.

ii) Organizing/decluttering the workspace. As there are numerous
different configuration files (including the mixversion,
clearversion, clearurl, and forthcoming mix bundle list), having
them in the .mixer directory provided a better form of decluttering
than having the files as hidden dot-files. However, ideally these
files will soon be merged into the builder.conf, when the that
file gets rewritten as a TOML file.

This commit also has minor side effects:
1) General code clean-up (e.g., use of filepath.Join())
2) Change Builder struct field names to match capitalization
conventions
3) Moves the location of InitMix within the file for organizational
purposes.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -49,21 +49,24 @@ var UseNewChrootBuilder = false
 // A Builder contains all configurable fields required to perform a full mix
 // operation, and is used to encapsulate life time data.
 type Builder struct {
-	Buildscript string
-	Buildconf   string
+	BuildScript string
+	BuildConf   string
 
-	Bundledir   string
-	Cert        string
-	Clearver    string
-	Format      string
-	Mixver      string
-	Repodir     string
-	RPMdir      string
-	Statedir    string
-	Versiondir  string
-	Yumconf     string
-	Yumtemplate string
-	Upstreamurl string
+	BundleDir       string
+	Cert            string
+	UpstreamVer     string
+	UpstreamVerFile string
+	Format          string
+	MixVer          string
+	MixVerFile      string
+	RepoDir         string
+	RPMDir          string
+	StateDir        string
+	VersionDir      string
+	YumConf         string
+	YumTemplate     string
+	UpstreamURL     string
+	UpstreamURLFile string
 
 	Signing int
 	Bump    int
@@ -73,8 +76,11 @@ type Builder struct {
 // default values.
 func New() *Builder {
 	return &Builder{
-		Buildscript: "bundle-chroot-builder.py",
-		Yumtemplate: "/usr/share/defaults/mixer/yum.conf.in",
+		BuildScript:     "bundle-chroot-builder.py",
+		YumTemplate:     "/usr/share/defaults/mixer/yum.conf.in",
+		UpstreamVerFile: "upstreamversion",
+		UpstreamURLFile: "upstreamurl",
+		MixVerFile:      "mixversion",
 
 		Signing: 1,
 		Bump:    0,
@@ -121,8 +127,11 @@ func (b *Builder) CreateDefaultConfig(localrpms bool) error {
 		return err
 	}
 
+	// Patch all default path prefixes to PWD
 	data := strings.Replace(string(raw), "/home/clr/mix", pwd, -1)
+
 	if localrpms {
+		// Add [Mixer] section
 		data += "\n[Mixer]\n"
 		data += "RPMDIR=" + pwd + "/rpms\n"
 		data += "REPODIR=" + pwd + "/local\n"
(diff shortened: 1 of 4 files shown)


kevincwells:
The discussion regarding deprecation thus far:
@cmarcelo writes:
>I'm wondering if there's something to be done about previous .clearver, .mixver files in case you find them: e.g. you could still read those as a fallback (if the good ones don't exist) and print a warning saying these filenames are deprecated (also make sure to not create the new files, but update the older one in case exists).

@kevincwells writes:
>I like this idea, as it makes it a deprecation rather than a hard switch. If people want this or think it is necessary, I'd be happy to implement it. But I'm a smidge hesitant to keep people on the old filename if they're using it -- I'd argue the tool should convert the file for the user if it finds the old one.

kevincwells:
@tmarcu what are your thoughts?

cmarcelo:
I'd also would do the CLI changes in a similar fashion (actually I thought we would be doing that), so the previous commands would still work. And then there's no hard synchronization needed for the update.

cmarcelo:
> Alternative approach: is to make the argument of doing all those changes synchronized.

You'll need @tmarcu blessing to take that route.

kevincwells:
>>Alternative approach: is to make the argument of doing all those changes synchronized.

>You'll need @tmarcu blessing to take that route.

That's currently the route being taken with the CLI. It was merged into master, but there wasn't going to be a release that put it into CLR until folks had a chance to switch over and test with the new version.

cmarcelo:
swupd-client also would have to be updated.

https://github.com/clearlinux/swupd-client/search?utf8=%E2%9C%93&q=.clearversion&type=
https://github.com/clearlinux/swupd-client/search?utf8=%E2%9C%93&q=.mixversion%7C.clearversion&type=

kevincwells:
>swupd-client also would have to be updated.

Would it actually? Those are both on the client. I agree it would be good to generally keep the two in sync in terms of naming conventions and terminology, but I wouldn't think either _depends_ on the other changing.

cmarcelo:
Client calls mixer, see also: https://github.com/clearlinux/swupd-client/blob/master/swupd-add-pkg

cmarcelo:
This is used for the feature of having an implicit local mix.

kevincwells:
That it does.  Which  means it needs to be updated for the new CLI either way. Great catch.

kevincwells:
PS - I'm willing to scrap the name-change for now (externally -- internally, I think it's still good to rename things) if this is changing too much at once, but I figured a) IMHO it needs to happen sometime, b) why not lump it into a time when other things need to change?

kevincwells:
PPS - I'm adding in deprecation code now :)

tmarcu:
@cmarcelo  @kevincwells  Ehhhh... ymmv. A couple files are easy to mitigate and have backward compatibility for, so this is fine. We can make that breaking change later or just make it automatically rename their version files until the following format bump, they're auto-updated usually anyway. Supporting backwards compatibility for the entire CLI is a nightmare code-wise. An option is to make the new CLI its own package, so people call mixer-cli for new flags, and mixer for the old. But then people have to port their scripts/calls twice, or they just won't do it (most likely case). Providing a transition script for getting onto the new - **standardized** - CLI is going to make the switch easy and not make us carry ugly code for N builds.
That being said, just cause we're putting a breaking change in like that doesn't mean we should keep piling them on ;)

kevincwells:
Just pushed a rebase to account for #148 merging.

PR #151, opened 2018-02-07
swupd: pack more deltas based on pre-calculated rename flag

cmarcelo:
When "unpacking" swupd client look for deltas based on the rename
flag, so use the same logic to pack deltas for it. With these the new
implementation of packs should be on par with the swupd-server.

### DIFF
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -604,8 +604,72 @@ func linkDeltaPeersForPack(c *config, oldManifest, newManifest *Manifest) error 
 		}
 	}
 
-	// TODO: Once rename logic is in. Add code to identify what extra DeltaPeers
-	// should be created based on rename flag in newManifest.
+	// TODO: At the moment swupd-client looks at the new manifest renames entries to make a
+	// decision whether it should get a delta or not. If we change that to look instead at the
+	// list of deltas in the pack, we could run the full rename detection logic between
+	// arbitrary pairs of manifests, i.e. for any pack combination, and possibly generating new
+	// deltas for that specific combination.
+
+	// Identify deltas that are relevant to be included based on the rename flag.
+	//
+	// First walk to find potential targets for a rename. We will generate only a single delta
+	// for each target.
+	targets := make(map[Hashval]*File)
+
+	for _, nf := range newManifest.Files {
+		if !bool(nf.Rename) || !nf.Present() {
+			// Not a rename target.
+			continue
+		}
+		if nf.DeltaPeer != nil {
+			// If there is a DeltaPeer matched by same name (not a rename),
+			// just keep that and don't look for another.
+			continue
+		}
+		_, ok := targets[nf.Hash]
+		if ok {
+			continue
+		}
+		targets[nf.Hash] = nf
+	}
+
+	// Then walk again to link the sources to the targets.
+	renames := make(map[string]*File)
+	for _, nf := range newManifest.Files {
+		if !bool(nf.Rename) || nf.Present() {
+			// Not a rename source.
+			continue
+		}
+		target, ok := targets[nf.Hash]
+		if !ok {
+			// TODO: This could be a warning to the caller. Not making it an error since
+			// the manifest is already written, and we can underdeliver deltas. Same
+			// applies for targets without suitable sources.
+			continue
+		}
+		renames[nf.Name] = target
+	}
+
+	// Finally walk the old manifest linking rename files that the source exist there. Note that
+	// we ignore the version when the rename was identified, because the client will look for
+	// this regardless that version.
+	for _, of := range oldManifest.Files {
+		if !of.Present() {
+			continue
+		}
+		if of.DeltaPeer != nil {
+			continue
+		}
+		nf, ok := renames[of.Name]
+		if !ok {
+			continue
+		}
+		if nf.DeltaPeer != nil {
+			continue
+		}
+		nf.DeltaPeer = of
+		of.DeltaPeer = nf
+	}
 
 	return nil
 }
--- a/swupd/packs_test.go
+++ b/swupd/packs_test.go
@@ -537,3 +537,115 @@ func TestTwoDeltasForTheSameTarget(t *testing.T) {
 		ts.checkExists(fmt.Sprintf("www/20/delta/10-20-%s-%s", hashA, hashB))
 	}
 }
+
+func TestPackRenames(t *testing.T) {
+	ts := newTestSwupd(t, "packing-renames-")
+	defer ts.cleanup()
+
+	ts.Bundles = []string{"os-core"}
+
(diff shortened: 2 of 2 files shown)


swupd/rename.go
```diff
@@ -93,6 +93,7 @@ func linkRenamePair(renameTo, renameFrom *File) {
 	renameTo.Rename = true
 	renameFrom.Rename = true
 	renameFrom.Type = TypeUnset
+	renameFrom.Hash = renameTo.Hash
```
cmarcelo: This should be gone when we rebase with the fix/test by Matthew.

matthewrsj: done and in master

swupd/manifest.go
```diff
@@ -541,8 +541,71 @@ func linkDeltaPeersForPack(c *config, oldManifest, newManifest *Manifest) error
...
+	// arbitrary pairs of manifests, i.e. for any pack combination, and possibly generating new
+	// deltas for that specific combination.
+
+	// Identify deltas that are relevant to be included based on the rename flag.
+	//
+	// First walk to find potential targets for a rename. We will generate only a single delta
+	// for each target.
+	targets := make(map[Hashval]*File)
+
+	for _, nf := range newManifest.Files {
```
matthewrsj: This is a complicated function and does a lot of walking of the manifest. With #154 a function is exposed to link old renames (`linkOldRenames`) that doesn't cost O(n<sup>2</sup>) to compute. 

In fact, now that `linkPeersAndChange` does all the linking even for old renames, does it make sense to going back to using that function?

swupd/packs_test.go
```diff
@@ -537,3 +537,106 @@ func TestTwoDeltasForTheSameTarget(t *testing.T) {
...
 	}
 }
+
+func TestPackRenames(t *testing.T) {
+	ts := newTestSwupd(t, "packing-renames-")
+	defer ts.cleanup()
+
+	ts.Bundles = []string{"os-core"}
+
+	content := strings.Repeat("CONTENT", 1000)
```
matthewrsj: much larger than it needs to be to catch a rename/delta.

swupd/packs_test.go
```diff
@@ -537,3 +537,106 @@ func TestTwoDeltasForTheSameTarget(t *testing.T) {
...
+	info = ts.createPack("os-core", 20, 30, ts.path("image"))
+	mustHaveDeltaCount(t, info, 1)
+	checkFileInPack(t, ts.path("www/30/pack-os-core-from-20.tar"), fmt.Sprintf("delta/20-30-%s-%s", hashIn20, hashIn30))
+
+	// Pack from 10->30 will contain a delta due to content change (and rename).
+	info = ts.createPack("os-core", 10, 30, ts.path("image"))
+	mustHaveDeltaCount(t, info, 1)
+	checkFileInPack(t, ts.path("www/30/pack-os-core-from-10.tar"), fmt.Sprintf("delta/10-30-%s-%s", hashIn10, hashIn30))
+
+	// TODO: When/If rename sticks flag, we can add a Version 40, that change an
```
matthewrsj: *sigh* you know my thoughts on this. Have a fix in #154 though.

cmarcelo: I'll rebase when it lands. The "/If" part was more because at the time I asked there was no consensus about this.

swupd/manifest.go
cmarcelo: I might be missing something, but these additions did not cause the function to be O(N^2).

linkOldRenames sets up DeltaPeers between files of the same manifest, we are doing something different here, we are linking with files from the old manifest, because their version might differ from the version mentioned in new manifest (in rename). That scenario is when a "new delta" may happen.

I still prefer handling things separatedly from linkPeersAndChange, that function still mutates files in different conditions we don't want. I rather we get things working, and then try to settle some duplication.

swupd/manifest.go
```diff
@@ -541,8 +541,71 @@ func linkDeltaPeersForPack(c *config, oldManifest, newManifest *Manifest) error
...
+		fmt.Println(nf.Name, nf.Rename)
```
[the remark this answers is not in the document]
cmarcelo: It was an intense day, will fix this :-)

swupd/manifest.go
matthewrsj: ok, right. Forgot this was between any two manifests, not previous/current.

swupd/manifest.go
```diff
@@ -541,8 +541,71 @@ func linkDeltaPeersForPack(c *config, oldManifest, newManifest *Manifest) error
...
+		if !bool(nf.Rename) || !nf.Present() {
+			// Not a rename target.
+			continue
+		}
+		if nf.DeltaPeer != nil {
```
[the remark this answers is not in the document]
cmarcelo: When we start, all the DeltaPeers are empty.

Then we do the first pass matching things by name. The only DeltaPeers are between old and new files that have the same name but different hashes. (This is the code that was already in the function).

Then we will figure (using rename information) if there are another pairs of old and new that must be created. First we find all the targets of renames, then we find the matching source (so we can have the filename). Then we iterate on the old manifest to find if there are files with those names, so we link them to the target of the rename.

If a target already has a deltapeer, we have two options: keep it that way (pack later will try to create that delta), or unlink it from its previous connection by same name and see if a connection by a rename will happen. I've decided to keep the first.


We are collecting all the rename targets. If one of them already has a delta peer, i.e. if in the first step of linking (where we match names) we found a matching peer, then we are going to give preference to that.

The alternative is to "disconnect

swupd/packs_test.go
matthewrsj: rename persistence now fixed and merged.

swupd/manifest.go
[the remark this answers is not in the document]
matthewrsj: Ah, ok thanks for clarifying.

swupd/manifest.go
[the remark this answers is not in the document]
cmarcelo: Done.

swupd/manifest.go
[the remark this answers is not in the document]
cmarcelo: Added a tweet sized version of this explanation as a comment.

swupd/packs_test.go
cmarcelo: Wrote the test and removed the comment.

(Now I figured that maybe your concern was about having a TODO instead of a failing test. We are lucky this is a non-problem again :-)

cmarcelo:
Rebased.

swupd/packs_test.go
tmarcu: Gets deleted after test runs though, es ok.

cmarcelo:
Pushed again just to add "Fixes ..." to the commit message.

PR #153, opened 2018-02-07
Support (and default to) "latest" in mixer init clear-version

kevincwells:
Support (and default to) "latest" in mixer init clear-version
    
Adds support for the keyword "latest" for `--clear-version` when
running mixer init. This becomes the default value if the flag is
omitted.
    
'--clear-version latest' will fetch the latest published upstream
version number, according to the value found via 'upstreamurl'
    
### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -23,6 +23,8 @@ import (
 	"io"
 	"io/ioutil"
 	"math"
+	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


builder/builder.go
```diff
@@ -161,6 +162,23 @@ func (b *Builder) createRpmDirs() error {
 	return nil
 }
 
+// Get latest CLR version
+func (b *Builder) getLatestUpstreamVersion() (string, error) {
+	resp, err := http.Get(b.UpstreamURL + "/latest")
```
matthewrsj: This is basically just a path. I would prefer to use `path.Join` here instead of `"somethingThatMightEndIn/" + "/..."`

kevincwells: I had the same thought, but read people saying that path.Join can mangle URLs, and they wanted me to import the `net/url` package and use a far more complicated `ResolveReference` method.  I figured I'd avoid the import and complication for now.

Think I should change it?

kevincwells: I'll give it a go. It's "the right thing to do".

cmarcelo: My experience was the same as reported by Kevin two comments above.

kevincwells: That was easy enough.

mixer/cmd/root.go
```diff
@@ -99,7 +99,15 @@ var initCmd = &cobra.Command{
 	Use:   "init",
 	Short: "Initialize the mixer and workspace",
 	Long:  `Initialize the mixer and workspace`,
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if _, err := strconv.Atoi(initFlags.clearver); err != nil {
+			if initFlags.clearver != "latest" {
+				// Note: output matches Cobra's default pflag error syntax, as
+				// if initFlags.clearver were an int all along.
+				return errors.Errorf("invalid argument \"%s\" for \"--clear-version\" flag: %s", initFlags.clearver, err)
```
cmarcelo: Please mention that "latest" is a valid argument.

cmarcelo:
Note I've approved the commit in question, not the other one :-)

PR #154, opened 2018-02-07
Deprecate and persist renames properly

matthewrsj:
Renames must persist so make sure the rename flag from the old manifest
sticks around and is copied into the new manifest.

Link these old renames together before linking new renames. After all
the links are set up deprecate any renames that have lost their link.
These are orphaned renames caused either by rename chains or by deleting
renameTo files.

Add a test to check that renames persist.

### DIFF
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -386,13 +386,13 @@ func (m *Manifest) linkPeersAndChange(oldManifest *Manifest, c config, minVersio
 	for nx < mFilesLen && ox < omFilesLen {
 		nf := m.Files[nx]
 		of := oldManifest.Files[ox]
-		of.Rename = false
 		if nf.Name == of.Name {
 			// if it is the same name check if anything about the file has changed.
 			// if something has changed update the version to current version and
 			// record that the file has changed
 			if of.Status == StatusDeleted || of.Status == StatusGhosted {
 				nf.Version = m.Header.Version
+				nf.Rename = of.Rename
 				added = append(added, nf)
 				ox++
 				nx++
@@ -400,6 +400,8 @@ func (m *Manifest) linkPeersAndChange(oldManifest *Manifest, c config, minVersio
 			}
 			if sameFile(nf, of) && of.Version >= minVersion {
 				// file did not change, set version to the old version
+				// persist the Rename flag
+				nf.Rename = of.Rename
 				nf.Version = of.Version
 			} else {
 				// if the file isn't exactly the same, record the change
@@ -459,7 +461,12 @@ func (m *Manifest) linkPeersAndChange(oldManifest *Manifest, c config, minVersio
 		removed = append(removed, of)
 	}
 
+	// link old and new renames in the manifest
+	oldRenames := m.linkOldRenames()
+	// detect new renames in this version
 	renameDetection(m, added, removed, c)
+	// deprecate any old renames here now that all relevant linking has been done
+	deprecateOldRenames(oldRenames)
 
 	// if a file still has deleted status here and is not a rename
 	// the has needs to be set to all zeros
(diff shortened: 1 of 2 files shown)


swupd/manifest.go
```diff
@@ -480,11 +487,52 @@ func (m *Manifest) linkPeersAndChange(oldManifest *Manifest, c config, minVersio
 func (m *Manifest) newDeleted(df *File) {
 	df.Version = m.Header.Version
 	df.Status = StatusDeleted
+	df.Type = TypeUnset
 	df.Modifier = ModifierUnset
 	// Add file to manifest
 	m.Files = append(m.Files, df)
 }
 
+func (f *File) renameFrom() bool {
```
icarus-sparry: I hate these names.

swupd/manifest.go
```diff
@@ -480,11 +487,52 @@ func (m *Manifest) linkPeersAndChange(oldManifest *Manifest, c config, minVersio
...
+}
+
+func (f *File) renameTo() bool {
+	return f.Rename == true && f.Present()
+}
+
+// linkOldRenames links renames present in a manifest already
+// this should be done before renameDetection is run
+func (m *Manifest) linkOldRenames() {
+	for _, f1 := range m.Files {
```
icarus-sparry: This is O(n^2) with large n. Make short lists of just things with an existing rename flag (O(n)) to reduce the amount of work and remove the need for the rename{From,To} functions.

Then sort both by hash, and do the linear pass over the hashes.

Possibly return the (maybe sorted) list so deprecateOldRename can use it to walk a much shorter list.

swupd/manifest.go
matthewrsj: Changed to `isRenameFrom/To`

swupd/manifest.go
```diff
@@ -480,11 +487,52 @@ func (m *Manifest) linkPeersAndChange(oldManifest *Manifest, c config, minVersio
 func (m *Manifest) newDeleted(df *File) {
 	df.Version = m.Header.Version
 	df.Status = StatusDeleted
+	df.Type = TypeUnset
```
icarus-sparry: I know we need this for compatibility, but I think it is wrong. I don't see why a deleted {file,directory.link} is no longer a {file,directory.link} for comparison purposes. It does mean that we should probably be checking for df.Status in more places.

swupd/manifest.go
matthewrsj: yeah ok. I took out a comment at the top of this function that explained that although this was O(n^2), it didn't matter because we would be removing rename flags anyways soon.

It isn't difficult to make this efficient. I'll do so.

icarus-sparry: I usually review the code as posted, not as I think the person intended to write it.

matthewrsj: @icarus-sparry I intended to write it the way I posted it, for the reason I gave above. I was only giving you context.

matthewrsj:
@icarus-sparry updated.

swupd/manifest.go
```diff
@@ -480,11 +487,69 @@ func (m *Manifest) linkPeersAndChange(oldManifest *Manifest, c config, minVersio
...
+
+	// return a list of all old renames for deprecateOldRenames
+	return append(fromRenames, toRenames...)
+}
+
+// deprecateOldRenames unmarks all unlinked renames in oldRenames
+// must be done after linkOldRenames and renameDetection has been done
+func deprecateOldRenames(oldRenames []*File) {
+	for _, f := range oldRenames {
+		if f.Rename == true && f.DeltaPeer == nil {
```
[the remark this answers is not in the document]
matthewrsj: Forgot to update this when changing the function to take a slice

> As I mentioned before, I am not sure if we need to clean up DeltaPeer before we run rename detection.

I just did a quick check through the source and I don't think linkPeersAndChange will ever be called on a manifest that already has peers set.

And a quick test shows that all our tests pass when always setting `nf.DeltaPeer` to `nil` but I don't see why we should do it if we don't need to.

matthewrsj:
@icarus-sparry updated again.

PR #155, opened 2018-02-08
New Command: Mixer bundle edit

kevincwells:
Adds new command, 'mixer bundle edit', that allows a user to edit local
and upstream bundle definition files. This command will locate the
bundle (looking first in local-bundles, then in upstream-bundles), and launch
an editor to edit it. If the bundle is only found upstream, the bundle file will
first be copied to your local-bundles directory for editing. If the bundle is
not found anywhere, a blank template is created with the correct name. When the
editor closes, the bundle file is then parsed for validity.

The editor is configured via environment variables. VISUAL takes precedence to
EDITOR. If neither are set, the tool defaults to nano. If nano is not installed,
the tool will skip editing, and act as if '--suppress-editor' had been passed.

Passing '--suppress-editor' will suppress launching the editor, and will thus
only copy the bundle file to local-bundles (if it is only found upstream), or
create the blank template (if it was not found anywhere). This can be useful if
you want to add a bundle to local-bundles, but wish to edit it at a later time.

Passing '--add' will also add the bundle(s) to your mix. Please note that
bundles are added after all bunles are edited, and thus will not be added if any
errors are encountered earlier on.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -16,6 +16,7 @@ package builder
 
 import (
 	"archive/tar"
+	"bufio"
 	"bytes"
 	"crypto/rsa"
 	"crypto/x509"
@@ -908,6 +909,143 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
 			}
 
 			fmt.Println(val)
+		}
+	}
+
+	return nil
+}
+
+func getEditorCmd() (string, error) {
+	cmd := os.Getenv("VISUAL")
+	if cmd != "" {
+		return cmd, nil
+	}
+
+	cmd = os.Getenv("EDITOR")
+	if cmd != "" {
+		return cmd, nil
+	}
+
+	return exec.LookPath("nano")
+}
+
+// editBundleFile launches an editor command to edit the bundle defined by path.
+// When the edit process ends, the bundle file is parsed for validity. If a
+// parsing error is encountered, the user is asked how to proceed: retry, revert
+// and retry, or skip.
+func editBundleFile(editorCmd string, bundle string, path string) error {
+	// Make backup
+	backup := path + ".orig"
+	if err := helpers.CopyFileNoOverwrite(backup, path); err != nil && !os.IsExist(err) {
+		return errors.Wrapf(err, "Could not backup bundle '%s' file for editing", bundle)
+	}
+
+	reader := bufio.NewReader(os.Stdin)
+	revert := false
+
+editLoop:
+	for {
+		if revert {
+			if err := helpers.CopyFile(path, backup); err != nil {
+				return errors.Wrapf(err, "Could not restore original from backup for bundle '%s'", bundle)
+			}
+		}
+
+		// Ignore return from command; parsing below is what will reveal errors
+		_ = helpers.RunCommandInput(os.Stdin, editorCmd, path)
+
+		_, err := parseBundleFile(path)
+		if err == nil {
+			// Clean-up backup
+			if err = os.Remove(backup); err != nil {
+				return errors.Wrapf(err, "Error cleaning up backup for bundle '%s'", bundle)
+			}
+			break editLoop
+		}
+
+		fmt.Printf("Error parsing bundle %s: %s\n", bundle, err)
+		for {
+			// Ask the user if they want to retry, revert, or skip
+			fmt.Print("Would you like to edit as-is, revert and edit, or skip [Edit/Revert/Skip]?: ")
+			text, err := reader.ReadString('\n')
+			if err != nil {
+				return errors.Wrapf(err, "Error reading input")
+			}
+			text = strings.ToLower(text)
+			text = strings.TrimSpace(text)
+			switch {
+			case text == "e" || text == "edit":
+				revert = false
+				continue editLoop
+			case text == "r" || text == "revert":
+				revert = true
+				continue editLoop
+			case text == "s" || text == "skip":
+				fmt.Printf("Skipping bundle '%s' despite errors. Backup retained as '%s'\n", bundle, bundle+".orig")
+				break editLoop
+			default:
+				fmt.Printf("Invalid input: '%s'", text)
+			}
+		}
+	}
+
+	return nil
+}
+
+// EditBundles copies a list of bundles from upstream-bundles to local-bundles
+// (if they are not already there), and launches an editor to edit them. Passing
+// true for 'copyOnly' will suppress the launching of the editor (and just do
+// the copy, if needed), and 'add' will also add the bundles to the mix.
+func (b *Builder) EditBundles(bundles []string, copyOnly bool, add bool, git bool) error {
+	// Fetch upstream bundle files if needed
+	if err := b.getUpstreamBundles(b.UpstreamVer, true); err != nil {
+		return err
+	}
+
+	editorCmd, err := getEditorCmd()
+	if err != nil {
+		fmt.Println("Cannot find a valid editor (see usage for configuration). Copying to local-bundles only.")
+		copyOnly = true
+	}
+
+	for _, bundle := range bundles {
+		var path string
+		path, err = b.getBundlePath(bundle)
+		if err != nil {
+			return err
+		}
+
+		if !b.isLocalBundle(path) {
+			localPath := filepath.Join(b.LocalBundleDir, bundle)
+			if err = helpers.CopyFile(localPath, path); err != nil {
+				return err
+			}
+			path = localPath
+		}
+
+		if copyOnly {
+			continue
+		}
+
+		if err = editBundleFile(editorCmd, bundle, path); err != nil {
+			return err
+		}
+	}
+
+	if add {
+		if err = b.AddBundles(bundles, false, false, false); err != nil {
+			return err
+		}
+	}
+
+	if git {
+		fmt.Println("Adding git commit")
+		if err := helpers.Git("add", "."); err != nil {
+			return err
+		}
+		commitMsg := fmt.Sprintf("Edited bundles: %v", bundles)
+		if err := helpers.Git("commit", "-q", "-m", commitMsg); err != nil {
+			return err
 		}
 	}
 
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -973,14 +973,14 @@ editLoop:
 			}
 			text = strings.ToLower(text)
 			text = strings.TrimSpace(text)
-			switch {
-			case text == "e" || text == "edit":
+			switch text {
+			case "e", "edit":
 				revert = false
 				continue editLoop
-			case text == "r" || text == "revert":
+			case "r", "revert":
 				revert = true
 				continue editLoop
-			case text == "s" || text == "skip":
+			case "s", "skip":
 				fmt.Printf("Skipping bundle '%s' despite errors. Backup retained as '%s'\n", bundle, bundle+".orig")
 				break editLoop
 			default:
@@ -992,11 +992,33 @@ editLoop:
 	return nil
 }
 
+const bundleTemplateFormat = `# [TITLE]: %s
+# [DESCRIPTION]: 
+# [STATUS]: 
+# [CAPABILITIES]:
+# [MAINTAINER]: 
+# 
+# List bundles one per line. Includes have format: include(bundle)
+`
+
+func createBundleFile(bundle string, path string) error {
+	f, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR|os.O_EXCL, 0666)
+	if err != nil {
+		return err
+	}
+
+	data := []byte(fmt.Sprintf(bundleTemplateFormat, bundle))
+	_, err = f.Write(data)
+	_ = f.Close()
+	return err
+}
+
 // EditBundles copies a list of bundles from upstream-bundles to local-bundles
-// (if they are not already there), and launches an editor to edit them. Passing
-// true for 'copyOnly' will suppress the launching of the editor (and just do
-// the copy, if needed), and 'add' will also add the bundles to the mix.
-func (b *Builder) EditBundles(bundles []string, copyOnly bool, add bool, git bool) error {
+// (if they are not already there) or creates a blank template if they are new,
+// and launches an editor to edit them. Passing true for 'suppressEditor' will
+// suppress the launching of the editor (and just do the copy or create, if
+// needed), and 'add' will also add the bundles to the mix.
+func (b *Builder) EditBundles(bundles []string, suppressEditor bool, add bool, git bool) error {
 	// Fetch upstream bundle files if needed
 	if err := b.getUpstreamBundles(b.UpstreamVer, true); err != nil {
 		return err
@@ -1005,25 +1027,30 @@ func (b *Builder) EditBundles(bundles []string, copyOnly bool, add bool, git boo
 	editorCmd, err := getEditorCmd()
 	if err != nil {
 		fmt.Println("Cannot find a valid editor (see usage for configuration). Copying to local-bundles only.")
-		copyOnly = true
+		suppressEditor = true
 	}
 
 	for _, bundle := range bundles {
-		var path string
-		path, err = b.getBundlePath(bundle)
-		if err != nil {
-			return err
-		}
-
+		path, _ := b.getBundlePath(bundle)
 		if !b.isLocalBundle(path) {
 			localPath := filepath.Join(b.LocalBundleDir, bundle)
-			if err = helpers.CopyFile(localPath, path); err != nil {
-				return err
+
+			if path == "" {
+				// Bunlde not found upstream, so create new
+				if err = createBundleFile(bundle, localPath); err != nil {
+					return errors.Wrapf(err, "Failed to write bundle template for bundle '%s'", bundle)
+				}
+			} else {
+				// Bundle found upstream, so copy over
+				if err = helpers.CopyFile(localPath, path); err != nil {
+					return err
+				}
 			}
+
 			path = localPath
 		}
 
-		if copyOnly {
+		if suppressEditor {
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


helpers/helpers.go
```diff
@@ -294,6 +294,7 @@ func RunCommand(cmdname string, args ...string) error {
 	cmd := exec.Command(cmdname, args...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
```
kevincwells: This is added to facilitate launching applications that read input (i.e., the editors). I considered adding the same thing to the `RunCommandOutput` function below, but didn't know how people would want it. Perhaps a separate `RunCommandInput` function that takes a `bytes.Buffer` as input? We can add that if we have a need.

I made sure the `helpers_tests` still ran as expected after adding this line.

mixer/cmd/bundles.go
```diff
@@ -114,10 +114,55 @@ var bundleListCmd = &cobra.Command{
...
+	Use:   "edit [bundle(s)]",
+	Short: "Edit local and upstream bundles",
+	Long: `Edit local and upstream bundle definition files. This command will locate the
+bundle (looking first in local-bundles, then in upstream-bundles), and launch
+an editor to edit it. If the bundle is only found upstream, the bundle file will
+first be copied to your local-bundles directory for editing. When the editor
+closes, the bundle file is then parsed for validity.
+
+The editor is configured in builder.conf via the BUNDLE_EDITOR field, or by flag
+override. Valid options include: nano, vim, emacs, joe, and ctags. The default
```
icarus-sparry: ctags is a valid editor?

builder/builder.go
```diff
@@ -836,6 +839,58 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
...
+			localPath := filepath.Join(b.LocalBundleDir, bundle)
+			if err := helpers.CopyFile(localPath, path); err != nil {
+				return err
+			}
+			path = localPath
+		}
+
+		if !noEdit {
+			if err := helpers.RunCommand(editorCmd, path); err != nil {
+				return errors.Wrapf(err, "Failed to edit bundle: %s", bundle)
```
icarus-sparry: This is high risk. Some editors return a non-zero exit code if some commands fail, which might be as simple as a search failing.

mixer/cmd/bundles.go
```diff
@@ -114,10 +114,55 @@ var bundleListCmd = &cobra.Command{
...
+value is nano.
+
+Passing '--no-edit' will suppress launching the editor, and will thus only copy
+the bundle file to local-bundles if it is only found upstream. This can be
+useful if you want to add a bundle to local-bundles, but wish to edit it at a
```
icarus-sparry: Is this a real use case? Would it be better as a different command?

Once you allow someone to choose their own editor, they can always use /bin/true to get the same effect.

mixer/cmd/bundles.go
```diff
@@ -114,10 +114,55 @@ var bundleListCmd = &cobra.Command{
...
+later time.
+
+Passing '--add' will also add the bundle(s) to your mix. Please note that
+bundles are added after all bunles are edited, and thus will not be added if any
```
icarus-sparry: spelling nitpick. "all bunles are"

helpers/helpers.go
kevincwells: Great idea. I'll redo this.

builder/builder.go
```diff
@@ -836,6 +839,58 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
...
+	"nano":  "/usr/bin/nano",
+	"vim":   "/usr/bin/vim",
+}
+
+//EditBundles prints out a bundle list in either a flat list or tree view
+func (b *Builder) EditBundles(bundles []string, noEdit bool, editor string, add bool) error {
+	if editor != "" {
+		b.BundleEditor = editor
+	}
+	editorCmd, exists := validEditors[b.BundleEditor]
```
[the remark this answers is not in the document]
kevincwells: This is actually exactly the original way I had this implemented -- if the user sets (or passes in) the path to the editor, use that, else try to look up $VISUAL, then $EDITOR, then fall back to nano.

The reason I went away from this is that, while this flexibility is awesome, basic security policy says to explicitly _not_ "trust your users!".  Whatever you configure here is going to just be passed off to an `exec`, effectively. Unlike `git`, which lets you configure whatever editor you want for merge conflicts, `mixer` is run via `sudo`. Yes, _these_ commands require no elevated privileges, but when so many of the others do, users will absolutely run them with `sudo` either out of a) laziness, b) misunderstanding, or c) because the files they're editing are owned by root because of some other operation.

As far as the list not including xemacs or gvim, I just pulled the list from the packages included in `editors` bundle -- if those packages include other variants I left out, I'm happy to add them.

builder/builder.go
```diff
@@ -836,6 +839,58 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
...
+			}
+
+			if _, err := parseBundleFile(path); err != nil {
+				return errors.Wrapf(err, "Failed to parse bundle %s. Please edit again", bundle)
+			}
+		}
+	}
+
+	if add {
+		if err := b.AddBundles(bundles, false, false, false); err != nil {
```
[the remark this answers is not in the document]
kevincwells: Nice, thanks.

builder/builder.go
```diff
@@ -836,6 +839,58 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
...
 
+var validEditors = map[string]string{
+	"joe":   "/usr/bin/joe",
+	"emacs": "/usr/bin/emacs",
+	"ctags": "/usr/bin/ctags",
+	"nano":  "/usr/bin/nano",
+	"vim":   "/usr/bin/vim",
+}
+
+//EditBundles prints out a bundle list in either a flat list or tree view
```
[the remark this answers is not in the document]
kevincwells: Oh, whoa, nope. Copypasta error.

mixer/cmd/bundles.go
kevincwells: Haha, is it not? Basic ignorance issue here. As mentioned above, I pulled the list from what packages are installed by `editors`, but I obviously should have done a basic google. Ctags is a package used by one of the actual editors?

I'll pull it.

mixer/cmd/bundles.go
kevincwells: Honestly, this is something people asked for. I don't remember exactly who or when, but when I proposed this feature back in December, this is the feedback I got.

Also, while yes -- if you could pick any editor you want, you could use `/bin/true` -- expecting users to know or do that is the opposite of user-friendly.

builder/builder.go
kevincwells: That's a great point.  In theory, I don't _really_ care what the editor returned, I just care if I can read/parse it afterward. I should probably just ignore the error here, and let the parsing trigger the error.

builder/builder.go
```diff
@@ -836,6 +839,58 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
...
+
+		if !b.isLocalBundle(path) {
+			localPath := filepath.Join(b.LocalBundleDir, bundle)
+			if err := helpers.CopyFile(localPath, path); err != nil {
+				return err
+			}
+			path = localPath
+		}
+
+		if !noEdit {
```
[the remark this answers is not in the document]
kevincwells: I like this, because it removes the double-negative (and some indentation).

builder/builder.go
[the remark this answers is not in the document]
kevincwells: Just to be clear (as I hope is evidenced by the fact that I originally implemented it the way you suggested), I also don't love the idea of limiting this to a small list of editors. I am absolutely open to ways to open this up that don't introduce mixer executing random binaries as root. :)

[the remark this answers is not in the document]
cmarcelo: I do understand your concern, but I think you are fixing the problem in the wrong place. We have other of those customizations of what program to call (BuildScript, --packager), that I think is unfeasible to whitelist all of them.

Also, we really don't want the user to create those files with root permissions (if he's keeping track of the repository in git for example).

I think the real way to deal with this is to contain where we really need sudo, and possibly call sudo ourselves when we really need (which also might be controversial, but I think is a fight on the right battle).

For this functionality (which I think is very good btw!), I'd say either:
- Just do what you did before or Icarus is suggesting
- Avoid editing for now and keep in some form the "copy this manifest from upstream" functionality
- Figure out a way to safely "drop privileges" by setuid yourself back to the original user when running this command (kind of the reverse of sudo). I don't think Go handles that very well yet, though. (See their issue 1435).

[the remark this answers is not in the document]
cmarcelo: > Personally I would be inclined to not have an option to pass in the editor via a flag at all.

I agree with that one, we can defer the flag until a real need surfaces. The EDITOR and [name]_EDITOR are common enough in the command line.

mixer/cmd/bundles.go
```diff
@@ -114,10 +114,55 @@ var bundleListCmd = &cobra.Command{
...
+bundle (looking first in local-bundles, then in upstream-bundles), and launch
+an editor to edit it. If the bundle is only found upstream, the bundle file will
+first be copied to your local-bundles directory for editing. When the editor
+closes, the bundle file is then parsed for validity.
+
+The editor is configured in builder.conf via the BUNDLE_EDITOR field, or by flag
+override. Valid options include: nano, vim, emacs, joe, and ctags. The default
+value is nano.
+
+Passing '--no-edit' will suppress launching the editor, and will thus only copy
```
cmarcelo: `mixer bundle edit --no-edit` sounds strange, maybe `mixer bundle edit --copy-only`?

mixer/cmd/bundles.go
cmarcelo: I think that is an important part of the feature, because that is a boring computer task that mixer can do for us: hey just get me the desktop and desktop-apps from the upstream. I've almost suggested something like `mixer bundle copy desktop desktop-apps` (or copy-upstream, or import, or whatever) as a standalone thing (so you don't have to deal with the "edit but no edit").

cmarcelo:
Besides the nits and the list of valid editors, this looks good to me.

builder/builder.go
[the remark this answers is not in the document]
icarus-sparry: The problem is that even if you restrict the user to only running 'vi' you can fire up vi. You then type ":sh" and enter, and you have a shell running as root! At that point you can run anything you like as root.

sudo is the wrong tool for increasing privileged status. We ought to see exactly what wants 'root', and see if we can eliminate the need as a medium term goal.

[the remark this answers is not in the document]
kevincwells: > We have other of those customizations of what program to call (BuildScript, --packager), that I think is unfeasible to whitelist all of them

`BuildScript` is hard-coded (and not read in from the config), and for `--packager`, you really are running `exec` on whatever the user passes in? I didn't review that part of the new BCB, but I would have definitely flagged that if I had.

> Also, we really don't want the user to create those files with root permissions (if he's keeping track of the repository in git for example).

Agreed.

> I think the real way to deal with this is to contain where we really need sudo

:+1: 

> Figure out a way to safely "drop privileges" by setuid yourself back to the original user when running this command (kind of the reverse of sudo). I don't think Go handles that very well yet, though. (See their issue 1435).

I've done this in other languages before. I'll look into if this is feasible.

mixer/cmd/bundles.go
icarus-sparry: ctags produces a 'tags' database so you can find definitions and uses of variables in C projects.

mixer/cmd/bundles.go
kevincwells: Hahah, I had that same thought last night. I couldn't come up with a better name (it was late and I was hungry), but that works pretty well.

builder/builder.go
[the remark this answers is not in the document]
cmarcelo: > We ought to see exactly what wants 'root', and see if we can eliminate the need as a medium term goal.

@icarus-sparry  created #158 with some more data on this. Want to tackle that problem? :-)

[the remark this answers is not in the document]
cmarcelo: > BuildScript is hard-coded (and not read in from the config),

That was surprising. I guess it was made that way to be configurable at some point. That should be a constant.

> and for --packager, you really are running exec on whatever the user passes in? I didn't review that part of the new BCB, but I would have definitely flagged that if I had.

Yep, to let the user pick some other dnf (either the system or a compiled one for testing). In that case, we could've retrofit it to be a flag that just says "use dnf" and use whatever is in PATH.

I guess the two examples were not that strong but I stand by the other arguments I gave :-)

mixer/cmd/bundles.go
kevincwells: @cmarcelo I had considered at one point making that a stand-alone `mixer bundle import` command, or even a flag off `add` (`mixer bundle add --local <bundle>`).

The reason I don't want that is that, while people might very well find it useful to pull things into `local-bundles` and edit at some later time, I want them to still be aware that doing this _overrides_ the upstream version. If you're not actually planning to edit the bundle, having a copy in `local-bundles` is almost certainly just going to cause confusion later (either for you, or for the tool when it tries to migrate to new upstream version). So keeping it a part of the `mixer bundle edit` command reminds people why they're moving things into `local-bundles`: to edit them.

On that note, one (potentially valid, but dangerous) use-case for copying into local and _not_ editing is exactly that -- persisting a bundle past a deletion in a future upstream version. What I mean is, if you add bundle_a to `local-bundles`, and in a future version of upstream bundle_a gets removed entirely, it now becomes equivalent to a "user-created" bundle -- it is no longer masking anything upstream. That very well might be a desirable thing for someone, but for most people that will likely just cause confusion.

kevincwells:
Pushed a new version that addresses some feedback. Still need to swap out the validEditors and, ideally, address @icarus-sparry's usability complaints (which I agree with)

kevincwells:
I just pushed an update that I believe addresses @icarus-sparry and @cmarcelo's issues.

The editor is now read from `VISUAL`, then `EDITOR`, then defaults to `nano`. If `nano` cannot be found, it skips editing and acts as if `--copy-only` was passed.

If parsing fails, the user is told why and prompted to choose between re-editing as-is, reverting and editing fresh, or skipping. If they skip, both the edited and a backed-up `<bundle>.orig` file are left behind, with output to the user telling them so.

kevincwells:
FYI - `mixer bundle list` does not filter out invalid bundles when it runs, so if you run `mixer bundle list --local` while you still have a `bundle.orig` file laying around, it'll error out saying that is an invalid bundle name. Frankly, I think this is probably a good thing, so people don't forget that they're in a broken state.

If people think this is bad, I would suggest a future patch (not part of this) that either a) skipps files with ".orig" at the end, or b) does the `validBundleNameRegex.MatchString` currently in `parseBundleFile` for something more thorough.

builder/builder.go
```diff
@@ -914,6 +915,124 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
...
+		// Ignore return from command; parsing below is what will reveal errors
+		_ = helpers.RunCommandInput(os.Stdin, editorCmd, path)
+
+		_, err := parseBundleFile(path)
+		if err == nil {
+			// Clean-up backup
+			if err = os.Remove(backup); err != nil {
+				return errors.Wrapf(err, "Error cleaning up backup for bundle '%s'", bundle)
+			}
+			break editLoop
```
icarus-sparry: Since all that 'break editloop' does is 'return nil', in this case you could just return nil. However leave it as it is, as the structure is nice.

builder/builder.go
```diff
@@ -914,6 +915,124 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
...
+			// Ask the user if they want to retry, revert, or skip
+			fmt.Print("Would you like to continue editing as-is, revert and edit, or skip? [c/r/s]?: ")
+			text, err := reader.ReadString('\n')
+			if err != nil {
+				return errors.Wrapf(err, "Error reading input")
+			}
+			text = strings.ToLower(text)
+			text = strings.TrimSpace(text)
+			switch {
+			case text == "c" || text == "continue":
```
icarus-sparry: When I suggested reedit/continue/restore I was rather thinking of editing more than one bundle.

So I would expect 'continue' to leave the current bundle and go on to the next, 'reedit' to rerun the editor on the current file, and reset to run the editor on a copy of the original file.

So your 'continue' is my 'reedit', your 'skip' is my 'continue' and your 'revert' is my 'restore'.

Let me first say there is nothing wrong with your names, but then say "More explanation is probably needed in the prompt text' so it is clear what 'continue' means.

kevincwells: Updated.

kevincwells:
@cmarcelo  is right, and the functionality I was putting in #171 belongs here. Gets rid of a lot of similar code, too!

This command will work exactly as it does now, except if they're editing a bundle that doesn't exist upstream, it'll "create" a new bundle for them and have them edit the template.

builder/builder.go
```diff
@@ -914,6 +915,143 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
...
+		for {
+			// Ask the user if they want to retry, revert, or skip
+			fmt.Print("Would you like to edit as-is, revert and edit, or skip [Edit/Revert/Skip]?: ")
+			text, err := reader.ReadString('\n')
+			if err != nil {
+				return errors.Wrapf(err, "Error reading input")
+			}
+			text = strings.ToLower(text)
+			text = strings.TrimSpace(text)
+			switch {
```
cmarcelo: You can switch on strings in Go, so `switch text {` then use `case "e", "edit":` and so on.

kevincwells: Huh. I knew you could switch on strings, but I'd never seen the `case "e", "edit":` syntax before. I definitely prefer that.

kevincwells:
Pushed change @cmarcelo suggested. It's pushed as a new commit, so people can review just that one, without worrying about reviewing the entire `mixer bundle edit` command again.

cmarcelo:
LGTM.

PR #156, opened 2018-02-08
Add -cover flag to go test

matthewrsj:
Make go test (make check) report the coverage percentage of the unit
tests that run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,16 @@ install: gopath
 	install -D -m 00644 yum.conf.in $(DESTDIR)/usr/share/defaults/mixer/yum.conf.in
 
 check: gopath
-	go test ${GO_PACKAGE_PREFIX}/...
+	go test -cover ${GO_PACKAGE_PREFIX}/...
+
+# TODO: when Go 1.10 comes out we will have support for passing multiple packages
+# to coverprofile, so there will be no need to pass an individual package.
+# At that time we can merge this target into check and run it against all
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,7 +45,11 @@ install: gopath
 	install -D -m 00644 yum.conf.in $(DESTDIR)/usr/share/defaults/mixer/yum.conf.in
 
 check: gopath
-	go test ${GO_PACKAGE_PREFIX}/...
+	go test -cover ${GO_PACKAGE_PREFIX}/...
+
+checkcoverage: gopath
+	test ${PKG}
```
cmarcelo: You might like to know that in Go 1.10 we'll have support for passing multiple packages with a coverprofile, so no need to do pkg. :-)

https://github.com/golang/go/commit/283558e42b88a6afa39da6ad4ae87558dc053776

Optional: add a TODO so we eventually use that capability.

Makefile
```diff
@@ -45,7 +45,11 @@ install: gopath
...
+	go test -cover ${GO_PACKAGE_PREFIX}/${PKG} -coverprofile=coverage.out
```
cmarcelo: Since checkcoverage seems to be focused on developer use (and not the CI, that uses check), I wonder if you could also call the coverage visualization tool, adding this line to the rules

    go tool cover -html=coverage.out

cmarcelo:
Approved again.

PR #157, opened 2018-02-08
Add index manifest to master manifest list

matthewrsj:
The index manifest must be added to the master manifest list in order to
later be compressed for download by client and so the chroot is cleaned
up during the minimize chroots step.


swupd/manifest.go
```diff
@@ -767,7 +767,7 @@ func constructIndex(c *config, ui *UpdateInfo, f2b []*bundleIndex) error {
 // sorts the index first by filename then by bundle name. writeIndexManifest creates a new
 // bundle in which the file will live. This bundle is added to the "full" manifest which
 // is part of the bundles slice. A path string to the new manifest is returned on success.
```
tmarcu: Update comment to Manifest pointer is returned.

matthewrsj: ah, thanks

PR #160, opened 2018-02-08
Strip dot characters from rename filepaths

matthewrsj:
In addition to digits strip '.' characters from filepaths for rename
partialNames because dots are common in versioned paths.

This enables renames to be detected against a higher number of files as
their versions change.

### DIFF
--- a/swupd/rename.go
+++ b/swupd/rename.go
@@ -135,10 +135,15 @@ type pairedNames struct {
 	partialName string // Filename with digits removed
 }
 
-// stripDigits is used with strings.Map to remove digits from filename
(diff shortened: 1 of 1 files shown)


matthewrsj:
When looking through the manifests I saw cases where versions would go from `2.2` to `2.2.1` and similar. Not all projects use semver which is why I thought it would be useful to add this. 

The link to file thing doesn't matter because we only perform renames on the same file type right?

This was just an improvement idea. If it won't be that helpful I don't mind dropping it.

icarus-sparry:
I agree not everyone uses semver. If there are cases where it helps then go for it.

PR #161, opened 2018-02-09
Add script used to validate new swupd code

matthewrsj:
This script creates builds from $1 to $2 if they exist upstream and
performs an os-install, bundle-add, bundle-remove, verify, and update on
each.

### DIFF
--- a/validate-swupd.sh
+++ b/validate-swupd.sh
@@ -50,19 +50,25 @@ for i in $(seq $1 10 $2); do
 		popd
 		exit 1
 	fi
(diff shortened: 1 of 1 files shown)


cmarcelo:
This worked for me when testing yesterday. Mentioned some small "usability" things with Matthew, will record them here:

- Show a usage message when called in isolation.
- Do not clutter the current directory by default.

Also, updates to take "--force" out from the call to bundle add.

matthewrsj:
done.

PR #165, opened 2018-02-09
Only support File->File renames

matthewrsj:
Only support renaming a file to a file since renames are not supported
for the other cases. This is done in trimRenamed by only appending to
the rename list if the file is deleted/ghosted or is a TypeFile. It is
necessary to allow all deleted/ghosted files through because we have
lost the file type association at this point. At least the added list
will only contain files and therefore directories and links will not be
renamed.

Add a TODO to stop removing file type information from deleted files.

### DIFF
--- a/swupd/rename.go
+++ b/swupd/rename.go
@@ -106,7 +106,15 @@ func trimRenamed(a []*File) []*File {
 		// and it is helpful to be able to ship just the differences for them.
 		// Worst case is that the files do not exist on the target system, in which
 		// case the swupd-client will fall back to doing a full download.
-		if f.DeltaPeer == nil {
+		//
+		// Renames are only supported for TypeFile right now so skip if the file is
+		// not the right type.
+		//
+		// Unfortunately deleted files have lost their association with a file type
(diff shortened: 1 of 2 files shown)


swupd/rename.go
```diff
@@ -106,7 +106,15 @@ func trimRenamed(a []*File) []*File {
...
-		if f.DeltaPeer == nil {
+		//
+		// Renames are only supported for TypeFile right now so skip if the file is
+		// not the right type.
+		//
+		// Unfortunately deleted files have lost their association with a file type
+		// so we have to leave all deleted files in the list
+		// TODO: refactor to leave file type in the deleted file records. There is no
+		// reason to clear them.
+		if f.DeltaPeer == nil && (!f.Present() || f.Type == TypeFile) {
```
cmarcelo: Looks correct to me.

Optional: because we use trimRenamed more than once in the function, we are performing redundant checks. In the second time we use, we are calling it composed with trimSmall, reallocating stuff twice. I'd consider renaming (!) trimSmall and include the DeltaPeer check there, but not this check, that would be on the first trim call.

PR #166, opened 2018-02-09
Fix CopyFile bug

kevincwells:
Previously, helpers.CopyFile left off os.O_TRUNC, meaning that
if the destination is larger than the source, the remainder of
the destination remained at the end. Now the file is overwritten
completely.


helpers/helpers.go
```diff
@@ -198,10 +198,11 @@ func UnpackFile(file string, dest string) error {
 	return nil
 }
 
-// CopyFile is used during the build process to copy a given file to the target
-// instead of dealing with the particulars of hardlinking.
+// CopyFile copies a file, overwriting the destination if it exists. It is used
+// during the build process to copy a given file to the target instead of
+// dealing with the particulars of hardlinking.
```
cmarcelo: Optional: I'd cut the comment about where it is used.

kevincwells: Yeah, almost did. Should have.

kevincwells: Did.

helpers/helpers.go
```diff
@@ -198,10 +198,11 @@ func UnpackFile(file string, dest string) error {
...
 func CopyFile(dest string, src string) error {
-	return copyFileWithFlags(dest, src, os.O_RDWR|os.O_CREATE)
+	return copyFileWithFlags(dest, src, os.O_RDWR|os.O_CREATE|os.O_TRUNC)
```
icarus-sparry: Makes me wonder if there are other users of copyFileWithFlags which are not doing the right thing, but the one other use (CopyFileNoOverwrite) looks fine.

Good catch.

rchiossi:
+1

PR #167, opened 2018-02-10
New Command: mixer bundle remove

kevincwells:
Removes bundles from your mix by modifying the Mix Bundle List
(stored in the 'mixbundles' file). The Mix Bundle List is parsed, the bundles
are removed, and the resultant list is written back out in sorted order. If
bundles do not exist in the mix, they are skipped.

Passing `--local` will also remove the corresponding bundle definition file from
local-bundles, if it exists. Please note that this is an irrevocable step.

`--mix` defaults to true. Passing `--mix=false` will prevent the bundle from
being removed from your Mix Bundle List. This is useful when used in conjunction
with `--local` to *only* remove a bundle from local-bundles. If the bundle being
removed is an edited version from upstream, the bundle will remain in your mix
and now reference the original upstream version. If the bundle was custom, and
no upstream alternative exists, a warning will be returned.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -739,6 +739,78 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool, 
 	return nil
 }
 
+// RemoveBundles removes a list of bundles from the Mix Bundles List. If a
+// bundle is not present, it is skipped. If 'local' is passed, the corresponding
+// bundle file is removed from local-bundles. Note that this is an irreversible
+// step. The Mix Bundles List is validated when read in, and the resulting Mix
+// Bundles List will be in sorted order.
+func (b *Builder) RemoveBundles(bundles []string, mix bool, local bool, git bool) error {
+	// Fetch upstream bundle files if needed
+	if err := b.getUpstreamBundles(b.UpstreamVer, true); err != nil {
+		return err
+	}
+
+	// Read in current mix bundles list
+	set, err := b.getMixBundlesListAsSet()
+	if err != nil {
+		return err
+	}
+
+	// Remove the ones passed in from the set
+	for _, bundle := range bundles {
+		_, inMix := set[bundle]
+
+		if local {
+			if _, err := os.Stat(filepath.Join(b.LocalBundleDir, bundle)); err == nil {
+				fmt.Printf("Removing bundle file for '%s' from local-bundles\n", bundle)
+				if err := os.Remove(filepath.Join(b.LocalBundleDir, bundle)); err != nil {
+					return errors.Wrapf(err, "Cannot remove bundle file for '%s' from local-bundles", bundle)
+				}
+
+				if !mix && inMix {
+					// Check if bundle is still available upstream
+					if _, err := b.getBundlePath(bundle); err != nil {
+						fmt.Printf("Warning: Invalid bundle left in mix: %s\n", bundle)
+					} else {
+						fmt.Printf("Mix bundle '%s' now points to upstream\n", bundle)
+					}
+				}
+			} else {
+				fmt.Printf("Bundle '%s' not found in local-bundles; skipping\n", bundle)
+			}
+		}
+
+		if mix {
+			if inMix {
+				fmt.Printf("Removing bundle '%s' from mix\n", bundle)
+				delete(set, bundle)
+			} else {
+				fmt.Printf("Bundle '%s' not found in mix; skipping\n", bundle)
+			}
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


builder/builder.go
```diff
@@ -739,6 +739,55 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool,
...
+		return err
+	}
+
+	// Remove the ones passed in from the set
+	for _, bundle := range bundles {
+		fmt.Printf("Removing bundle %s\n", bundle)
+		delete(set, bundle)
+
+		if local {
+			if err := os.RemoveAll(filepath.Join(b.LocalBundleDir, bundle)); err != nil {
```
matthewrsj: `RemoveAll`? Do we expect this to be a directory?

kevincwells: Nope. As @cmarcelo pointed out to me in the past, `os.RemoveAll` doesn't return an error in the event that the file does not exist, and thus you can save yourself the call to `os.Stat` you'd need with `os.Remove` if you're not sure it's there. This kind of short-hand could be dangerous if you're removing files in a shared or more-complicated directory (where you might care, for example, if you're accidentally deleting a directory rather than a file), but this seemed pretty safe to me here.

matthewrsj: Sure, but `os.RemoveAll` does an `Lstat` of its own (https://golang.org/src/os/path.go?s=1634:1667#L56). This is fine it just seems strange to me to be potentially removing a directory when you only want to remove a file.

kevincwells: Yeah -- I meant it saves _our code_ from doing an `Stat`. Just cleaner/less nesting. And I mean, `os.Remove` can remove a directory too. Are you suggesting I have a separate `os.State` to check the `FileInfo` for `IsDir()`?

I do like the idea of supporting nested directories in `local-bundles` down the line (so that you could organize your custom bundles how you pleased) -- almost included that in the original implementation, but figured I'd mirror how `upstream-bundles` worked first, and add that as an optimization if people wanted it. If we start supporting a non-flat `local-bundles`, this section (and a couple other places) would definitely need to be more sophisticated to accommodate that.

matthewrsj: Ok, fair enough.

mixer/cmd/bundles.go
```diff
@@ -131,5 +164,8 @@ func init() {
 	bundleAddCmd.Flags().BoolVar(&bundleAddFlags.allUpstream, "all-upstream", false, "Add all upstream bundles; takes precedence over bundle list")
 	bundleAddCmd.Flags().BoolVar(&bundleAddFlags.git, "git", false, "Automatically apply new git commit")
 
+	bundleRemoveCmd.Flags().BoolVar(&bundleRemoveFlags.local, "local", false, "Also remove bundle file from local-bundles (irrevocable)")
```
cmarcelo: Optional: I'd remove this flag for now and see if we miss it.

mixer/cmd/bundles.go
```diff
@@ -65,6 +65,38 @@ resultant list is written back out in sorted order.`,
...
+	local bool
+	git   bool
+}
+
+var bundleRemoveFlags bundleRemoveCmdFlags
+
+var bundleRemoveCmd = &cobra.Command{
+	Use:   "remove [bundle(s)]",
+	Short: "Remove bundles from your mix",
+	Long: `Removes bundles from your mix by modifying the Mix Bundle List
```
cmarcelo: Thanks for writing long description!

mixer/cmd/bundles.go
kevincwells: Two reasons I put this in:

1. It's a use case I often have when testing a mix
1. I want some fashion of validity checking. Doing it this way binds removing the local bundle definition file with the removal of that bundle from the Mix Bundle List. If you _don't_ bind these actions (i.e., if you manually `rm` the file on your own), you can be left with a bundle in your mix that doesn't exist. The caveat of this is that if you want to **only** remove the your edited copy (but still have the bundle in your mix, referring to upstream), you'd have to _re-add_ the bundle after running this.

I'm planning to push an update to this that introduces a `--mix` flag that defaults to true. That would mean the behavior described above would be the default, but if you want, you can do `mixer bundle remove editors --local --mix false`. That would keep `editors` in your mix, but just remove your edited copy. If `--mix` is false, I can easily check if that bundle still exists upstream, and error to the user if they're leaving a non-existent bundle in their mix.

builder/builder.go
kevincwells: I decided to change this after all, and include the additional `os.Stat` with additional output, 'cause in testing I found it way easier to understand what's going on.

PR #168, opened 2018-02-12
Sensitively load groups.ini file

matthewrsj:
Some bundles have uppercase letters in their names. DO NOT make these
lowercase - this has the result of a manifest not being generated for
them because it cannot find the chroot to walk.


phmccarty:
+1

PR #171, opened 2018-02-12
New Command: Mixer bundle create

kevincwells:
**Note**: This PR is based on top of #155, because it uses the `EditBundles` function introduced there. **If you are reviewing code, please only review the new commit.**

-------

Create custom local bundle definition files. This command will generate an
empty bundle definition file template with the appropriate name in your
local-bundles directory, and launch an editor to edit it. When the editor
closes, the bundle file is then parsed for validity. If a bundle already exists
in local-bundles, it will be skipped.

The editor is configured via environment variables. VISUAL takes precedence to
EDITOR. If neither are set, the tool defaults to nano. If nano is not installed,
the tool will skip editing, and act as if '--create-only' had been passed.

Passing '--create-only' will suppress launching the editor, and will thus only
create the empty bundle template in local-bundles. This can be useful if you
want to create a new bundle, but wish to edit it at a later time.

Passing '--add' will also add the bundle(s) to your mix. Please note that
bundles are added after all bundles are created, and thus will not be added if
any errors are encountered earlier on.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -16,6 +16,7 @@ package builder
 
 import (
 	"archive/tar"
+	"bufio"
 	"bytes"
 	"crypto/rsa"
 	"crypto/x509"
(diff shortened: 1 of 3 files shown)


cmarcelo:
Should we do something if the user asks to create a bundle that already exists upstream?

cmarcelo:
I like (2) too, but then I think we don't a separate create command: couldn't I just use`mixer bundle edit newbundle` and get the bundle template format?

kevincwells:
> I like (2) too, but then I think we don't a separate create command: couldn't I just usemixer bundle edit newbundle and get the bundle template format?

That's... yeah, that's a good point. The behavior of `mixer bundle edit` could easily be `edit a blank template` if it's not upstream.

Issue #54, opened 2017-12-23
Config parsing


kevincwells:
# Proposal
Rewrite the mixer code that parses `builder.conf` using the [Viper library](https://github.com/spf13/viper).

# Benefits
- Remove fragile regex-based parsing
- Correctly support config "Section" to value affiliation (right now sections are ignored)
- Allow easy generation of default `builder.conf`
- Integrate with Cobra to fluidly support CLI flag overrides of config values (optional).

# Notes
Viper will treat `builder.conf` as a TOML file. Technically, as it is read by `bundle-chroot-builder.py`, which relies on the Python [ConfigParserr](https://docs.python.org/3/library/configparser.html) library, it is not a full TOML file, and instead limited to the subset INI format. This should not be a problem with current usage.

cmarcelo:
I would take the opportunity to:
- Remove any unused or redundant configurations if still exist
- Review the names of the configurations (get feedback from people working on the tools around)
- Review how we handle the default values: my overall impression is that the configuration feels repetitive, most of time there'll be a "main" work directory and I'm fine with all the default names for the subdirectories

(I also think is fine to split this off into a separate task, but felt it fits the overall config theme).

tmarcu:
@rchiossi working on this right now.

tmarcu:
@rchiossi close this ?

rchiossi:
Yep. Closing.

PR #172, opened 2018-02-13
Use go-ini for builder.conf parsing

rchiossi:
So, to @cmarcelo 's delight, I ended up giving go-ini a shot.

This PR separates the all config related structures and functions into a new  `mixconfig.go` file, so if we ever decide to change to a new config system, there shouldn't be much impact in the rest of the code.

Some notes about this implementation:
- Go-Ini has a mapper to map a config file to a structure, but it can't be used in our case  due to a bunch of issues:
-- The section name has the name of a type structure, which means that for section Builder we need another Builder object. This would require moving mixconfig to a separate package to avoid conflict with the actual Builder structure
-- swupd section has lowercase name which mean a structure with that name would not be exported. Go-Ini ignores non-exported structures
-- None of the name converters worked in our case. One converter would have the key names match the variable name (e.g. RepoDir="" instead of REPO_DIR="") and the other would create some funny side effects (e.g. R_P_M_DIR="" instead of RPM_DIR="")
- In order to reduce the change, I embedded the MixConfig structure inside Builder. This is the less verbose solution, but it is simple to move it into its own variable if needed for clarity.
- I made this out of my frustration with viper in the past two days, so I haven't touched the swupd config APIs yet.

Some notes about why go-ini and not Viper:
Even though Viper is much simpler to use, it has so many small issues that at the end of the day it was not paying off. Here are a couple of issues I had with it:
- When loading keys, Viper expects it to be in the format Section.Key. It will pass that string to `ToLower()`, so both the section and the key will be lowercase. This makes the config file incompatible with other tools like 
`bundle-chroot-builder.py`, which expects the section `Builder` instead of `builder`.
- When saving string values, Viper will save them in the format `key="value"` but again it is incompatible with bcb since python's configparser expects it to be in the format `key=value` (no double quote)
- Viper does not work well with config files were the extension does not reflect its format and requires some workaround to deal with them (e.g. .conf instead of .toml)

Given that Viper also had a lot of dependencies (7 iirc), I decided to try and implement an alternative with go-ini using a similar structure to what I had done for viper.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -28,7 +28,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -52,25 +51,18 @@ var UseNewChrootBuilder = false
 // A Builder contains all configurable fields required to perform a full mix
 // operation, and is used to encapsulate life time data.
 type Builder struct {
+	MixConfig
+
 	BuildScript string
 	BuildConf   string
 
-	BundleDir       string
-	Cert            string
-	Format          string
-	LocalBundleDir  string
 	MixVer          string
 	MixVerFile      string
 	MixBundlesFile  string
-	RepoDir         string
-	RPMDir          string
-	StateDir        string
 	UpstreamURL     string
 	UpstreamURLFile string
 	UpstreamVer     string
 	UpstreamVerFile string
-	VersionDir      string
-	YumConf         string
 	YumTemplate     string
 
 	Signing int
@@ -99,56 +91,10 @@ func NewFromConfig(conf string) (*Builder, error) {
 	if err := b.LoadBuilderConf(conf); err != nil {
 		return nil, err
 	}
-	if err := b.ReadBuilderConf(); err != nil {
-		return nil, err
-	}
 	if err := b.ReadVersions(); err != nil {
 		return nil, err
 	}
 	return b, nil
-}
-
-// CreateDefaultConfig creates a default builder.conf using the active
-// directory as base path for the variables values.
-func (b *Builder) CreateDefaultConfig(localrpms bool) error {
-	pwd, err := os.Getwd()
-	if err != nil {
-		return err
-	}
-
-	builderconf := filepath.Join(pwd, "builder.conf")
-
-	err = helpers.CopyFileNoOverwrite(builderconf, "/usr/share/defaults/bundle-chroot-builder/builder.conf")
-	if os.IsExist(err) {
-		// builder.conf already exists. Skip creation.
-		return nil
-	} else if err != nil {
-		return err
-	}
-
-	fmt.Println("Creating new builder.conf configuration file...")
-
-	raw, err := ioutil.ReadFile(builderconf)
-	if err != nil {
-		return err
-	}
-
-	// Patch all default path prefixes to PWD
-	data := strings.Replace(string(raw), "/home/clr/mix", pwd, -1)
-
-	// Add [Mixer] section
-	data += "\n[Mixer]\n"
-	data += "LOCAL_BUNDLE_DIR=" + filepath.Join(pwd, "local-bundles") + "\n"
-
-	if localrpms {
-		data += "LOCAL_RPM_DIR=" + filepath.Join(pwd, "local-rpms") + "\n"
-		data += "LOCAL_REPO_DIR=" + filepath.Join(pwd, "local-yum") + "\n"
-	}
-
-	if err = ioutil.WriteFile(builderconf, []byte(data), 0666); err != nil {
-		return err
-	}
-	return nil
 }
 
 // initDirs creates the directories mixer uses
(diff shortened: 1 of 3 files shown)


builder/mixconfig.go
```diff
@@ -0,0 +1,194 @@
...
+	builderconf := filepath.Join(pwd, "builder.conf")
+
+	if localrpms {
+		config.RPMDir = filepath.Join(pwd, "local-rpms")
+		config.RepoDir = filepath.Join(pwd, "local-yum")
+	}
+
+	fmt.Println("Creating new builder.conf configuration file...")
+
+	cfg, err := config.mapToIni()
```
cmarcelo: Prefer a template approach to generate the config for the following reasons:

(a) we can opt to not add some values (or add them commented), to let the user take the default
(b) we can add some other comments into the configuration file (like explain sections and how certain variables are used).

builder/mixconfig.go
```diff
@@ -0,0 +1,194 @@
...
+	key   string
+	value *string
+}
+
+type configMapping struct {
+	section string
+	pairs   []configPair
+}
+
+func (config *MixConfig) getMapping() []configMapping {
```
cmarcelo: I see how mapping made sense if we are writing and reading, but I think you should just drop it and use a couple of switch statements to handle the cases. Reasons:

(a) in case of some options, we really want to have lists (bundle list) or numbers (versions, in the future), so we would need can do a preparse on the spot.
(b) we can add more validation as part of the reading of those values, i.e. never have an invalid value in config.

My suggestion is remove getMapping(), inline mapFromIni() into LoadConfig. To avoid long indentation, for each section you may want to call a different function to parse that section or something similar.

builder/builder.go
```diff
@@ -52,25 +51,18 @@ var UseNewChrootBuilder = false
 // A Builder contains all configurable fields required to perform a full mix
 // operation, and is used to encapsulate life time data.
 type Builder struct {
+	MixConfig
```
cmarcelo: This was a nice move to make your change more focused.

builder/mixconfig.go
kevincwells: > Prefer a template approach to generate the config

While I agree with both of your reasons when it comes to creating the default config file, I strongly want the ability to write the config _from a data structure_.  Ultimately, I want the ability to move the Mix Bundles List out of `mixbundles` and into `builder.conf`, and that means "editing the Mix Bundles List" means editing `builder.conf`.

builder/mixconfig.go
kevincwells: @cmarcelo See my other comment about wanting the ability to write, not just read, the config programmatically.

kevincwells:
So first things first, I think what you wrote is pretty cool. But overall this is not the direction I'd like things to go. Or at least not where I want them to end up. But as an intermediate? I mean, it's definitely better than the regex-matching, section-ignoring solution we have now. And for #137's introduction of `LOCAL_BUNDLE_DIR` and whatnot, we need a solution that accepts default values for things not currently in people's existing `builder.conf`. (That said, it'd be easy to throw a patch in that simply pre-loads these (or even _everything_) with default values up where `YumTemplate` and whatnot get set... Why aren't we already doing that?)

Some thoughts:
- I love separating config out of `builder.go`. This should happen no matter what tool/library/solution we end up going with.
- While I understand for read/write simplicity why values currently not in `builder.conf` (e.g., `YumTemplate`) didn't get moved over to `MixConfig`, as a _user_ of `Builder`, this feels arbitrary. (That is, some things that "feel like configuration" are in `Builder.MixConfig`, and others are just in `Builder`.) Feels like we could have a cleaner design.
- I like reusing libraries we're already using, where appropriate.
- As I've said before, I _really_ want to move away from INI to TOML. If we don't end up using Viper (or any other _configuration_ library), I still want us to use a TOML _parsing_ library. There are several. Viper used [go-toml](https://github.com/pelletier/go-toml), but I see [toml](https://github.com/BurntSushi/toml) mentioned more often. I haven't looked closely at either, but honestly, you could probably swap what you have with one of these and only have to change a couple lines, as our _current_ `builder.conf` is valid basic TOML. While `go-ini` at least supports comments and nested sections, it doesn't appear to support any form of list (you'd have to fake it with comma-separated strings that you split on your own). And INI is just the worst.
  - In regards to above comment about reusing libraries, I'd much prefer going all-TOML rather than trying to stay all-INI.
- It feels wrong to me to not use the built-in map-to-data-structure interface. It feels like Go often encourages you to reinvent the wheel, but I would very strongly encourage not doing this if possible. I read your list of reasons why you didn't above, but most don't seem like big problems to me -- changing our code to use "Swupd" instead of "swupd" for a section title, for example, is trivial, and uniformity like that should be encouraged, not worked around.
  - Perhaps one of the TOML libraries mentioned above has mappings that are more flexible?
- As an extension of my above comment, I'm not married to Viper by any means, and I know not everyone (&lt;cough&gt; @cmarcelo &lt;/cough&gt;) likes libraries like these, but I would strongly encourage us to at least _try_ to find a workable _configuration_ library, not just a _config-file-format parsing_ library, and then writing our own implementation of things like default values, ENV/config/flag precedence, etc.
  - That said, IMHO, none of the things you mentioned as wrong with Viper seem that bad to me. Maintaining compatibility with the soon-to-be-gone `bundle-chroot-builder.py` isn't something I think we should be concerned with. One of the motivations for rewriting BCB into mixer was to have consolidated config-parsing logic. I'd rather hide a Viper-based config behind `--new-chroots` for the time being than invest in something less full-featured just for a soon to not exist compatibility issue.

A higher level, potentially controversial thought:
I know, I know, breaking changes are awful. And we have a ton right now. But this is also a great opportunity to _rethink_ our configuration at a higher level than this. What values do we really need? What should they be called? What should this file even be called (certainly `builder.conf` doesn't make a lot of sense without a stand-alone BCB)? Should it even be one file, or several? Or can we merge several files we already have into one? What kinds of interactions do we need with these configs that a library needs to support (i.e., read-only? write from template? write from data structure? default values? aliases?)

builder/mixconfig.go
rchiossi: I don't see why those things couldn't live together. We can have a template and keep the ability of writing to the config file. The only downside that I can see now is that with the current lib, any comments on the config file would be gone once we edit it.

builder/mixconfig.go
rchiossi: @cmarcelo my original implementation (https://github.com/rchiossi/mixer-tools/commit/305d2c9c25e4484186629c13ae3c1010e851ef63) had a discrete parsing of all the values, which would make it easier to validate all the entries.

rchiossi:
> But overall this is not the direction I'd like things to go. Or at least not where I want them to end up. But as an intermediate?

This is by no means what I see as the end step for the config system, it is just a base that I can build over.

> While I understand for read/write simplicity why values currently not in builder.conf (e.g., YumTemplate) didn't get moved over to MixConfig, as a user of Builder, this feels arbitrary. (That is, some things that "feel like configuration" are in Builder.MixConfig, and others are just in Builder.) Feels like we could have a cleaner design.

That is exactly why I embedded MixConfig into Builder the way I did, so that the user can keep using `builder.YumConf` and `builder.YumTemplate`. This patch is focused on isolating everything that is exclusive to the config file. I agree that some values might be better in MixConfig instead of Builder, but I expect this refactor to happen once we are set on a foundation for the config system.

> It feels wrong to me to not use the built-in map-to-data-structure interface. It feels like Go often encourages you to reinvent the wheel, but I would very strongly encourage not doing this if possible. I read your list of reasons why you didn't above, but most don't seem like big problems to me -- changing our code to use "Swupd" instead of "swupd" for a section title, for example, is trivial, and uniformity like that should be encouraged, not worked around.

The point here is create something that does not depend on our user changing what they already have. Yes, I can change the code so that Swupd is accepted instead of swupd (or the other way around for the other sections if we were to use Viper), or for it to accept key="value" instead of key=value, but for the time being it would also mean that we need to tell everybody to update their builder.conf, I that is what I want to avoid. For this patch I wan t a drop in solution that can be easily adapted once we have more control over the other user of the conf file.

> I know, I know, breaking changes are awful. And we have a ton right now. But this is also a great opportunity to rethink our configuration at a higher level than this. What values do we really need? What should they be called? What should this file even be called (certainly builder.conf doesn't make a lot of sense without a stand-alone BCB)? Should it even be one file, or several? Or can we merge several files we already have into one? What kinds of interactions do we need with these configs that a library needs to support (i.e., read-only? write from template? write from data structure? default values? aliases?)

That is were I wanna go next, but luckly rethinking and rearranging the values is something that can be done independenly, without impacting which config system we use. In an overall, I'm 100% with you on reviewing what we need and how we use the config system,

rchiossi:
@cmarcelo @kevincwells Thanks for the review, but one thing that I'd like some feedback is on the API or rather how the config is used by the rest of the system, since for me that is the most important part of this patch. If we settle on that, it doesn't matter how we parse values or how we write them back to file since changing it would require isolated changes in this file only and not in the rest of the system.

builder/mixconfig.go
cmarcelo: I like that one better than the pairs of name / pointer to property.

cmarcelo:
Quoting myself in the review "I think that this is a good direction even if we want to take the Viper route.", you can replace Viper with TOML. We three all agree that it is a good step to improve what we have, so I say let's add it. Specially since I've (we've) been working around (and avoiding) changing that area of the code to not conflict with this change for a while.

These patches are both (a) improving the condition of the code base for changes that are orthogonal to the config file format   (b) preparing the codebase for an easier transition when we start using a different parsing library.

> What values do we really need? What should they be called? What should this file even be called (certainly builder.conf doesn't make a lot of sense without a stand-alone BCB)? Should it even be one file, or several? Or can we merge several files we already have into one?

Agreed. See also https://github.com/clearlinux/mixer-tools/issues/54#issuecomment-356373028. I just think we can do iterative improvement of the code while do that work.

Optional: I'd tackle the BCB before landing those bigger changes, so we have one less compatibility issue to deal with. Which doesn't prevent deciding what the improvements are sooner.

builder/mixconfig.go
cmarcelo: > I strongly want the ability to write the config from a data structure.

The suggestion won't prevent that. My point was more: for the initial mix, it make more sense to use the template approach. The template can take a config object and run with it, or a function can take a config object and use Fprintfs.

Note about using library facilities for writing: there will be challenges if we want to keeping user comments (and in general don't cause disruptive changes to the file), so unless library was made thinking about that case, we'll have to handle this ourselves.

cmarcelo:
@rchiossi 

Your original implementation resolves one of my comments. The other about writing still stands: we want to add commentary to the generated config, so the way you are generating right now is not very helpful. Given is straightforward change, I think is in scope for this PR.

Given I'm unlikely to have bandwidth to review this in the future, consider the write request "Optional". (I'm also lifting the Requested Changes).

rchiossi:
@cmarcelo Thanks, I'll take your suggestions into consideration and make a reviewed version of this PR.

tmarcu:
Opening new PR with new implementation.

PR #173, opened 2018-02-13
chroots: always use dnf

### LINKED ISSUE #115: Use dnf instead of yum to build chroots
See https://github.com/clearlinux/mixer-tools/pull/102#discussion_r164014488.

cmarcelo:
DNF already works in Clear and at the moment Yum is not working, so it
is a good time to stop calling it.

Drop the --packager, if an alternative version of DNF have to be used,
just put it in PATH before the system one.

Fixes #115.

### DIFF
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -35,9 +35,6 @@ type buildCmdFlags struct {
 	noPublish  bool
 	keepChroot bool
 	template   string
-
-	// Name of the program to use instead of yum (or dnf in Fedora >= 22).
-	packager string
 }
 
 var buildFlags buildCmdFlags
@@ -48,7 +45,7 @@ var buildCmd = &cobra.Command{
 	Short: "Build various pieces of OS content",
 }
 
-func buildChroots(builder *builder.Builder, packager string, signflag bool) error {
+func buildChroots(builder *builder.Builder, signflag bool) error {
 	// Create the signing and validation key/cert
 	if _, err := os.Stat(builder.Cert); os.IsNotExist(err) {
 		fmt.Println("Generating certificate for signature validation...")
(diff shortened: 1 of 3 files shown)


cmarcelo:
@tmarcu @rchiossi

mixer/cmd/build.go
```diff
@@ -266,7 +260,7 @@ func init() {
 	externalDeps[buildChrootsCmd] = []string{
 		"m4",
 		"rpm",
-		"yum",
+		"dnf",
```
rchiossi: Yum is still a dependency if you try to run without --new-chroots (not a mixer dependency, but a dependency still).
I guess this is ok since the old one will not work on new systems anyway, but it's a small inconsistency for the time being.

mixer/cmd/build.go
```diff
@@ -250,9 +247,6 @@ func init() {
 	buildChrootsCmd.Flags().BoolVar(&buildFlags.noSigning, "no-signing", false, "Do not generate a certificate to sign the Manifest.MoM")
 	buildChrootsCmd.Flags().BoolVar(&builder.UseNewChrootBuilder, "new-chroots", false, "EXPERIMENTAL: Use new implementation of build chroots")
 
-	// TODO: Remove or rewrite the message when we let yum go.
-	buildChrootsCmd.Flags().StringVar(&buildFlags.packager, "packager", "", "When using build-chroots, use passed packager program instead of yum (or dnf in Fedora >= 22)")
-
```
rchiossi: Is there any use case for a packager other than yum and dnf? You could change to dnf without removing this feature, but if this won't be used, then It is a good idea to just remove it like you did.
Should we deprecate this to avoid breaking the API for now ?

cmarcelo: There wasn't any release of mixer with this flag, so we should be fine.

mixer/cmd/build.go
cmarcelo: It seems the old one is moving to only using dnf too, so I think we are good here.

PR #174, opened 2018-02-13
Mixer Bundle Add: List bundles as separate args

kevincwells:
When 'mixer add-bundle' was first introduced, the list of bundles
being added was implemented as a single-argument comma-separated
list. It is more intuitive to specify bundles as separate arguments,
and this is the format all of the new 'mixer bundle ___' commands are
taking. (This also simplifies forthcoming mixer autocomplete code.)

This patch changes 'mixer bundle add' to support this new usage.
To make it more clear how bundles are to be specified, the "Usage"
string of help output has been updated for all commands to comply with
more standard conventions. The output now uses square brackets to
indicate optional elements, ellipses to indicate optional expansion,
and angle brackets to indicate variables as opposed to literal constants.
For example:

`Usage: mixer bundle add <bundle> [<bundle>...] [flags]`

and

`Usage: mixer bundle list [mix|local|upstream] [flags]`

### DIFF
--- a/mixer/cmd/bundles.go
+++ b/mixer/cmd/bundles.go
@@ -15,8 +15,6 @@
 package cmd
 
 import (
-	"strings"
-
 	"github.com/clearlinux/mixer-tools/builder"
 
 	"github.com/pkg/errors"
@@ -39,28 +37,29 @@ type bundleAddCmdFlags struct {
 var bundleAddFlags bundleAddCmdFlags
 
 var bundleAddCmd = &cobra.Command{
-	Use:   "add [bundle(s)]",
+	Use:   "add <bundle> [<bundle>...]",
 	Short: "Add local or upstream bundles to your mix",
 	Long: `Adds local or upstream bundles to your mix by modifying the Mix Bundle List
 (stored in the 'mixbundles' file). The Mix Bundle List is parsed, the new
 bundles are confirmed to exist and are added, duplicates are removed, and the
 resultant list is written back out in sorted order.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
-		var bundles []string
 		if !bundleAddFlags.allLocal && !bundleAddFlags.allUpstream {
 			if len(args) == 0 {
 				return errors.New("bundle add requires at least 1 argument if neither --all-local nor --all-upstream are passed")
 			}
-			bundles = strings.Split(args[0], ",")
 		}
+
 		b, err := builder.NewFromConfig(config)
 		if err != nil {
 			fail(err)
 		}
-		err = b.AddBundles(bundles, bundleAddFlags.allLocal, bundleAddFlags.allUpstream, bundleAddFlags.git)
+
+		err = b.AddBundles(args, bundleAddFlags.allLocal, bundleAddFlags.allUpstream, bundleAddFlags.git)
 		if err != nil {
 			fail(err)
 		}
+
 		return nil
 	},
 }
@@ -75,7 +74,7 @@ type bundleRemoveCmdFlags struct {
 var bundleRemoveFlags bundleRemoveCmdFlags
 
 var bundleRemoveCmd = &cobra.Command{
-	Use:   "remove [bundle(s)]",
+	Use:   "remove <bundle> [<bundle>...]",
 	Short: "Remove bundles from your mix",
 	Long: `Removes bundles from your mix by modifying the Mix Bundle List
 (stored in the 'mixbundles' file). The Mix Bundle List is parsed, the bundles
@@ -164,7 +163,7 @@ type bundleEditCmdFlags struct {
 var bundleEditFlags bundleEditCmdFlags
 
 var bundleEditCmd = &cobra.Command{
-	Use:   "edit [bundle(s)]",
+	Use:   "edit <bundle> [<bundle>...]",
 	Short: "Edit local and upstream bundles, or create new bundles",
 	Long: `Edits local and upstream bundle definition files. This command will locate the
 bundle (looking first in local-bundles, then in upstream-bundles), and launch


mixer/cmd/bundles.go
```diff
@@ -51,7 +51,13 @@ resultant list is written back out in sorted order.`,
 			if len(args) == 0 {
 				return errors.New("bundle add requires at least 1 argument if neither --all-local nor --all-upstream are passed")
 			}
-			bundles = strings.Split(args[0], ",")
+			// If just one arg passed, try splitting for compatibility
+			if len(args) == 1 {
```
matthewrsj: So I bet this is going to break with a weird error if someone does this:
```
mixer bundle add mixer,c-basic editors
```
But that is clearly user error at that point so 🚀

kevincwells: Yup. It will will try to treat `mixer,c-basic` as a single bundle name, and error out saying it doesn't exist. I decided I was fine with this.

tmarcu: Or, we can trivially solve this and not break on users missing a comma or adding a space with: 
```
        var list []string
        for _, arg := range args {
                list = append(list, strings.Split(arg, ",")...)
        }
        for i, item := range list {
                if item == "" {
                        list = append(list[:i], list[i+1:]...)
                }
        }
        bundles = list
```

tmarcu:
Unfortunately "this is a user error" doesn't fly, and you will just be re-doing this and having to reply to Jiras when a user inevitably has an itchy thumb and hits spacebar. I'd argue this isn't enough validation even, but it's better and covers what most people will enter as a list of items.

icarus-sparry:
There comes a point where too much `helping` becomes a hindrance. What happens when we want to have a space or a comma in the name? Do we switch from strings.Split() to something that knows about say backslash as an escape character? Of course we will have to allow for backslash backslash to enable us to pass on a backslash.

The unix shells don't have very good support for creating comma separated lists or even whitespace separated ones. You can see this in the special case code that is involved in handling `$( ... )` constructs. With this **ALL** trailing newlines are stripped off. This means if you want to preserve exactly the output you end up writing something like

    A="$(some command or other ; echo x)" # add a trailing x to preserve trailing newlines from some
    A=${A%x} # remove the trailing x

If you have a list, then you can convert it to a comma separated string using a similar 2 step process

    tocomma() {
            local var
            printf -v var "%q," "$@"
            printf "%s" "${var%,"}"
    }

Things are much easier if you have a comma **delimited** string. (I don't think it is a coincidence that go requires trailing commas in elements in structs whilst C does not). However this means you need to worry about strings with zero length in your go code.

So for backward compatibility I think we should attempt no more than one fixup, if there is only one argument then split it on commas, or if there is more than one, just use the args as they are. We should however add a check for a zero length string and just skip it.

kevincwells:
> @icarus-sparry: So for backward compatibility I think we should attempt no more than one fixup, if there is only one argument then split it on commas, or if there is more than one, just use the args as they are. We should however add a check for a zero length string and just skip it.

So the difference from what is in this patch and what you're suggesting is: the single-arg, comma-separated string could be `bundleA,,bundleB`, and right now it will treat that middle empty string as a bundle name and error out, but you think we should purge such strings. Did I read that right?

@tmarcu You're suggesting we support `mixer bundle add bundleA,bundleB bundleC bundleD,bundleE ...`?

mixer/cmd/bundles.go
kevincwells: My gut says that this is a solution to a problem that doesn't really exist.

icarus-sparry:
@kevincwells Actually my concern is not bundleA,,BundleB but bundleA,bundleB, (note the trailing comma), which I would like treated as 'bundleA', 'bundleB' and skip the empty string caused by the trailing comma.

This is the difference between comma **separated** values and comma **delimited** values.

kevincwells:
@icarus-sparry gotcha. Both will result in the same empty-sting-as-error scenario in this case.

I also just gave this a try, and the error you get is pretty cryptic -- when it tries to open the bundle file to parse it, it fails saying `ERROR: read /home/clr/mix/local-bundles: is a directory`. Not very intuitive at all. I will definitely add in a catch to purge such empty strings.

mixer/cmd/bundles.go
kevincwells: FYI - I'm assuming this was just meant to be representative (so apologies if you already know this). but just in case, I wanted to mention that you can't strip values from a slice the way you are here. The items returned by `range` will not be updated when you modify the slice inside the loop, and thus this code will go out of bounds.

I just included a strip-empty-strings update to this PR that does the following:
```go
bundles = strings.Split(args[0], ",")
for i := 0; i < len(bundles); i++ {
	if bundles[i] == "" {
		bundles = append(bundles[:i], bundles[i+1:]...)
		i-- // Because removal shifts things left
	}
}
```

icarus-sparry: @kevincwells 
Worth looking at https://github.com/golang/go/wiki/SliceTricks

In particular a better way to remove an item from a list if you don't care about the order is to copy the last item to the current position, set the last item to nil, so the garbage collector doesn't see a reference, and then shorten the slice by 1.

This is linear in the number of items to remove, rather than O(n^2)

kevincwells: @icarus-sparry Yeah, that's pretty slick. Worth redoing given n here is quite small?

Definitely will keep in mind for the future either way.

tmarcu: This is nice to know ^! Except for n usually being on the order of single or couple digits long I don't think we're losing that much performance here.

kevincwells:
@tmarcu @matthewrsj @icarus-sparry others

Pushed a new version based on further discussions on IRC. This drops comma support altogether, and updates the "Usage" strings to specify how arguments are actually supposed to be passed.

PR #177, opened 2018-02-14
New Command: mixer bundle validate

kevincwells:
New command checks local bundle definition files for validity as a separate step. While
'mixer bundle edit' will automatically validate bundles as you edit them, it can
be useful to manually run this validation step on bundles you have previously
written, or have edited using a different tool. Only local bundle files are
checked; upstream bundles are trusted as valid. Valid bundles yield no output.
Any invalid bundles will yield a non-zero return code.

An optional '--strictness-level' flag allows you to control how strictly bundles
are checked. All levels check basic syntax and structure.
  low:   Check file name and bundle Title match and are valid bundle names
  med:   All of 'low', plus check Description and Maintainer are non-empty
  high:  All of 'med', plus check Status and Capabilities are non-empty

Passing '--all-local' will run validation on all bundles in local-bundles.

This patch also adds tests to check that bundle header parsing is working
and that different types of validation errors are indeed being caught, and calls
the new validation command in a few new places.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -560,6 +560,13 @@ func (b *Builder) getBundleFromName(name string) (*bundle, error) {
 	if err != nil {
 		return nil, err
 	}
+	if err = validateBundle(bundle, BasicValidation); err != nil {
+		return nil, err
+	}
+	if err = validateBundleFileName(name, bundle); err != nil {
+		return nil, err
+	}
+
 	return bundle, nil
 }
 
@@ -1026,7 +1033,7 @@ editLoop:
 		// Ignore return from command; parsing below is what will reveal errors
 		_ = helpers.RunCommandInput(os.Stdin, editorCmd, path)
 
-		_, err := parseBundleFile(path)
+		err := validateBundleFile(path, BasicValidation)
 		if err == nil {
 			// Clean-up backup
 			if err = os.Remove(backup); err != nil {
@@ -1146,6 +1153,42 @@ func (b *Builder) EditBundles(bundles []string, suppressEditor bool, add bool, g
 		if err := helpers.Git("commit", "-q", "-m", commitMsg); err != nil {
 			return err
 		}
+	}
+
+	return nil
+}
+
+// ValidateLocalBundles runs bundle parsing validation on all local bundles.
+func (b *Builder) ValidateLocalBundles(lvl ValidationLevel) error {
+	files, err := ioutil.ReadDir(b.LocalBundleDir)
+	if err != nil {
+		return errors.Wrap(err, "Failed to read local-bundles")
+	}
+
+	bundles := make([]string, len(files))
+	for i, file := range files {
+		bundles[i] = file.Name()
+	}
+
+	return b.ValidateBundles(bundles, lvl)
+}
+
+// ValidateBundles runs bundle parsing validation on a list of local bundles. In
+// addition to parsing errors, errors are generated if the bundle is not found
+// in local-bundles.
+func (b *Builder) ValidateBundles(bundles []string, lvl ValidationLevel) error {
+	invalid := false
+	for _, bundle := range bundles {
+		path := filepath.Join(b.LocalBundleDir, bundle)
+
+		if err := validateBundleFile(path, lvl); err != nil {
+			invalid = true
+			fmt.Printf("Invalid: %q:\n%s\n\n", bundle, err)
+		}
+	}
+
+	if invalid {
+		return errors.New("Invalid bundles found")
 	}
 
 	return nil
--- a/builder/bundleset.go
+++ b/builder/bundleset.go
@@ -3,6 +3,7 @@ package builder
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
@@ -12,13 +13,23 @@ import (
 )
 
 var (
-	validBundleNameRegex  = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
-	validPackageNameRegex = regexp.MustCompile(`^[A-Za-z0-9-_+.]+$`)
+	validBundleNameRegex   = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
+	validPackageNameRegex  = regexp.MustCompile(`^[A-Za-z0-9-_+.]+$`)
+	bundleHeaderFieldRegex = regexp.MustCompile(`^# \[([A-Z]+)\]:\s*(.*)$`)
 )
+
+type bundleHeader struct {
+	Title        string
+	Description  string
+	Status       string
+	Capabilities string
+	Maintainer   string
+}
 
 type bundle struct {
 	Name     string
 	Filename string
+	Header   bundleHeader
 
 	DirectIncludes []string
 	DirectPackages []string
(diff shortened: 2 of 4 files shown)


builder/builder.go
```diff
@@ -1151,6 +1151,46 @@ func (b *Builder) EditBundles(bundles []string, suppressEditor bool, add bool, g
 	return nil
 }
 
+// ValidateBundles runs bundle parsing validation on a list of local bundles, or
+// all local bundles if allLocal is true. In addition to parsing errors, errors
+// are generated if the bundle is not found in local-bundles.
+func (b *Builder) ValidateBundles(bundles []string, allLocal bool) error {
```
cmarcelo: Optional: I've found the allLocal parameter here kind of odd. I'd try either one of those: (a) a separate function take lists the local bundles (and we pass that)  (b) a separate ValidateLocalBundles() that takes no arguments  (c) read the list of bundles in the call site and pass that.

builder/builder.go
```diff
@@ -1151,6 +1151,46 @@ func (b *Builder) EditBundles(bundles []string, suppressEditor bool, add bool, g
 	return nil
 }
 
+// ValidateBundles runs bundle parsing validation on a list of local bundles, or
```
cmarcelo: Optional: the ValidateBundles don't actually call our bundle validation function. These are different validations. Maybe rename one of them to avoid confusion. Context: I was initially puzzled how you could validate just the local bundles.

kevincwells: Ah, I see why you were thrown by that name. 

I had considered other names. I don't think the using "validate" in both is bad; both are validating things, they're just validating different things -- one is "validating a `bundleSet`, the other is "validating a list of bundle files". To make the name more descriptive, I considered calling this function `ValidateBundleFiles`, but 1) I wanted it to keep parallel with the other `<Action>Bundles` functions (Add, Remove, Edit), and 2) that name makes me think it should be taking filepaths. And maybe it should? But again a) I wanted to keep it parallel with the others that all take a list of bundle names, and b) that would push the path lookup to the caller (the `bundleValidateCmd.Run`), and i) that would require exporting `getBundlePath`, and ii) I favor keeping the CLI layer thin and decoupled from how bundle paths are dealt with.

Man, I had to bust out Roman Numerals to keep those lists separate. Clear sign I wrote too complicated of a response! :stuck_out_tongue_winking_eye:

builder/builder.go
kevincwells: Let me do a second pass at this logic to see if I can refactor in a way that feels better.

builder/builder.go
cmarcelo: Is the name 'check' a way to go for evaluation the bundles in isolation?

kevincwells:
I dissmissed @cmarcelo's approval review, because I added a lot on top since he gave that. That said, I followed his advice to split up the `ValidateBundles` and `ValidateLocalBundles` inside. @cmarcelo -- if you're bored or feeling generous, feel free to review again.

What's new since I first put this up:

- Took this opportunity to add in more full-fledged bundle validation.
  - This includes now parsing the previously-ignored bundle Headers
  - Introduced three levels of bundle validation strictness
  - Add checks that the file name matches the bundle title, that header values are not empty, etc.
- Added low-level validation to more places in the code
- Updated the parsing tests to validate the new Header parsing
- Added new tests to check that bundle validation errors were indeed being caught

builder/bundleset.go
```diff
@@ -114,46 +125,150 @@ func sortBundles(bundles bundleSet) ([]*bundle, error) {
...
+	var b bundle
+	var includes, packages []string
+
 	line := 0
 	for scanner.Scan() {
 		line++
 		text := scanner.Text()
 		comment := strings.Index(text, "#")
 		if comment > -1 {
+			if matches := bundleHeaderFieldRegex.FindStringSubmatch(text); len(matches) > 2 {
```
cmarcelo: Optional: I'd also only match the comment properties if the line starts with a comment. Otherwise you need think about lines like:  `package   # TITLE: name`.

builder/bundleset.go
```diff
@@ -114,46 +125,150 @@ func sortBundles(bundles bundleSet) ([]*bundle, error) {
...
 }
 
+// ValidationLevel represents a specific validation level
+type ValidationLevel int
+
+// Enum of available validation levels
+const (
+	StrictnessLow ValidationLevel = iota
+	StrictnessMed
+	StrictnessHigh
```
cmarcelo: Optional: I don't think the low/med/high are worth to have. If an extra level of check is needed, start with `--strict`, and choose some meaningful properties to be there. I think in general all those properties are optional for the pipeline, but I'm not familiar with the long term plan for those.

mixer/cmd/bundles.go
```diff
@@ -198,12 +201,74 @@ any errors are encountered earlier on.`,
...
+
+var bundleValidateCmd = &cobra.Command{
+	Use:   "validate <bundle> [<bundle>...]",
+	Short: "Validate local bundle definition files",
+	Long: `Checks local bundle definition files for validity as a separate step. While
+'mixer bundle edit' will automatically validate bundles as you edit them, it can
+be useful to manually run this validation step on bundles you have previously
+written, or have edited using a different tool. Only local bundle files are
+checked; upstream bundles are trusted as valid. Valid bundles yield no output.
+Any invalid bundles will yield a non-zero return code.
```
cmarcelo: Optional: I felt like this is a bit too much about 'edit'. I think you can shave some words by ignoring that fact (there's a balance: we want long descriptions to be as short as they can :). Describe what the command does (no need to say it is a separate step), and then mention that "mixer bundle edit" automatically calls the same checks.

cmarcelo: Optional: some user-consumable description of what is being checked would be nice.

mixer/cmd/bundles.go
```diff
@@ -198,12 +201,74 @@ any errors are encountered earlier on.`,
...
+
+An optional '--strictness-level' flag allows you to control how strictly bundles
```
cmarcelo: Optional: which strictness version does edit use?

builder/bundleset.go
```diff
@@ -12,13 +13,23 @@ import (
...
 
 var (
-	validBundleNameRegex  = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
-	validPackageNameRegex = regexp.MustCompile(`^[A-Za-z0-9-_+.]+$`)
+	validBundleNameRegex   = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
+	validPackageNameRegex  = regexp.MustCompile(`^[A-Za-z0-9-_+.]+$`)
+	bundleHeaderFieldRegex = regexp.MustCompile(`^# \[([A-Z]+)\]:\s*(.*)$`)
 )
 
+type bundleHeader struct {
```
cmarcelo: Optional: my first take with this would be just create a `map[string]string` (since we don't really care much about the actual strings at the moment to give them attributes).

builder/bundleset.go
tmarcu: I agree with @cmarcelo here, if we're vetting then go all in and make it all compliant. There's no benefit to having more simply written bundles, especially if we're ever to change the format to something else - i.e json. There will be either a working definition or an unacceptable one.

builder/builder.go
```diff
@@ -1151,6 +1151,55 @@ func (b *Builder) EditBundles(bundles []string, suppressEditor bool, add bool, g
...
+
+// ValidateBundles runs bundle parsing validation on a list of local bundles. In
+// addition to parsing errors, errors are generated if the bundle is not found
+// in local-bundles.
+func (b *Builder) ValidateBundles(bundles []string, lvl ValidationLevel) error {
+	invalid := false
+	for _, bundle := range bundles {
+		path := filepath.Join(b.LocalBundleDir, bundle)
+
+		if _, err := os.Stat(path); err != nil {
```
rchiossi: Do you need this check here? Just a bit down the line, parseBundleFile will fail if the file doesn't exist. Also, if you remove this check, you can change ValidateLocalBundles to just call ValidateBundles passing the files list.

builder/bundleset.go
kevincwells: The regex is `^# \[([A-Z]+)\]:\s*(.*)$`, so it does only match if it starts with a comment. One of my tests uses
```
# [TITLE]: realtitle
# [DESCRIPTION]: a description
# [STATUS]: 
# [CAPABILITIES]: 
# [MAINTAINER]: 
include(a)
pkg1 # [TITLE]: wrongtitle
```
To make sure that lines like that aren't checked.

builder/bundleset.go
kevincwells: I originally had two levels -- basic and strict -- where (on top of basic pars-ability) basic checked that the bundle header Title was existent, valid, and matched the bundle file name, and strict additionally checked that all of the other bundle header fields were non-empty.

The reason I split it out into a third level is that quite a few upstream bundles have a blank Capabilities header, and were thus coming back as invalid. Most, it seems. I'm not actually sure what this header was intended for (the `games` bundle, for example, has `# [CAPABILITIES]: Makes you forget about work, for a little bit`), so it probably should just be removed if it's not meaningful and most bundles don't use it. But it looked bad saying most of our bundles are invalid. 😜

builder/builder.go
kevincwells: You're right that `validateBundleFile` would eventually die out when `parseBundleFile` tries to open the file. But a) the error message you get back there would be a little cryptic, and b) I really wanted to make it explicit that what's wrong in this case is that the bundle _isn't found in local bundles_. Almost everywhere else, the tool will "automatically look up where the bundle is located, looking first in local bundles and then in upstream" or whatever, but that's not the case here. This checks _only_ local bundles (because it makes no sense for a user to check upstream bundles), and I wanted to make that clear.

Even as is, `ValidateLocalBundles` could just call `ValidateBundles`, but (ignoring the redundant local check you're suggesting I remove) it'd mean looping through the list of bundles a second time (once to form the `[]string` out of the `[]FileInfo` in `ValidateLocalBundles`, and then again to call `validateBundleFile` down in `ValidateBundles`). I considered doing this, but figured `validateBundleFile` was enough of an abstraction/re-use.

builder/bundleset.go
cmarcelo: Ah, haven't noticed the `^`. I'd still add the check, but indeed it is not important for correctness.

builder/bundleset.go
matthewrsj: My view of a strict vs basic is basic should just check if they are valid at all (have all the *required* fields). Strict can then check the remaining headers. In my opinion it is ok to fail strict validation on bundles with empty optional headers. I don't see any issues with upstream bundles failing strict validation if they pass basic validation, since basic is the one that matters.

I agree with @cmarcelo and @tmarcu that three levels is too much. Users will need to know too much about bundle format and what is optional/required/semi-optional to be able to realistically pick the level they want to validate at.

builder/builder.go
rchiossi: I don't like the idea that if you change one function you mostly likely gonna have to remember to change the other. You can check the error type to have something less cryptic if that is the issue. This would save a stat operation and any cryptic TOCTOU error that could eventually show up.

builder/bundleset.go
kevincwells: Gotcha. Like I said, that's how I originally implemented it too. Happy to go back, as long as people are happy to have upstream bundles fail strict.

builder/builder.go
kevincwells: I checked what the error output is, and it was less cryptic than I had thought -- you basically just get a "does not exist" message:
```
clr@machine /home/clr/mix # mixer bundle validate foobar
Invalid: "foodbar":
open /home/clr/mix/local-bundles/foobar: no such file or directory
```

As a result, I pulled this out and had `ValidateLocalBundles` just call directly into this like you suggested.

mixer/cmd/bundles.go
kevincwells: Fixed in new push.

mixer/cmd/bundles.go
kevincwells: Fixed in new push.

builder/bundleset.go
kevincwells: Fixed in new push.

kevincwells:
I just pushed a new version that I think addresses people's concerns. The basic changes are:

- There are now just "basic" and "strict" validation.
- The long help message describes what is actually checked in these a little more clearly
- `ValidateLocalBundles` now calls `ValidateBundles`
- Added in a missing filename == bundle header Title check to the `getBundleByName` function that gets called throughout the `mixer bundle _____` commands. Meant to have that, just forgot. It was already doing this check for `validate` and `edit`.

builder/bundleset.go
kevincwells: Input taken, but I chose to leave it this way, for a couple reasons:

1. For now we have a fixed, consistent header, but it could change. I think it's good to have the programatic definition of that strictly defined in one place, as opposed to implicitly defined by what map keys get looked up in different places in the code. 
1. This makes it easier to write code that uses it. `b.Header.Fake` is a compile-time error, while `b.Header['fake']` is a runtime error.
1. At some point we'll likely switch to using TOML (or JSON) for bundle definition files. Parsing libraries for those usually require a type defined for it to map to. While the format might not end up exactly like what I have here (it might not be in a nested `Header` object), this gets us a head start.

builder/bundleset.go
matthewrsj: > as long as people are happy to have upstream bundles fail strict.

Yep, I have no issues with this. We can either ignore it or take it as an AR to update the bundle headers.

PR #182, opened 2018-02-15
Allow controlling number of workers via command line

cmarcelo:


cmarcelo:
@matthewrsj @tmarcu I'll add another like this for chroots later.

matthewrsj:
@cmarcelo I'll look at this later, but please do not do this for chroot creation. Big changes coming there...

cmarcelo:
> @cmarcelo I'll look at this later, but please do not do this for chroot creation. Big changes coming there...

I've done, but will not add to this PR then.

My intent was to at least get the --new-chroots to be as fast as old BCB. Then it is easy validate it and phase out BCB. Adding goroutines did not changed the innards of chroot generation, but I'll wait your patch and decide whether I rebase or drop the '--chroots-workers'.

matthewrsj:
@cmarcelo my changes are pretty invasive. You can see them at https://github.com/matthewrsj/mixer-tools/blob/bundle-improvements/builder/chroots.go

If it isn't going to conflict too much feel free to push. I do change the internals quite a bit though.

cmarcelo:
From a quick glance there will be simple to solve conflicts in the text. I'm not changing the operations that are done, just making sure the code in the loop runs inside a goroutine. I'm still running a test here to see if it will be enough, I'll follow up.

cmarcelo:
Note that this PR (as-is) is suitable for review / being applied.

cmarcelo:
At least on my machine, with about 20 bundles, the change get us on par with BCB (slightly better actually). So I'll push the `--chroot-workers` in a separate PR.

cmarcelo:
Mixer changes are only adding the flag and filling the data to swupd.

matthewrsj:
Please resolve conflicts.

cmarcelo:
Done.

mixer/cmd/build.go
```diff
@@ -260,6 +266,7 @@ func init() {
 	}
 
 	buildCmd.PersistentFlags().IntVar(&buildFlags.numFullfileWorkers, "fullfile-workers", 0, "Number of parallel workers when creating fullfiles, 0 means number of CPUs")
+	buildCmd.PersistentFlags().IntVar(&buildFlags.numDeltaWorkers, "delta-workers", 0, "Number of parallel workers when creating deltas, 0 means number of CPUs")
```
tmarcu: You mean number of extra CPUs?

swupd/delta.go
```diff
@@ -35,8 +35,8 @@ type Delta struct {
 // CreateDeltas creates all delta files between the previous and current version of the
 // supplied manifest. Returns a list of deltas (which contains information about
 // individual delta errors). Returns error (and no deltas) if it can't assemble the delta
-// list.
-func CreateDeltas(manifest, statedir string, from, to uint32) ([]Delta, error) {
+// list. If number of workers is zero or less, 1 worker is used.
```
tmarcu: Would be good to add to the description of the cmd line arg, so it's clear what default is.

swupd/delta.go
```diff
@@ -67,15 +67,15 @@ func createDeltasFromManifests(c *config, oldManifest, newManifest *Manifest) ([
 		return []Delta{}, nil
 	}
 
-	// TODO: Unify these maxRoutines constants, get a better default (maybe NumCPUs)
-	// and make it configurable.
-	const maxRoutines = 3
+	if numWorkers < 1 {
```
tmarcu: If you change the default to 1 instead of 0 you don't need this check anymore.

cmarcelo: Not sure how can I avoid this check. People can still pass 0 to the function when calling it.

swupd/delta.go
cmarcelo: Note the default for the package is different than what we expose in the command line. Usually library shouldn't be making assumptions about how much resources to use, that is caller code job.

matthewrsj:
@cmarcelo we have conflicts again.

tmarcu:
Sorry @cmarcelo

PR #183, opened 2018-02-15
Only unset file type when deprecating rename-from files

matthewrsj:
Unsetting the rename-to file type causes a file with no flags set (e.g.
....).  Do not unset the file type when deprecating a rename-to file,
(e.g. F..r). Only unset the file type when deprecating a rename-from
file (e.g. .d.r).

### DIFF
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -542,7 +542,9 @@ func deprecateOldRenames(oldRenames []*File) {
 	for _, f := range oldRenames {
 		if f.DeltaPeer == nil {
 			f.Rename = false
-			f.Type = TypeUnset
+			if !f.Present() {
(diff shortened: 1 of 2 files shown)


swupd/manifest.go
```diff
@@ -542,7 +542,9 @@ func deprecateOldRenames(oldRenames []*File) {
 	for _, f := range oldRenames {
 		if f.DeltaPeer == nil {
 			f.Rename = false
-			f.Type = TypeUnset
+			if !f.Present() {
```
icarus-sparry: Told you we should not set the type to TypeUnset.

matthewrsj: yeah yeah...

matthewrsj: 😭

PR #184, opened 2018-02-15
mixer: fix output of list when have included bundles

cmarcelo:
Account for the size of included bundles for formatting the output of
list.

Before:

```
bootloader     (upstream)
c-basic        (upstream)
desktop        (upstream)
desktop-apps   (upstream) (included)
desktop-assets (upstream) (included)
desktop-gnomelibs (upstream) (included)
desktop-locales (upstream) (included)
kernel-native  (upstream)
libX11client   (upstream) (included)
os-core        (upstream)
os-core-update (upstream)
python3-basic  (upstream) (included)
sysadmin-basic (upstream) (included)
```

After:

```
bootloader        (upstream)
c-basic           (upstream)
desktop           (upstream)
desktop-apps      (upstream) (included)
desktop-assets    (upstream) (included)
desktop-gnomelibs (upstream) (included)
desktop-locales   (upstream) (included)
kernel-native     (upstream)
libX11client      (upstream) (included)
os-core           (upstream)
os-core-update    (upstream)
python3-basic     (upstream) (included)
sysadmin-basic    (upstream) (included)
```

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -33,6 +33,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"text/tabwriter"
 
 	"github.com/clearlinux/mixer-tools/helpers"
 	"github.com/clearlinux/mixer-tools/swupd"
@@ -904,14 +905,6 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
 		bundles = upstreamBundles
 	}
 
-	// Used for pretty-printing lists
-	var max int
-	for _, bundle := range bundles {
-		if len(bundle.Name) > max {
-			max = len(bundle.Name)
-		}
-	}
-
 	// Create the full, parsed set
 	set, err := b.getFullBundleSet(bundles)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


kevincwells:
Hmm... good catch, but I"m not sure this is the right solution.

This has the unintended side effect of making `mixer bundle list local` print oddly. E.g.:

```
clr@machine /home/clr/mix # mixer bundle edit desktop
clr@machine /home/clr/mix # mixer bundle list local
desktop           (in mix) (masking upstream)
```

The reason is because, `mixer bundle list` is a list of _only_ the "available local bundles", and not a list of what those bundles themselves include. This patch uses the full list of the included bundles to set the width, and hence the above big gap (because it's formatting based on the length of `desktop-gnomelibs`).

kevincwells:
FYI - with the code as it was (i.e., without this patch), only `mixer bundle list` is affected, because it's the only one of the three that is displaying included content too

cmarcelo:
Yeah, I've overlooked that extra space. I'll try something else and we see if looks better.

cmarcelo:
Could you take another look?

kevincwells:
`tabwriter` -- cool library. Glad to know about it!

This is nice and clean. :+1:

PR #187, opened 2018-02-16
Add `--chroot-workers` to build command

cmarcelo:
By default, do the same as BCB and use NumCPUs. After this commit our `--new-chroots` should be on par with BCB.

This PR is for the last commit. The others are part of #182.


cmarcelo:
@matthewrsj I rather get his before further changes to chroot building logic. I don't think this will cause much issue for you to rebase, but let me know if that's not the case.

Issue #145, opened 2018-02-06
Fail when the user is trying to create a mix that breaks format bumps


cmarcelo:
This is a first step towards #59: first we identify the undesired situation and do not generate a potentially invalid update. This is an easier task than providing auto format bumps, and identifying the case is necessarily anyway. Optionally add a flag to allow forcing create the mix even in the invalid scenario.

This only applies to mix creation, so regular Clear Linux workflow shouldn't be affected (no mix version in that workflow).

tmarcu:
I just want to make sure your PR doesn't ignore the mixversion, as upstream workflow still uses it, but it happens to be that .mixversion = .clearversion.

cmarcelo:
I thought that upstream workflow used mixversion == "" and only set clearversion.

cmarcelo:
I think the "versions update" command (#188) will be sufficient to close this. It does prevent unwanted bumps until we support them.

tmarcu:
Check merged into mastre.

Issue #59, opened 2017-12-26
Handle creating a mix that touches or skips a format bump in Clear Linux


cmarcelo:
Mixer should be aware about the format bumps in the upstream (Clear Linux), since right now it might be producing a stream that is not usable (skipping a bump in upstream, leading to some invalid state).

"Handle" here might refer first to "be aware and fail if the input is invalid" and then to "create the intermediate mixes as needed", so might make sense implement this in steps. Just the warning would go great length to avoid issues.

See also #30 description that have a concrete example. That PR added more information with the intent of making life easy to identify this bumps.

For the record @mdhorn implemented this outside mixer for our test mix.

cmarcelo:
@tmarcu I know this was in your Roadmap, added to GitHub so we have a way to keep track (and comment) on this, since we are doing the format bump work from outside the mixer at the moment. Any updates on this?

kevincwells:
@tmarcu -- wanted to make an observation about how we might implement this, with regards to when we attempt to deal with format bumps and how we expose this to the user.

I was talking with @cmarcelo about steering the CLI away from "what tasks does the tool need to do" and more toward "what goals does the user want to accomplish". With that in mind, we run into the need for auto format bumps when the user wants to update the upstream version their mix is based on, and that version happens to be one or more formats above the old one. As such, I would recommend attempting to do the auto-bump at the time the user changes their upstream CLR version, rather than as a reaction to a new CLR version at a later event (like a build or a deliberate format bump).

This has the nice benefit of tying the format bump activities to the user's action that actually causes their need, as well as simplifying the already complicated output of a later build. (Imagine explaining to the user "You wanted to do a build, but at some other point you changed something else, so we tried to do these other builds you didn't ask for because you didn't know you needed them, and somewhere it all went wrong!" Something along the lines of "Warning: Updating to upstream version 19350 requires additional format bump builds to keep up with changes to upstream." is way easier to understand.)

Right now it's not ~~possible~~ easy to detect "the time the user changes their upstream CLR version", because it's just the `.clearversion` file; this is another reason why a command like `mixer config set UpstreamVersion 19350` (or whatever it ends up looking like) is useful.

Not sure what your thoughts on this were already, but wanted to put that out there.

cmarcelo:
Proposal on how we get this moving forward:

- Start with #145 so we stop producing "invalid" mixes.
[...]

Issue #122, opened 2018-01-30
Add "file format" information to Manifest files (which is different than the OS content format)


cmarcelo:
To make life easier when we ever evolve the manifest files, would be good to have a way to describe the version of the file format we are using.

I think in the past this was conflated with the "format" for the OS (as it implied what version of swupd was used and so on). But as we create mixes, this correspondence is gone.

See also the description https://github.com/clearlinux/swupd-server/wiki/Manifest-structure
```
Each line in the header has 2 fields; the first is a keyword, and the second is a value. Accepted keywords are:

    MANIFEST: The first line of manifest (the "magic"). Second field indicates the format number used to create the manifest.
```

Note we also probably need to fix our manifest parsing to treat MANIFEST (or whatever we come up) like the description above.

cmarcelo:
@icarus-sparry do you have thoughts on this?

icarus-sparry:
Yes, I think the idea is very sound.

I would implement it as a list of fields that the "body" of the manifest has, rather than as a magic version number. This makes the data essentially self describing, and so allows us to add new fields without breaking older readers of the data (they should be programmed to ignore unknown fields). This list of fields would go in the header.

I would also allow additional header lines, again unknown ones are ignored (or probably better generate warnings).

tmarcu:
@icarus-sparry @cmarcelo can you please describe this a bit more, do you suggest chaning MANIFEST: to FORMAT: ? Or adding more fields? I'm closing irrelevant issues and trying to add more detail for ones we can complete in Q3.

icarus-sparry:
The basic suggestion is that the header of a MANIFEST should look something like
```
MANIFEST
version:	10
previous:	0
filecount:	1
timestamp:	1451936779
contentsize:	17929151
fields: flags hash version filename
```

then we have a blank line, and then we have 4 tab delimited fields per line. If we decide to add the size of the compressed file and the size of the uncompressed file to avoid the DOS that we are currently vulnerable to, then the `fields` line would become `fields: flags dsize size hash version filename".

Parsing code can just zip through the lines which make up the body. If an older version of swupd doesn't support the dsize field it can just ignore the dsize field.

One thing that needs to be checked carefully is that new fields are added in a compatible manner.

matthewrsj:
That's a good idea

cmarcelo:
CC: @otaviobp 

[...]

PR #188, opened 2018-02-16
mixer: add 'update-versions' command

cmarcelo:
Replace the workflow of manually updating the two files to get a new
version.

It does check the upstream format and do not cross format bumps. So it
helps preventing people generating wrong content.

Updates #145.
Updates #59.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -187,8 +187,14 @@ func (b *Builder) initDirs() error {
 
 // Get latest CLR version
 func (b *Builder) getLatestUpstreamVersion() (string, error) {
+	return b.DownloadFileFromUpstream("/latest")
+}
+
+// DownloadFileFromUpstream will download a file from the Upstream URL
+// joined with the passed subpath. It will trim spaces from the result.
+func (b *Builder) DownloadFileFromUpstream(subpath string) (string, error) {
 	// Build the URL
-	end, err := url.Parse("/latest")
+	end, err := url.Parse(subpath)
 	if err != nil {
 		return "", err
 	}
@@ -197,14 +203,20 @@ func (b *Builder) getLatestUpstreamVersion() (string, error) {
 		return "", err
 	}
 
+	resolved := base.ResolveReference(end).String()
 	// Fetch the version and parse it
-	resp, err := http.Get(base.ResolveReference(end).String())
+	resp, err := http.Get(resolved)
 	if err != nil {
 		return "", err
 	}
 	defer func() {
 		_ = resp.Body.Close()
 	}()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("got status %q when downloading: %s", resp.Status, resolved)
+	}
+
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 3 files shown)


cmarcelo:
@kevincwells could you take a look?

cmarcelo:
I'm not strongly committed to any of the names (the command, the flags, defaults), only the features: "do a bump".

I know we have a flag after update to perform the change in version, but I never use it, I've always opt to explicitly increment the versions in a separate step.

mixer/cmd/versions.go
```diff
@@ -0,0 +1,141 @@
...
+	upstreamVersion uint32
+	set             bool
+}
+
+func init() {
+	RootCmd.AddCommand(updateVersionsCmd)
+	updateVersionsCmd.Flags().StringVarP(&config, "config", "c", "", "Builder config to use")
+	updateVersionsCmd.Flags().Uint32Var(&updateVersionsFlags.mixVersion, "mix-version", 0, "Next mix version (0 means increment current one)")
+	updateVersionsCmd.Flags().Uint32Var(&updateVersionsFlags.upstreamVersion, "upstream-version", 0, "Next upstream version (0 means take the latest possible)")
+	updateVersionsCmd.Flags().Uint32Var(&updateVersionsFlags.upstreamVersion, "clear-version", 0, "Alias for --upstream-version")
```
tmarcu: I think we should choose one or the other since this is new anyway.

cmarcelo: I prefer `--upstream-version`, but since init uses `--clear-version`, I thought would be nice to alias it. But happy to remove either of the lines.

kevincwells: Haha, `init` is only still on `--clear-version` 'cause people (rightfully) asked me to not add even more changes to my earlier patches

mixer/cmd/versions.go
```diff
@@ -0,0 +1,141 @@
...
+	updateVersionsCmd.Flags().BoolVarP(&updateVersionsFlags.set, "set", "s", false, "Set the new versions in the mix")
+}
+
+func runUpdateVersions(cmd *cobra.Command, args []string) {
```
kevincwells: Our convention (which I support) thus far has been to keep the CLI code quite thin, and have the "guts" of the logic elsewhere. Historically that's all been in `builder.go`, which is now massive, but either way I'd prefer it not be here.

builder/builder.go
```diff
@@ -178,8 +178,14 @@ func (b *Builder) initDirs() error {
 
 // Get latest CLR version
 func (b *Builder) getLatestUpstreamVersion() (string, error) {
+	return b.DownloadFileFromUpstream("/latest")
+}
+
+// DownloadFileFromUpstream will download a file from the Upstream URL
+// joined with the passed subpath. It will trim spaces from the result.
+func (b *Builder) DownloadFileFromUpstream(subpath string) (string, error) {
```
kevincwells: Nice generalization :+1:

mixer/cmd/versions.go
```diff
@@ -0,0 +1,141 @@
...
+	b, err := builder.NewFromConfig(config)
+	if err != nil {
+		fail(err)
+	}
+
+	nextMix := updateVersionsFlags.mixVersion
+	nextUpstream := updateVersionsFlags.upstreamVersion
+
+	currentMix, err := parseUint32(b.MixVer)
```
kevincwells: I like this, and I think it'd be good for us to normalize how we parse the versions (when it's needed as a numeric value instead of a string) throughout the code. Not sure how many such places actually exist now (the auto-increment comes to mind, but that will likely be rewritten to utilize what's in this, however it ends up landing), but they should be converted.

mixer/cmd/versions.go
tmarcu: I don't like inconsistency... so sadly lets keep --clear-version if that is what the other commands also use. We can batch change them then.

mixer/cmd/versions.go
```diff
@@ -0,0 +1,141 @@
...
+
+var updateVersionsCmd = &cobra.Command{
+	Use:   "update-versions",
+	Short: "Update the mix and upstream versions",
+	Long: `Choose the next mix and upstream versions based.
+
+Without any flags, it increments mix by 10 and upstream to the latest
+version of the same format used by current upstream. Use --mix-version
+and --upstream-version to set a specific version. By default the
+versions are shown but not set, to set use --set (-s).
```
kevincwells: YMMV, but it feels more standard to me to make it set by default, and support a `--dry-run` flag for when people want to test.

mixer/cmd/versions.go
```diff
@@ -0,0 +1,141 @@
...
+var updateVersionsFlags struct {
+	mixVersion      uint32
+	upstreamVersion uint32
+	set             bool
+}
+
+func init() {
+	RootCmd.AddCommand(updateVersionsCmd)
+	updateVersionsCmd.Flags().StringVarP(&config, "config", "c", "", "Builder config to use")
+	updateVersionsCmd.Flags().Uint32Var(&updateVersionsFlags.mixVersion, "mix-version", 0, "Next mix version (0 means increment current one)")
```
kevincwells: Would prefer "increment current value by 10"; if I were a newcomer that hadn't paid attention to CLR's versioning, I would interpret "increment" as `+= 1`

mixer/cmd/versions.go
```diff
@@ -0,0 +1,141 @@
...
+		failf("Couldn't parse latest upstream version: %s", err)
+	}
+
+	switch {
+	case nextUpstream == 0:
+		nextUpstream = latestUpstream
+	case nextUpstream < currentUpstream:
+		failf("Next upstream version (%d) must be at least the current upstream version (%d)", nextUpstream, currentUpstream)
+	case nextUpstream > latestUpstream:
+		failf("Can't update upstream to %d. The latest upstream version in format %s is %d.", nextUpstream, upstreamFormat, latestUpstream)
```
kevincwells: For now this is exactly the right thing to do. In the future (post auto-format-bumps), this is when we'd instead say `"Can't update upstream to %d without performing a format bump. The latest upstream version in format %s is %d.\nWould you like to apply a format bump now? [y/N]:"`

mixer/cmd/versions.go
cmarcelo: OK!

kevincwells: Can we leave this as-is and just drop the alias when we change the other flag?

For the record, a quick scan says there's only that one other place the flag name exists.

cmarcelo: For consistency: I could add the alias to init as well... It pairs nice there with update-url.

cmarcelo:
The goal of the command in the long-run is to support checks like you mentioned. I agree with the ideas of extra checks, but will let you and others add those :-)

> I don't mind at all the idea of a master/hybrid command like this for setting both, I just would prefer it be easy to set one at at time if you want.

Walking mix version only is "harmless". Walking upstream version only could be dangerous, because it might make you override an older version without noticing. So, another check that could be useful is whether the mix version in question was already created.

mixer/cmd/versions.go
cmarcelo: I'm trying to keep the flag juggling in the CLI, and pushing the bulk of the command into builder.go.

mixer/cmd/versions.go
cmarcelo: I'm parsing the versions in ReadVersions (this took parsing code from a few places).

mixer/cmd/versions.go
cmarcelo: Yep. I even wrote a version with --dry-run but changed my mind with people "accidentaly" changing versions.

With the new scheme, I think that is less of an issue.

mixer/cmd/versions.go
cmarcelo: Fixed the help to reflect that (and explain why 10), also giving the user an --increment flag.

mixer/cmd/versions.go
cmarcelo: That's the spirit, or even give a flag that tells the user it can bump regardless.

Note that one problem raised in previous discussions was: we would need to make sure metadata we generate for the "last version of a format" is compatible to the swupd running there. I think we are in a better position now to guarantee that (c.f. Icarus comments about why is good to generate manifests via templates) via mixer itself. See also https://github.com/clearlinux/mixer-tools/issues/122.

builder/builder.go
```diff
@@ -1916,3 +1928,99 @@ func writeMetaFiles(path, format, version string) error {
...
+
+	if nextUpstream < b.UpstreamVersion {
+		return fmt.Errorf("invalid upstream version to update (%d), need to be at least the same as current upstream version (%d)", nextUpstream, b.UpstreamVersion)
+	}
+
+	switch {
+	case nextUpstream == 0:
+		nextUpstream = latest
+
+	case nextUpstream < b.UpstreamVersion:
```
kevincwells: Perhaps I'm missing something, but is this not the same check as L1980 above?

mixer/cmd/versions.go
```diff
@@ -0,0 +1,122 @@
...
+	b, err := builder.NewFromConfig(config)
+	if err != nil {
+		fail(err)
+	}
+
+	var nextMix uint32
+	if versionsUpdateFlags.mixVersion > 0 {
+		nextMix = versionsUpdateFlags.mixVersion
+	} else {
+		nextMix = b.MixVersion + versionsUpdateFlags.increment
```
kevincwells: I like this approach to controlling how much to increment.

builder/builder.go
```diff
@@ -1916,3 +1928,99 @@ func writeMetaFiles(path, format, version string) error {
...
+		// TODO: Could we relax this rule? Use case: something is broken in a mix because
+		// upstream change, then next mix version can rollback to a previous good upstream
+		// version. If so, check against the "first" in the same format instead.
```
kevincwells: Another potential use case is: what if you update your upstream version, _do nothing else_, and decide you want to update it back before you even make a mix?

builder/builder.go
```diff
@@ -81,6 +81,10 @@ type Builder struct {
 	NumFullfileWorkers int
 	NumDeltaWorkers    int
 	NumChrootWorkers   int
+
+	// Parsed versions.
+	MixVersion      uint32
+	UpstreamVersion uint32
```
kevincwells: I like what you're doing in this commit, I just don't like the naming convention.  `MixVer` == string while `MixVersion` == uint32 will definitely cause confusion later on.  If we want to preserve both (which is fine), perhaps `MixVer` and `MixVerInt` or `MixVerUint32`?

cmarcelo: Chose MixVerUint32 and UpstreamVerUint32. I prefer not touch the existing variable for this change.

Please consider revamping those names later: MixVersion and MixVersionStr seems good names for the uint and string variables.

builder/builder.go
cmarcelo: That TODO has your name on it.

builder/builder.go
cmarcelo: What if I started with if then switched to switch?

Fixed.

kevincwells:
I only checked how you changed to format range, so assuming you didn't throw anything else sneaky in there, LGTM

cmarcelo:
Ping

tmarcu:
Merged.

PR #189, opened 2018-02-16
swupd: return information about fullfile creation

cmarcelo:
Fix an old TODO, this gives visibility in every build about the
compression algorithms being used.

Note the accounting "gzip" (and not "external-gzip") because at the
moment it is being used for links and directories. It probably should
just migrate to use the external gzip later.

```
Output looks like:

=> CREATE FULLFILES
- Already created: 0
- Not compressed:  0
- Compressed
  - external-bzip2       40
  - gzip                 389
  - external-xz          2520
  - external-gzip        953
Total fullfiles: 3902
```


cmarcelo:
This is mostly swupd side change, the mixer is only printing information it gets as part of `build update`.

PR #191, opened 2018-02-16
mixer: add a default value for mix version in init

### LINKED ISSUE #123: Provide better defaults to "mixer init"
At the moment, calling mix init with no arguments I get:

```
$ mixer init
Error: Required flag(s) "clear-version", "mix-version" have/has not been set
Usage:
  mixer init [flags]

Flags:
      --all                   Initialize mix with all upstream bundles automatically included
      --clear-version int     Supply the Clear version to compose the mix from (default 1)
      --config string         Supply a specific builder.conf to use for mixing
  -h, --help                  help for init
      --local-rpms            Create and configure local RPMs directories
      --mix-version int       Supply the Mix version to build
      --upstream-url string   Supply an upstream URL to use for mixing (default "https://download.clearlinux.org")

Global Flags:
      --new-swupd   EXPERIMENTAL: Use new implementation of swupd-server when possible
```

Things that could be improved:

- There is a "default 1" for --clear-version, which is not possible because it is a required argument. The version "1" also is not good when passed to mixer itself (it would fail).
- We could default the clear-version to be the latest published release at the moment (or the current version of Clear Linux the user is running).
- We could default the mix-version to be just 10 (or even 1).

Note that one side effect is that "mixer init" would just work without arguments, so the other facilities it provides should be easy to get some other way.

cmarcelo:
And print the versions being used.

Fixes #123. (in combination with #153 that was already merged)

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -238,6 +238,8 @@ func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allU
 		upstreamVer = ver
 	}
 
+	fmt.Printf("Initializing mix version %s from upstream version %s\n", mixVer, upstreamVer)
+
 	// Deprecate '.clearversion' --> 'upstreamversion'
 	if _, err := os.Stat(filepath.Join(b.VersionDir, ".clearversion")); err == nil {
 		b.UpstreamVerFile = ".clearversion"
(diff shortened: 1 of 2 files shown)


builder/builder.go
```diff
@@ -238,6 +238,8 @@ func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allU
 		upstreamVer = ver
 	}
 
+	fmt.Printf("Initializing mix version %s from upstream version %s\n", mixVer, upstreamVer)
```
kevincwells: Any reason this is down here instead of at the top of the function?

cmarcelo: Wanted to print the actual version instead of latest.

kevincwells: :+1:

PR #192, opened 2018-02-17
Improve progress output for chroot building

matthewrsj:
In particular add a "done" message to help with progress indication
during parallelized chroot builds.

### DIFF
--- a/builder/chroots.go
+++ b/builder/chroots.go
@@ -256,7 +256,7 @@ src=%s
 		return errors.Wrap(err, "couldn't fix os-release file")
 	}
 
-	fmt.Println("Creating chroots for bundles")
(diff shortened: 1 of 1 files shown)


builder/chroots.go
```diff
@@ -256,7 +256,7 @@ src=%s
 		return errors.Wrap(err, "couldn't fix os-release file")
 	}
 
-	fmt.Println("Creating chroots for bundles")
+	fmt.Printf("Creating chroots for %d bundles\n", len(set))
```
cmarcelo: Since os-core is already done, maybe you want len(set)-1.

cmarcelo: Just throw this print down below near the for loop that feed the channel, so this will be clearer.

PR #196, opened 2018-02-18
Perform Lstat instead of Stat on rename files

matthewrsj:
We want information about the link, not the linked file, when doing a
stat on files in the update. The File.Info field holds FileInfo about
the file, not the linked file.

Doing a 'Stat' instead of an 'Lstat' caused an issue where a link was
changed in an update but the target file was not. The target file
therefore did not exist in the full chroot for that update. Performing a
'Stat' caused a panic because the target file could not be found.

### DIFF
--- a/swupd/rename.go
+++ b/swupd/rename.go
@@ -188,7 +188,7 @@ func fixupStatFields(needed []*File, m *Manifest, c *config) error {
 			bundleChroot = filepath.Join(c.imageBase, fmt.Sprint(m.Header.Previous), "full")
 		}
 		path := filepath.Join(bundleChroot, needed[i].Name)
-		fi, err := os.Stat(path)
+		fi, err := os.Lstat(path)
 		if err != nil {
 			return err
 		}


swupd/rename.go
```diff
@@ -188,7 +188,7 @@ func fixupStatFields(needed []*File, m *Manifest, c *config) error {
 			bundleChroot = filepath.Join(c.imageBase, fmt.Sprint(m.Header.Previous), "full")
 		}
 		path := filepath.Join(bundleChroot, needed[i].Name)
-		fi, err := os.Stat(path)
+		fi, err := os.Lstat(path)
```
icarus-sparry: I suspect that we need more logic around this than just swapping Stat to Lstat. My thought is that it doesn't make sense to rename symbolic links, the chances are that they are too small to be worth renaming.

matthewrsj: I agree, but to do that we have to persist the type flag for deleted files. The reason this link is sticking around and not cleaned up by trimRenamed is because it is deleted at that version and the file type is lost.

This is a quick fix for now to fix my testing.

Issue #77, opened 2018-01-05
Usability: Implement basic autocomplete using Cobra


kevincwells:
# Proposal
Cobra has built-in functionality to generate both bash and zsh autocomplete configuration for mixer, including all commands and flags. Using this functionality will significantly improve mixer's usability.

There are a number of ways to do this, so design details will likely be filled in when the issue is taken up.

**Note** Cobra also supports advanced, contextual auto-complete for specific functions (e.g. typing `mixer bundle add edi[tab][tab]` autocompletes to `editors`), but those will need to be implemented on a command-by-command basis.

kevincwells:
# Implementation Options
I looked into how some other Cobra apps do this, so I wrote this up mostly as a way to organize my thoughts and pass my learnings on to whoever picks this up (in case it's not me).

## Background
Autocompletion for commandline tools works by sourcing a configuration file that describes the autocomplete options available for a command. Bash and zsh each have their own format for how and where to put these configuration files.

Cobra's autocomplete generation works via the [GenBashCompletion](https://godoc.org/github.com/spf13/cobra#Command.GenBashCompletion) and [GenBashCompletionFile](https://godoc.org/github.com/spf13/cobra#Command.GenBashCompletionFile) commands (and their zsh counterparts). Essentially you create another runnable command, and the business logic of that command calls one of these functions on your apps RootCmd.

## Generating the autocomplete file
### Command in Mixer
This option would involve adding a new command to mixer itself to generate the autocomplete. This is the option taken by [Hugo](https://gohugo.io/commands/hugo_gen_autocomplete/) and [Kubernetes](https://kubernetes.io/docs/tasks/tools/install-kubectl/#enabling-shell-autocompletion).
### Standalone Util
Alternatively, a separate executable could be made in the same package as mixer -- say, `mixer_autocomplete` -- that's sole purpose is to import the root command and generate its autocomplete file. The pros to this is there isn't a command hanging off mixer for autocomplete. The con to this is you have a separate executable that gets built (though obviously doesn't have to be installed)

## What we do with the autocomplete file
### "Installing" it
Hugo takes the approach of writing directly to `/etc/bash_completion.d` (though they let you specify a different path for the completion file). This isn't done automatically (i.e., at hugo install time); you need to manually run `sudo hugo gen completion`, and you must run it with `sudo` due to writing to `/etc`

[...]

PR #198, opened 2018-02-21
Add mixer bash and zsh completion

kevincwells:
Please see #77 (as well as the commit message below) for info on why I implemented this as a separate executable, and what the alternatives were.

Please also note that, as addressed in #77, this only implements _basic_ completion, which includes all of mixer's subcommands and their flags (and known valid arguments for some commands, like `mixer bundle list`). More sophisticated completion (e.g., `mixer bundle add ed[TAB] --> mixer bundle add editors`) is also possible, but will be added on a command-by-command basis.

----

This patch introduces a new `mixer-completion` executable that generates `bash`
and `zsh` completion files for `mixer`. The executable is built via `make`, but not
installed via `make install`. Instead, `make install` runs the executable,
which writes the `bash` and `zsh` completion files to `/etc/bash_completion.d/` and
`/usr/share/zsh/site-functions/`, respectively.

This patch uses Cobra's built-in `bash`/`zsh` completion generation functionality.
Some Cobra applications choose to have this completion-file generation as a
permanent command off the root command (e.g., `kubectl completion bash`). This
patch makes this a stand-alone executable to avoid cluttering the main `mixer`
tool with this functionality. Many Cobra applications also leave creating/
installing these completion files up to the user as a manual, secondary step.
This patch completes these tasks at mixer install time.

### DIFF
--- a/mixer-completion/cmd/completion.go
+++ b/mixer-completion/cmd/completion.go
@@ -1,0 +1,68 @@
+// Copyright © 2018 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"errors"
+	"os"
+
+	mixerCmd "github.com/clearlinux/mixer-tools/mixer/cmd"
+	"github.com/spf13/cobra"
+)
+
+// CompletionCmd represents the base command for mixer-completion
+var CompletionCmd = &cobra.Command{
+	Use:   "mixer-completion",
+	Short: "Generate autocomplete files for mixer",
+	Long: `Generates autocomplete files for mixer. If no arguments are passed, or if
+'bash' is passed, a bash completion file is written to /etc/bash_completion.d/.
+If 'zsh' is passed, a zsh completion file is written to
+/usr/share/zsh/site-functions/.
+
+These files allow you to type, for example:
+  mixer bun[TAB] -> mixer bundle a[TAB] -> mixer bundle add
+
+Note that this command must be run as root.`,
+	Args:      cobra.OnlyValidArgs,
+	ValidArgs: []string{"bash", "zsh"},
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) > 1 {
+			return errors.New("mixer-completion takes at most one argument")
+		}
+
+		var shell string
+		if len(args) > 0 {
+			shell = args[0]
+		} else {
+			shell = "bash"
+		}
+
+		var err error
+		switch shell {
+		case "bash":
+			if err = os.MkdirAll("/usr/share/bash-completion/completions", 0755); err != nil {
+				return err
+			}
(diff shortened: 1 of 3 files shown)


mixer-completion/cmd/completion.go
```diff
@@ -0,0 +1,68 @@
...
+		if len(args) > 0 {
+			shell = args[0]
+		} else {
+			shell = "bash"
+		}
+
+		var err error
+		switch shell {
+		case "bash":
+			if err = os.MkdirAll("/etc/bash_completion.d", 0755); err != nil {
```
kevincwells: I went with `0755` here (and below), which is what this directory is on my host system. Anyone have objection?

icarus-sparry: If this is something that we install then clear seems to use

     /usr/share/bash-completion/completions

rchiossi:
It would be nice to be able to generate those files in the user home as well. Maybe as an extra arg?

mixer-completion/cmd/completion.go
kevincwells: I considered putting it there, but I read a number of places (like [this one](https://lists.fedoraproject.org/pipermail/devel/2011-June/153012.html) about Fedora from 2011) that claim `/usr/share/bash-completion/` is where the bash-completion package itself installs some completions (and symlinks to `/etc/bash_completion.d/` to enable/disable), and it is this second location where completions are actually loaded from, and where custom completions are to be installed from packages.

It's entirely possible that's no longer true, or that it was never true for Clear.  The vast majority of places just tell you to put things in `/etc/bash_completion.d/` and don't mention anywhere else.  I did not try putting things in `/usr/share/bash-completion/`, so I don't know if it would pick them up from there, but it does successfully pick them up from `/etc/bash_completion.d/`

kevincwells:
> It would be nice to be able to generate those files in the user home as well. Maybe as an extra arg?

Are you saying instead of `/etc/bash_completions.d/`, or in addition to? I know some people implement custom completions by, say, writing this file to `~/.bash/completions/` and then `source`ing it in their `.bashrc`. But I don't really see a need for supporting that use case if the tool can just set it up automatically at install time.  It would be pretty easy to allow an override path of where to write the file as a flag if we want to support that at some time. Right now I (deliberately) set this up so that the `mixer-completions` executable is not installed, so obviously we wouldn't support additional things like that unless we planned to install it as well.

rchiossi:
What I had in mind as a use case was local development. That way I could have completion without installing the dev version into the system. In that case I would have access to mixer-completions, so not installing wouldn't be an issue. This is a nice-to-have feature, but it is by no means a blocker to this PR.

mixer-completion/cmd/completion.go
icarus-sparry: The point is that for packaging purposes and because we are stateless, we *never* ship anything in /etc. see packages/bash/stateless.patch in the clear source tree for example where we pick up /usr/share/defaults/etc/bash.bashrc rather than /etc/bash.bashrc.

mixer-completion/cmd/completion.go
```diff
@@ -0,0 +1,68 @@
...
+				return err
+			}
+			err = mixerCmd.RootCmd.GenBashCompletionFile("/etc/bash_completion.d/mixer")
```
icarus-sparry: We need this path changed.

kevincwells: Gotcha. And you said above that, on CLR, the correct path would be `/usr/share/bash-completion/completions/mixer`, correct?

kevincwells:
Updated the path for bash completion per @icarus-sparry's feedback, and tested that it still works.

kevincwells:
> @tmarcu: I don't think 60 lines is a lot of clutter. Is there any other specific reason this is a separate executable instead of just 1 command added to the current tool?

It's not about cluttering the code, it's about cluttering the tool. I didn't want to have `completion` show up as one of the commands when you do `mixer -h`, for example.

tmarcu:
@kevincwells ok fair enough reason.

PR #201, opened 2018-02-22
Fix manifest 'previous' header for minversions

matthewrsj:
The 'previous' header for a minversion manifest should be set to itself
instead of the usual previous manifest.

Also fix test error wording.

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -147,6 +147,7 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 		if ver == 0 {
 			ver = ui.lastVersion
 		}
+		bundle.Header.Previous = ver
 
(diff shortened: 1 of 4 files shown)


phmccarty:
From what I can tell, this is not how swupd-server behaves -- previous version is not treated specially.

Consider https://download.clearlinux.org/update/20450/Manifest.MoM

What is the reason behind this change? By "minversion manifest", do you mean the special case where "minversion" is set to the same version you are building?

matthewrsj:
@phmccarty aaahhh, ok. I was trying to avoid the case in my testing where previous was "0" and was basing my expectations off of https://download.clearlinux.org/update/300/Manifest.MoM (so, you know, ancient history).

matthewrsj:
Yeah, current behavior is expected.

PR #207, opened 2018-02-28
 Add a default value for LOCAL_BUNDLE_DIR

kevincwells:
`LOCAL_BUNDLE_DIR` is a new `builder.conf` field introduced in mixer 4.0. It is included in the auto-generated conf from `mixer init`, and it is referenced in the updated documentation (tbd).

Previously, however, it was marked as a "required" flag in the `ReadBuilderConf` function. This meant that existing users, who did not already have this value, were given an error message saying it was
missing, but no useful information about how to fix this.

This patch provides default fall-back value of the user's `PWD + "/local-bundles"` if the field is missing. If this happens, a warning message is output, informing the user to update their conf.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -356,7 +356,7 @@ func (b *Builder) ReadBuilderConf() error {
 		{`^CERT\s*=\s*`, &b.Cert, true},
 		{`^CLEARVER\s*=\s*`, &b.UpstreamVer, false},
 		{`^FORMAT\s*=\s*`, &b.Format, true},
-		{`^LOCAL_BUNDLE_DIR\s*=\s*`, &b.LocalBundleDir, true},
+		{`^LOCAL_BUNDLE_DIR\s*=\s*`, &b.LocalBundleDir, false},
 		{`^MIXVER\s*=\s*`, &b.MixVer, false},
 		{`^LOCAL_REPO_DIR\s*=\s*`, &b.RepoDir, false},
 		{`^LOCAL_RPM_DIR\s*=\s*`, &b.RPMDir, false},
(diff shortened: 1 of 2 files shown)


kevincwells:
PS - I am open to removing the fall-back PWD value, and instead still making it a mandatory field, but just providing the more-helpful error message. This would be achieved by leaving the "required" field as `false`, so the regular terse error message is skipped, and making the "WARNING" become a fatal error.

Note that while all of this config parsing and default value code goes away with @rchiossi's forthcoming config patch, we will _still_ have the dilemma of "what's a good default value" for a field like this.

rchiossi:
Providing a default looks fine, but I'm not sure you need that warning message. The correct behavior to me would be to either warn and fail for all values or silent use the default for all values. Treating one particular value in a different way seems to just complicate the code and possibly confuse the user.

kevincwells:
> Providing a default looks fine

Where this gets tricky is for new fields like this one and existing users. Yes, that's the entire point of default values -- to "still work" even if the user doesn't have a field defined. My point is just that, the value we choose for the default needs to a) make sense, and b) be communicated to the user somehow (or at least documented). Defaulting to the working directory seems reasonable to me, but might not to someone who uses their `builder.conf` to point to all sorts of different locations and isn't expecting stuff to be written to the working dir as a result.

> Treating one particular value in a different way seems to just complicate the code and possibly confuse the user.

I agree. It's not great to special case values. I added the message because just seeing `buildconf missing entry for variable: LOCAL_BUNDLE_DIR` doesn't actually help an existing user know what to do to fix the problem. What does `LOCAL_BUNDLE_DIR` mean? What are they supposed to set it _to_?

To that point, though, the same is true for basically any value in `builder.conf`.  Perhaps the better solution would be to just amend the current output to say "Please see mixer documentation for how to use builder.conf", and maybe even point to the URL or something (don't like that idea, 'cause URLs change).

rchiossi:
> Defaulting to the working directory seems reasonable to me, but might not to someone who uses their builder.conf to point to all sorts of different locations and isn't expecting stuff to be written to the working dir as a result.

This other person should not expect to be running fine with the defaults. Using the active directory as default is more than resonable.

> Defaulting to the working directory seems reasonable to me, but might not to someone who uses their builder.conf to point to all sorts of different locations and isn't expecting stuff to be written to the working dir as a result.

I don't think we have to go that far. As long as we have a decent explanation of all those values in the documentation, just the mention of that missing value should be enough for the user to find the solution. maybe as a future addition we could have a man page for mixer :)

PR #208, opened 2018-02-28
Add an "offline" mode for mixer

kevincwells:
This patch adds a persistent `--offline` flag to the top-level mixer
command, which means every command can pass this flag to make mixer
work offline.

When mixer works offline, it skips caching upstream bundle definition
files. This has the implication that every bundle in the user's mix
must be available in `local-bundles`.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -51,6 +51,10 @@ var UseNewSwupdServer = false
 // building the chroots. This is an experimental feature.
 var UseNewChrootBuilder = false
 
+// Offline controls whether mixer attempts to automatically cache upstream
+// bundles. In offline mode, all necessary bundles must exist in local-bundles.
+var Offline = false
+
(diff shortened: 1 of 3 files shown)


kevincwells:
This is an uncommon use case, but one that dev-ops-style users would benefit from.

I've tested this, and it seems to work. The "look first in local-bundles" nature of mixer's design means this directory being empty isn't ever a problem, so long as the bundles exist in `local-bundles` The only case where it mattered was with `mixer bundle list`, which pulls upstream bundles to do all kinds of checking for nice output. It was easy to catch that, though, and simply return an empty `bundleSet`.

`mixer init --offline` is a bit tricky, because `init` attempts to add 4 bundles right off. Not a problem if you've already copied all your bundles for your build (including those) over into `local-bundles`. The average user might not understand to do that, but this really is a niche feature the average user wouldn't use.

PR #210, opened 2018-02-28
Fix Mix Bundle List file bug

kevincwells:
Fixes a bug in the location for where the Mix Bundle List
(stored in the 'mixbundles' file) is read/write during 'mixer
bundle add' and 'mixer bundle remove' commands.

### DIFF
--- a/mixer-completion/cmd/completion.go
+++ b/mixer-completion/cmd/completion.go
@@ -29,7 +29,6 @@ type completionCmdFlags struct {
 
 var completionFlags completionCmdFlags
 
-
 // CompletionCmd represents the base command for mixer-completion
 var CompletionCmd = &cobra.Command{
 	Use:   "mixer-completion",
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -602,7 +602,7 @@ func (b *Builder) getBundleFromName(name string) (*bundle, error) {
 func (b *Builder) getMixBundlesListAsSet() (bundleSet, error) {
 	set := make(bundleSet)
 
-	bundles, err := helpers.ReadFileAndSplit(b.MixBundlesFile)
+	bundles, err := helpers.ReadFileAndSplit(filepath.Join(b.VersionDir, b.MixBundlesFile))
 	if os.IsNotExist(err) {
 		return set, nil
 	} else if err != nil {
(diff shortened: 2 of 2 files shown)


mixer-completion/cmd/completion.go
```diff
@@ -29,7 +29,6 @@ type completionCmdFlags struct {
 
 var completionFlags completionCmdFlags
 
-
```
matthewrsj: unrelate change?

kevincwells: Correct. Left over from a un-linted commit earlier. (I think mine)

kevincwells: (AKA, what's on Master right now doesn't pass linting :P)

matthewrsj: but we lint in travis

kevincwells:
This is a "major bug", but one that doesn't manifest in visible errors for most users.

1. If user uses their working dir as their `VERSIONS_DIR`, this bug does not affect them.
1. If the user points `VERSIONS_DIR` somewhere else, this bug will make a `mixbundles` file in the correct place, but then _re_-make and use a different `mixbundles` file in the directory where they're running mixer. So it "appears to work" for the user, it's just reading and writing to the wrong file.

The only real way this would manifest as an error for the user is if the tool doesn't have write-access to wherever it is being run (because the user expects all input/output to be in other locations configured in `builder.conf`)

mixer-completion/cmd/completion.go
kevincwells: If I'm understanding correctly, this is left over from my hasty `--path` addition to mixer-completion. It's just a stray newline that linting wants removed.

kevincwells: As in, I believe I forgot to `make lint`, and because travis takes 8 years to run, we merged in a hurry to get the release out.

matthewrsj: > travis takes 8 years to run, we merged in a hurry

:disappointed:

tmarcu: Huh, weird..I don't merge things that haven't passed Travis... I think it passed and then you had to force push an update which maybe broke linting? And we didn't wait for that one to pass again?

kevincwells: @tmarcu -- probably. I'm 99% sure it was a mistake I made either way. Luckily it's just a bonus newline.

kevincwells:
@tmarcu Agreed, it'd be good to carefully look at how and where we are configuring things, and think through how/where we _should_ be doing it instead. I think changes would likely go in as part of @rchiossi's config.

Off the top of my head, I can say that the only thing I *know* is not configurable is the `upstream-bundles` directory, which is currently always made in the working directory. This was chosen because they really should be treated as temporary working files, but we definitely could let you configure it right alongside `LOCAL_BUNDLE_DIR` -- an `UPSTREAM_BUNDLE_DIR`.  As I mentioned elsewhere, my only real concern with doing that is that people might attempt to configure this to a common space that multiple concurrent builds point to (this is, in theory, something you could easily do with `local-bundles`). The tool clears out any other cached versions not corresponding to the current build, so concurrent builds with different `upstreamversion`'s writing to the same dir would battle each other. Easy enough to document that people shouldn't use it that way.

Other than that, `upstreamversion`, `upstreamurl`, `mixversion`, and `mixbundles` are all created in the dir defined by `VERSIONS_DIR`. Likely that ought to be renamed to something else, and we might want to split up what all goes in there.

PR #211, opened 2018-03-02
builder.go: Hardlink only with --keep-chroots

tmarcu:
Due to using nosync, hardlinking could start before all of the individual
bundle chroots are actually removed. However, we do not need to hardlink if
--keep-chroots is not passed, because the full chroot implicitly will not
have any duplicate files. In this case, just delete the individual bundle
chroots and don't hardlink.


builder/builder.go
```diff
@@ -1483,14 +1483,15 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string,
...
-	// Hardlink the duplicate files. This helps when keeping the bundle chroots.
-	hardlinkcmd := exec.Command("hardlink", "-f", b.StateDir+"/image/"+b.MixVer+"/")
-	hardlinkcmd.Stdout = os.Stdout
-	hardlinkcmd.Stderr = os.Stderr
-	err = hardlinkcmd.Run()
-	if err != nil {
-		return errors.Wrapf(err, "couldn't perform hardlink step")
+	} else {
+		// Hardlink the duplicate files ONLY when keeping the bundle chroots.
+		hardlinkcmd := exec.Command("hardlink", "-f", b.StateDir+"/image/"+b.MixVer+"/")
```
matthewrsj: Would prefer `filepath.Join()`

PR #212, opened 2018-03-02
Mixer bundle basic validation change

kevincwells:
Previously, "basic" bundle validation included checking that the
bundle filename (currently used as the "bundle name") matched the
"Title" field in the bundle header (currently not used), and that
the "Title" field itself was valid. This was a step toward future
usage, where mixer will use the header "Title" itself.

This caused mixer to be incompatible with previous versions of CLR,
for which these filename-to-title missmatches were allowed and thus
present.

This patch moves this checking exclusively to "--strict" validation,
and removes bundle content validation from "basic" altogether. Basic
validation now only checks that the bundle filename is valid and that
the bundle contents syntax can be parsed.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -604,10 +604,7 @@ func (b *Builder) getBundleFromName(name string) (*bundle, error) {
 	if err != nil {
 		return nil, err
 	}
-	if err = validateBundle(bundle, BasicValidation); err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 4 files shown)


kevincwells:
@matthewrsj you're welcome :stuck_out_tongue:

kevincwells:
Aw dang, lemme update the command help text :(

PR #214, opened 2018-03-02
Update 'mixer bundle add' to skip existing bundles

### LINKED ISSUE #213: Repeated calls to 'mixer bundle add'
If a user calls `mixer bundle add <bundle>` multiple times, mixer still outputs `Adding bundle <bundle> from upstream bundles` every time. To improve clarity to users and better mirror the output of other commands, this output should instead report something like `Bundle <bundle> already found in mix; skipping`

kevincwells:
Fixes #213 
------
Previously, repeated calls to `mixer bundle add <bundle>` would yield
the same `Adding bundle <bundle> from <source>` output. Because the
Mix Bundle List is treated as a set, and duplicate bundles are removed,
these redundant additions do not change the resultant list, and are
effectively no-ops.

This patch:
a) Explictly skips the bundles if they already exist.
b) Reports this case to the user

As a result, functionality is equivalent, but the no-op behavior is
more explicit and output is more clear.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -738,6 +738,11 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool, 
 
 	// Add the ones passed in to the set
 	for _, bName := range bundles {
+		if _, exists := set[bName]; exists {
+			fmt.Printf("Bundle %s already in mix; skipping\n", bName)
+			continue
+		}
+
 		bundle, err := b.getBundleFromName(bName)
 		if err != nil {
 			return err
@@ -758,6 +763,11 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool, 
 		}
 
 		for _, bundle := range localSet {
+			if _, exists := set[bundle.Name]; exists {
+				fmt.Printf("Bundle %s already in mix; skipping\n", bundle.Name)
+				continue
+			}
+
 			set[bundle.Name] = bundle
 			fmt.Printf("Adding bundle %s from local bundles\n", bundle.Name)
 		}
@@ -772,6 +782,11 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool, 
 		}
 
 		for _, bundle := range upstreamSet {
+			if _, exists := set[bundle.Name]; exists {
+				fmt.Printf("Bundle %s already in mix; skipping\n", bundle.Name)
+				continue
+			}
+
 			set[bundle.Name] = bundle
 			fmt.Printf("Adding bundle %s from upstream bundles\n", bundle.Name)
 		}


builder/builder.go
```diff
@@ -738,14 +738,19 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool,
 
 	// Add the ones passed in to the set
 	for _, bName := range bundles {
+		if _, exists := set[bName]; exists {
```
matthewrsj: optional: I prefer the `if _, ok := set[bName]; ok {` idiom represented in all the go docs.

builder/builder.go
```diff
@@ -754,12 +759,17 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool,
 	if allLocal {
 		localSet, err := b.getDirBundlesListAsSet(b.LocalBundleDir)
 		if err != nil {
-			return errors.Wrapf(err, "Failed to read local bundles dir: %s", b.LocalBundleDir)
+			return errors.Wrapf(err, "Failed to read local bundles dir: %q", b.LocalBundleDir)
 		}
 
 		for _, bundle := range localSet {
+			if _, exists := set[bundle.Name]; exists {
```
matthewrsj: optional `ok`

builder/builder.go
```diff
@@ -768,12 +778,17 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool,
 		upstreamBundleDir := getUpstreamBundlesPath(b.UpstreamVer)
 		upstreamSet, err := b.getDirBundlesListAsSet(upstreamBundleDir)
 		if err != nil {
-			return errors.Wrapf(err, "Failed to read upstream bundles dir: %s", upstreamBundleDir)
+			return errors.Wrapf(err, "Failed to read upstream bundles dir: %q", upstreamBundleDir)
```
matthewrsj: these `%s`->`%q` changes don't seem related

builder/builder.go
kevincwells: I thought the `ok` convention was used when the boolean returned represented success or failure, not existence. But I'm not purporting to be an expert at that at all.

kevincwells: https://blog.golang.org/go-maps-in-action

Yeah, looks like you're right. That's.... um... huh. I guess that's a convention I don't agree with, but am willing to follow.

builder/builder.go
kevincwells: That's fair. In general I want to normalize output (and have a separate patch I'm doing right now that changes over the output for the rest of the commands to be consistent with each other), but I can make these changes there if you'd prefer.

matthewrsj: I prefer commits to be a single functional or logical change. This is a rule I often mess up myself, but there you go... At least for small commits like this it is easy to manage.

builder/builder.go
kevincwells: Looks like I've thought this for a while and have several other places in the code I used "exists" like this. Hmm...

builder/builder.go
kevincwells: I totally agree, I just often get caught up in trying to reduce future work at the expense of breaking this rule. So here, I knew I wanted to convert them all to `%q` in my other patch, so lazily tried to get it over with while I was editing that section.

But again, totally agree, and it's something I need to be way stricter about.

kevincwells:
>I'm not approving because I'm not entirely sure I like this (I'm not disapproving though, so @tmarcu can approve if he wishes). It seems a little verbose - I don't really care if mixer is skipping, I just want it to exist in my bundle list at the end. Who cares if this is a noop?

IMHO, it's not a critical issue by any means, but I think this is better than it is today, because
1. It's more representative of what's really going on. (At minimum, @bktan8 found the current behavior unclear.)
1. It's more consistent with the `mixer bundle remove` command. (Trying to `remove` a bundle that isn't there yields a `Bundle 'foo' not found in mix; skipping` output.)

As we discussed elsewhere, I am also in favor of a mode of operation which does not output messages like this at all (be it through a `--verbose` or `--silent` flag), but while the tool is outputting messages, I think this commit makes those messages more clear.

builder/builder.go
kevincwells: Took out all these changes. :)

builder/builder.go
kevincwells: Leaving this in for now 'cause that's how I did it everywhere else in the code. Can clean up later all together.

PR #215, opened 2018-03-05
Build bundle information using dnf commands

matthewrsj:
* Build bundle information using dnf commands
  Deprecate use of chroot-building for each individual bundle. This allows
us to scale much easier on machines with limited disk space. Because
chroots are decoupled from manifest creation this also makes it possible
in the future to move to doing bundle builds on distributed machines and
allows our bundle numbers to scale. The bundle file list is stored in a
JSON file that is read later by the manifest creation step.

  This leaves a fallback in swupd/fullchroot.go to allow users to add
content to a chroot in order to extend a typical build.

* Overhaul tests to use new bundle creation method
  This overhaul leaves several tests using the old method of using chroots
to generate the update to verify we aren't totally deprecating the old
method.

Two other small fixes.

~~**NOTE**: I'm sure there will be changes that need to be made to this, just opening it up for discussion now so we can talk about the general direction. I'll keep working on cleaning it up and ask for reviews when I think it's ready for a full review.~~

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -1362,6 +1362,14 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
 		}
 	}
 
+	// Generate the certificate needed for signing verification if it does not exist
+	if signflag == false && template != nil {
+		err := helpers.GenerateCertificate(b.Cert, template, template, &privkey.PublicKey, privkey)
+		if err != nil {
+			return err
+		}
+	}
+
 	if UseNewChrootBuilder {
 		// Get the set of bundles for which to build chroots
 		set, err := b.getFullMixBundleSet()
@@ -1405,30 +1413,22 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
 		if err != nil {
 			return err
 		}
-	}
 
-	// Generate the certificate needed for signing verification if it does not exist and insert it into the chroot
-	if signflag == false && template != nil {
-		err := helpers.GenerateCertificate(b.Cert, template, template, &privkey.PublicKey, privkey)
-		if err != nil {
-			return err
-		}
-	}
-
-	// Only copy the certificate into the mix if it exists
-	if _, err := os.Stat(b.Cert); err == nil {
-		certdir := b.StateDir + "/image/" + b.MixVer + "/os-core-update/usr/share/clear/update-ca"
-		err = os.MkdirAll(certdir, 0755)
-		if err != nil {
-			helpers.PrintError(err)
-			return err
-		}
-		chrootcert := certdir + "/Swupd_Root.pem"
-		fmt.Println("Copying Certificate into chroot...")
-		err = helpers.CopyFile(chrootcert, b.Cert)
-		if err != nil {
-			helpers.PrintError(err)
-			return err
+		// Only copy the certificate into the mix if it exists
+		if _, err := os.Stat(b.Cert); err == nil {
+			certdir := filepath.Join(b.StateDir, "image", b.MixVer, "os-core-update/usr/share/clear/update-ca")
+			err = os.MkdirAll(certdir, 0755)
+			if err != nil {
+				helpers.PrintError(err)
+				return err
+			}
+			chrootcert := filepath.Join(certdir, "/Swupd_Root.pem")
+			fmt.Println("Copying Certificate into chroot...")
+			err = helpers.CopyFile(chrootcert, b.Cert)
+			if err != nil {
+				helpers.PrintError(err)
+				return err
+			}
 		}
 	}
 
--- a/builder/bundleset.go
+++ b/builder/bundleset.go
@@ -32,8 +32,10 @@ type bundle struct {
 	Header   bundleHeader
 
 	DirectIncludes []string
-	DirectPackages []string
+	DirectPackages map[string]bool
 	AllPackages    []string
+
+	Files map[string]bool
 }
 
 type bundleSet map[string]*bundle
@@ -54,7 +56,11 @@ func validateAndFillBundleSet(bundles bundleSet) error {
 	}
 	for _, b := range sortedBundles {
 		var allPackages []string
-		allPackages = append(allPackages, b.DirectPackages...)
+		for k, v := range b.DirectPackages {
+			if v {
+				allPackages = append(allPackages, k)
+			}
+		}
 		for _, include := range b.DirectIncludes {
 			allPackages = append(allPackages, bundles[include].AllPackages...)
 		}
@@ -297,7 +303,10 @@ func parseBundle(contents []byte) (*bundle, error) {
 	}
 
 	b.DirectIncludes = includes
-	b.DirectPackages = packages
+	b.DirectPackages = make(map[string]bool)
+	for _, p := range packages {
+		b.DirectPackages[p] = true
+	}
 
 	return &b, nil
 }
--- a/builder/chroots.go
+++ b/builder/chroots.go
@@ -3,11 +3,13 @@ package builder
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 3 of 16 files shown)


swupd/bundleinfo.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+type bundleInfo struct {
```
kevincwells: This looks nearly exactly like the `bundle` type. Why do we have both?

matthewrsj: Because I didn't want to import the builder package just to get the bundle struct... I suppose I could move both definitions to helpers? or some internal package not yet defined?

builder/bundleset.go
```diff
@@ -32,8 +32,10 @@ type bundle struct {
 	Header   bundleHeader
 
 	DirectIncludes []string
-	DirectPackages []string
+	DirectPackages map[string]bool
 	AllPackages    []string
+
+	Files map[string]bool
```
kevincwells: Is this actually used? I'm probably just missing it, but all references to `Files` I see in this patch are to the `bundleinfo` type or the `Manifest` type.

matthewrsj: Uh, yes. This is probably the most important data structure of this entire PR. Look in chroots.go. This is where all the files are stored (instead of there being actual chroots). This exact struct then gets written to a JSON file.

kevincwells: :wink: 'twas a case of GitHub helpfully hiding the entire `chroots.go` file

builder/bundleset.go
```diff
@@ -297,7 +301,11 @@ func parseBundle(contents []byte) (*bundle, error) {
 	}
 
 	b.DirectIncludes = includes
-	b.DirectPackages = packages
+	b.Files = make(map[string]bool)
```
kevincwells: It looks like this also gets init'd at L196 of `chroots.go` (aka, first line of `resolveFilesForBundle`). What do you think about leaving it nil here (i.e., don't use memory until we need it)? Or not re-init'ing it in `resolveFilesForBundle`.  I suggested leaving it nil first, though, just 'cause none of the `mixer bundle` commands use this field. Guess it's a pretty small amount of memory, though :)

swupd/bundleinfo.go
kevincwells: That's a pretty good reason. And [The Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) argues we shouldn't refactor it yet.

[The Go Blog](https://blog.golang.org/package-names) argues against us having a `helper` package altogether, and for pretty good reason.

Probably (someday) the right thing to do is to move the purely "bundle" related types and methods (the stuff that currently lives in `bundleset.go`, mostly) into a non-exported/internal `bundle` package that both builder and swupd import.

As I've said before, Mixer could use an overall organizational code refactor.

builder/chroots.go
```diff
@@ -80,6 +82,416 @@ func readBuildChrootsConfig(path string) (*buildChrootsConfig, error) {
...
+		}
+	}
+}
+
+func isBannedPath(path string) bool {
+	if path == "/" {
+		return true
+	}
+
+	bannedPaths := []string{
```
kevincwells: Since this doesn't change, should this get pulled up so it's not reinitialized every time `isBannedPath` is called (which, if I understand correctly, is for every file in the build)

matthewrsj: Absolutely, I'll do that. Good call.

builder/chroots.go
```diff
@@ -80,6 +82,416 @@ func readBuildChrootsConfig(path string) (*buildChrootsConfig, error) {
...
+	if strings.HasPrefix(path, "/usr/lib64/") {
+		return path
+	}
+	re = regexp.MustCompile(`^(/lib/)`)
+	return re.ReplaceAllLiteralString(path, "/usr/lib64/")
+}
+
+func addFileAndPath(destination map[string]bool, absPathsToFiles ...string) {
+	for _, file := range absPathsToFiles {
+		if isBannedPath(file) {
```
kevincwells: Is there any chance we want to know (i.e., output/log of some kind) when this case hits? (i.e., would the user want to know that a file is being skipped?)

builder/chroots.go
```diff
@@ -80,6 +82,416 @@ func readBuildChrootsConfig(path string) (*buildChrootsConfig, error) {
...
+	return nil
+}
+
+func resolveFiles(numWorkers int, set bundleSet, packagerCmd []string) error {
+	var err error
+	var wg sync.WaitGroup
+	fmt.Printf("Resolving files using %d workers\n", numWorkers)
+	wg.Add(numWorkers)
+	bundleCh := make(chan *bundle)
+	errorCh := make(chan error, numWorkers)
```
icarus-sparry: I think this code needs much more in the way of explanation. As far as I can see the ideas are very confused. You start up N workers, which keep reading work from the input channel, if they have any error then they write to the error channel, and the parent stops writing to the input chan.

So why does the worker try to read more work after it has sent an error? In particular this can lead to deadlock if you get many errors.

Because you have the error channel buffered this will potentially leave worker routines hanging around as the worker can send an error response and go back to reading the next task, the driver can choose at random if it should read something from the error channel or write something to the task channel, so the worker which has just detected a problem could (at random) start up new work, which can then error.

If the error channel say gets to 90% full, and then the driver exhausts the set and so calls wg.Done, and the remaining workers report more than another 10% of errors, then the workers will hang waiting for their error to be accepted into the error channel, and the driver will hang waiting for all the workers to exit.

When the worker detects an error send it down the error channel and break out of the range and hence call wg.Done rather than trying to create potentially more problems.

This ensures that each worker sends no more than one error response down the channel, and so the error channel is never unable to take all the messages and so the workers never block.

You should drain all of the errorCh, not just take either 0 or 1 value from it after you call wg.Wait.

builder/chroots.go
```diff
@@ -80,6 +82,416 @@ func readBuildChrootsConfig(path string) (*buildChrootsConfig, error) {
...
+		err = <-errorCh
+	}
+	return err
+}
+
+func parseNoopInstall(installOut string) []string {
+	pkgDeps := []string{}
+	inBlock := false
+	re := regexp.MustCompile("[[:space:]]([[:graph:]]*)[[:space:]]")
+	for _, line := range strings.Split(installOut, "\n") {
```
icarus-sparry: Why not split the installOut string into 3 parts first, at the "Installing dependencies" and the first blank line, and then process the lines in the middle part?

Using a regular expression to split the line apart is a bit heavyweight!

If you want to keep using a regular expression, compile it only (at most) once per run of mixer, not once per call to parseNoopInstall.

builder/chroots.go
```diff
@@ -80,6 +82,416 @@ func readBuildChrootsConfig(path string) (*buildChrootsConfig, error) {
...
+			}
+			fmt.Printf("... done with %s\n", bundle.Name)
+		}
+		wg.Done()
+	}
+	for i := 0; i < numWorkers; i++ {
+		go packageWorker()
+	}
+
+	for _, bundle := range set {
```
icarus-sparry: See comments about almost identical code around line 220.

builder/chroots.go
```diff
@@ -80,6 +82,416 @@ func readBuildChrootsConfig(path string) (*buildChrootsConfig, error) {
...
+		if !inBlock {
+			if !strings.Contains(line, "Installing dependencies:") {
+				continue
+			} else {
+				inBlock = true
+			}
+		}
+
+		// at this point we are inBlock of packages list to be installed
+		// next line break is the end of the packages list
```
icarus-sparry: I would like an example of what installOut should look like in the comments. It would make it a lot simpler to see if the code was doing the right thing, rather than seeing what the code is doing and hoping that if it is doing something that seems reasonable then it is correct. It also protects against future changes in dnf output.

builder/bundleset.go
matthewrsj: Fair. I think I was debugging the panic I got from not initing it at all and I accidentally put it in two places.

builder/chroots.go
matthewrsj: No, it would really spam if we reported every time we hit a banned path. I don't think we need to tell the user some things are getting skipped that have always been skipped by swupd.

builder/chroots.go
```diff
@@ -80,6 +82,416 @@ func readBuildChrootsConfig(path string) (*buildChrootsConfig, error) {
...
+		}
+	}
+	return false
+}
+
+// Clear Linux OS ships several symlinks in the filesystem package, and RPM
+// packages may install files to paths with one of these symlinks as a path
+// prefix. Short term, resolve file paths using the symlinks from the
+// filesystem. Long term, we need to check every path component and resolve.
+func resolveFileName(path string) string {
```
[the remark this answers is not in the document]
matthewrsj: Pushed a new implementation. I'm not sure I like it but let me know what you think. It is certainly faster than what I had before.

matthewrsj:
By the way, after this PR is reviewed I'll change `chroots.go` and other related `chroots` references to `bundles.go` and `bundles` references. I'm holding off on doing that for now so you can review this without *everything* being a change.

builder/bundleset.go
matthewrsj: done.

builder/chroots.go
matthewrsj: done.

builder/chroots.go
matthewrsj: done.

builder/chroots.go
matthewrsj: done.

validate-swupd.sh
```diff
@@ -131,7 +131,7 @@ for i in $(ls update/www); do
 
 	set -x
 	echo $next | sudo tee $(pwd)/update/www/version/format$3/latest
-	swupd update --path os-$i-install -u file://$(pwd)/update/www -C $(pwd)/Swupd_Root.pem
+	swupd update --path os-$i-install -F $3 -u file://$(pwd)/update/www -C $(pwd)/Swupd_Root.pem
```
icarus-sparry: If you are changing the line already, then please put double quotes around everything that should not be broken apart on whitespace, i.e.

    swupd update --path "os-$i-install" -F "$3" -u "file://$(pwd)/update/www" -C "$(pwd)/Swupd_Root.pem"

Also might be worth using `"$PWD"` rather than `"$(pwd)"` to avoid the need to start up a new shell.

builder/builder.go
```diff
@@ -1405,30 +1413,22 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
-		}
-		chrootcert := certdir + "/Swupd_Root.pem"
-		fmt.Println("Copying Certificate into chroot...")
-		err = helpers.CopyFile(chrootcert, b.Cert)
-		if err != nil {
-			helpers.PrintError(err)
-			return err
+		// Only copy the certificate into the mix if it exists
+		if _, err := os.Stat(b.Cert); err == nil {
+			certdir := b.StateDir + "/image/" + b.MixVer + "/os-core-update/usr/share/clear/update-ca"
```
icarus-sparry: filepath.Join perhaps

builder/builder.go
```diff
@@ -1405,30 +1413,22 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
...
+			err = os.MkdirAll(certdir, 0755)
+			if err != nil {
+				helpers.PrintError(err)
+				return err
+			}
+			chrootcert := certdir + "/Swupd_Root.pem"
```
icarus-sparry: filepath.Join perhaps

builder/bundleset.go
```diff
@@ -54,7 +56,9 @@ func validateAndFillBundleSet(bundles bundleSet) error {
 	}
 	for _, b := range sortedBundles {
 		var allPackages []string
-		allPackages = append(allPackages, b.DirectPackages...)
+		for k := range b.DirectPackages {
```
icarus-sparry: Should this check that the bool associated with k is true or do we assert we will only every remove things using delete rather than setting the value to false?

matthewrsj: I *was* relying on removing the keys if necessary, but it would be safer and more future-proof to check the bool as well.

validate-swupd.sh
matthewrsj: Done.

builder/builder.go
matthewrsj: Done.

builder/builder.go
matthewrsj: Done.

builder/bundleset.go
matthewrsj: Done.

builder/chroots.go
```diff
@@ -80,6 +82,459 @@ func readBuildChrootsConfig(path string) (*buildChrootsConfig, error) {
...
+	}
+
+	return ioutil.WriteFile(filepath.Join(bundleDir, bundle.Name), nil, 0644)
+}
+
+func buildFullChroot(cfg *buildChrootsConfig, b *Builder, set *bundleSet, packagerCmd []string, chrootVersionDir, version string) error {
+	fmt.Println("Installing all bundles to full chroot")
+	totalBundles := len(*set)
+	i := 0
+	for _, bundle := range *set {
```
tmarcu: `Optionally` next time could save a line and just do for i, bundle := ... etc 

:smile: :100:

icarus-sparry: Actually saves 2 lines, the `i := 0` and the `i++`

builder/builder.go
```diff
@@ -1362,6 +1362,14 @@ func (b *Builder) BuildChroots(template *x509.Certificate, privkey *rsa.PrivateK
 		}
 	}
 
+	// Generate the certificate needed for signing verification if it does not exist
```
tmarcu: Optional Comment: This  move seems completely random, looks better next to the block that installs it.

builder/chroots.go
matthewrsj: doh, I know how to write Go I swear. Done.

builder/builder.go
matthewrsj: Actually the move is not completely random. The cert needs to be generated before the bundles are built so it can be optionally put into the os-core-update bundle and the full chroot. I did modify the comment because I'm not doing the insertion at this point.

tmarcu: :+1:

PR #216, opened 2018-03-06
build-deltas: Do not fail on missing previous versions.

matthewrsj:
A version can be missing if the minversion is newer than the previous
version being checked. This is not an error, but should be printed as a
warning to the user in case this is not expected.


matthewrsj:
that travis error doesn't seem related to my change...

matthewrsj:
@tmarcu looks like it is an issue with clrtrust generate running in the container.

matthewrsj:
Got info from clrtrust maintainer (@busykai) that there is an issue with clrtrust in clearlinux:latest. He told me how to fix the makefile to repair this. The fix will not break future fixes to clrtrust, but @busykai will let us know when we can remove it.

matthewrsj:
Passes \o/

PR #218, opened 2018-03-08
Allow for 'm' flag to be used in rename flag spot, fail on invalid manifests

matthewrsj:
The 'm' flag is used as a special flag by mixer-integration on the
client side to indicate a manifest record in the Manifest.MoM was
created by a local mix. Allow this flag and just set the rename flag to
renameUnset.

Fail when the old manifest exists but is invalid and cannot be parsed.
This is a clear failure case because the old manifest does exist at that
path and therefore the user expects it to be used.


matthewrsj:
Updated to reflect what was discussed in IRC.

PR #219, opened 2018-03-08
Add support for adding extra files to a bundle

matthewrsj:
Allow a user to add extra files to a bundle by adding a file per line to
`<imageDir>/<version>/<bundleName>-extra-files`. Each file must be an
absolute path. At this point the user must manually create the file in
the full chroot as well.

The workflow for a user adding additional files to a bundle would be
this:

```bash
$ mixer build chroots
$ echo /path/to/bar >> update/image/my-bundle-extra-files
$ mkdir -p update/image/full/path/to
$ cp /local/path/to/bar update/image/full/path/to/bar
$ mixer build update
```

This can definitely be improved in the future but this is a first step
that allows additional content on top of what dnf installs.

In the rare case that a post-install action for dnf install adds a file,
this file would already exist in the full chroot and would not need to
be copied in. The user would only need to append the path to that file
to the `<bundle>-extra-files` file to make sure it is included in that
bundle's manifests.

**based on #215** last commit is the relevant one here.

### DIFF
--- a/swupd/bundleinfo.go
+++ b/swupd/bundleinfo.go
@@ -60,7 +60,30 @@ func (m *Manifest) getBundleInfo(path string) error {
 		return err
 	}
 
-	return json.Unmarshal(biBytes, &m.bundleInfo)
+	err = json.Unmarshal(biBytes, &m.bundleInfo)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


matthewrsj:
I mentioned in the PR comment, but I'll reiterate. This is based on #215 and only the last commit is relevant for this PR.

matthewrsj:
`make batcheck` passes

PR #227, opened 2018-03-12
Install DirectPackages to full chroot

matthewrsj:
Do not just install the AllPackages list to the full chroot, since this
is just the list of all included packages. The DirectPackages need to be
installed instead since this is the fully resolved list for each bundle.
AllPackages do not need to be installed because all bundles are being
installed to the same place.

~~**DO NOT MERGE UNTIL I COMPLETE TESTING**~~

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -328,6 +328,24 @@ func parseNoopInstall(installOut string) []string {
 		if len(line) == 0 {
 			break
 		}
+
+		// When not running in a TTY dnf sets terminal line-width to a
+		// default 80 characters, then wraps their fields in an
+		// unpredictable way to be pleasing for they eye but not for
+		// the parser. If the line starts with more than one space (' ')
+		// character the previous line was wrapped and continues on
+		// this line.  Since we only care about the first field, the
+		// package name, just continue to the next line.
+		//
+		// example lines:
+		// ^ really-long-package-name-overflows-field$
+		// ^                         x86_64             12-10         clear$
+		// ^ reasonablepkgname       x86_64             deadcafedeadbeefdeadcafedeadbeef$
+		// ^                                                          clear$
+		if strings.HasPrefix(line, "  ") {
+			continue
+		}
+
 		pkgDeps = append(pkgDeps, strings.Fields(line)[0])
 	}
 	return pkgDeps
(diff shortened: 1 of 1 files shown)


kevincwells:
Huh. `bundle.AllPackages` was introduced as part of the chroot rewrite. If we don't need it now for the full chroot, can we get rid of it altogether? Or do we use it somewhere else?

matthewrsj:
We don't use it now, but I can see it being interesting in the future. DirectPackages is all packages and dependencies of the top-level bundle definition. If we ever want to expose a more complete list of packages we would need AllPackages since that has all the packages in the included bundles.

It isn't expensive to calculate and is cheap to keep around and may help us in the future, so I don't mind letting it hang out for now.

builder/bundles.go
```diff
@@ -475,12 +493,20 @@ func genUpdateBundleSpecialFiles(chrootDir string, cfg *buildBundlesConfig, b *B
...
-	args = append(args, bundle.AllPackages...)
-	err := helpers.RunCommandSilent(args[0], args[1:]...)
-	if err != nil {
-		return err
+	if len(bundle.DirectPackages) > 0 {
+		// There were packages directly included for this bundle so
+		// install to full chroot. This check is necessary so we don't
+		// call dnf install with no package listed.
+		for p := range bundle.DirectPackages {
+			args = append(args, p)
```
kevincwells: Why does this need to be in a loop, instead of just `args = append(args, bundle.DirectPackages...)` like above?

matthewrsj: Because this is a map not a slice.

kevincwells: Oh. Yes. Very well, then.

builder/bundles.go
```diff
@@ -328,6 +328,24 @@ func parseNoopInstall(installOut string) []string {
...
+		// unpredictable way to be pleasing for they eye but not for
+		// the parser. If the line starts with more than one space (' ')
+		// character the previous line was wrapped and continues on
+		// this line.  Since we only care about the first field, the
+		// package name, just continue to the next line.
+		//
+		// example lines:
+		// ^ really-long-package-name-overflows-field$
+		// ^                         x86_64             12-10         clear$
+		// ^ reasonablepkgname       x86_64             deadcafedeadbeefdeadcafedeadbeefd$
```
tmarcu: s/beefd/beef

matthewrsj: ^^ the real review

PR #232, opened 2018-03-13
Add trailing newline to mixbundles

kevincwells:
Previously mixer wrote the mix bundle list as a newline-separated list,
rather than as a newline-delimited list. This made things like "wc -l"
report one fewer than expected, as the final line contained no newline.

This trailing newline does not affect the existing parsing code.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -676,7 +676,7 @@ func (b *Builder) getDirBundlesListAsSet(dir string) (bundleSet, error) {
 // writeMixBundleList writes the contents of a bundle set out to the Mix Bundles
 // List file. Values will be in sorted order.
 func (b *Builder) writeMixBundleList(set bundleSet) error {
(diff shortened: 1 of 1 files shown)


builder/builder.go
```diff
@@ -676,7 +676,7 @@ func (b *Builder) getDirBundlesListAsSet(dir string) (bundleSet, error) {
 // writeMixBundleList writes the contents of a bundle set out to the Mix Bundles
 // List file. Values will be in sorted order.
 func (b *Builder) writeMixBundleList(set bundleSet) error {
-	data := []byte(strings.Join(getBundleSetKeysSorted(set), "\n"))
+	data := []byte(strings.Join(getBundleSetKeysSorted(set), "\n") + "\n")
```
icarus-sparry: This says to me "newline delimited list" but the commit message says "command delimited list"

kevincwells: Whoops!

kevincwells: fixed

PR #234, opened 2018-03-15
Rewrite the DNF conf generation

kevincwells:
Previously, mixer used m4 to generate the dnf conf (.yum.conf), and only
did so if the file did not already exist, so as to not overwrite people's
custom conf files. However, this meant that the check if a "[local]"
section was needed only ran if the file didn't exist. If a mix had already
been built, this check was skipped, and the local section was never inserted.

This patch removes the use of the yum.conf.in file and m4, instead using an
embedded docstring. Additionally, if the file already exists but the local
section is needed, mixer will now check the file to see if it has the necessary
local section. If not, the section will be appended to the file.


kevincwells:
Note: With just this patch, bats tests fail because of the rpm url being invalid.

builder/builder.go
```diff
@@ -1331,37 +1358,37 @@ func (b *Builder) BuildBundles(template *x509.Certificate, privkey *rsa.PrivateK
 	defer timer.WriteSummary(os.Stdout)
 
 	timer.Start("BUILD BUNDLES")
+	var addLocalConf bool
 	if _, err := os.Stat(b.DNFConf); os.IsNotExist(err) {
-		outfile, err := os.Create(b.DNFConf)
+		if err = ioutil.WriteFile(b.DNFConf, []byte(fmt.Sprintf(dnfConfFormat, b.UpstreamURL)), 0644); err != nil {
```
matthewrsj: Optional: This line is really long. I would put `if err != nil {` on the next line just to make it easier to read.

builder/builder.go
```diff
@@ -1312,6 +1310,35 @@ func (b *Builder) createMixBundleDir() error {
...
+obsoletes=1
+gpgcheck=0
+plugins=0
+installonly_limit=3
+reposdir=/root/mash
+
+[clear]
+name=Clear
+failovermethod=priority
+baseurl=%s/releases/$releasever/clear/x86_64/os/
```
matthewrsj: Since you have a `%s` here I see this as a pretty good use case for a template instead of a format string, especially if we want to make the conf more configurable in the future.

builder/builder.go
```diff
@@ -1312,6 +1310,35 @@ func (b *Builder) createMixBundleDir() error {
...
+enabled=1
+gpgcheck=0
+`
+const dnfConfLocalFormat = `
+[local]
+name=Local
+failovermethod=priority
+baseurl=file://%s
```
matthewrsj: Same template comment here, due to the `%s`

builder/builder.go
```diff
@@ -1331,37 +1358,37 @@ func (b *Builder) BuildBundles(template *x509.Certificate, privkey *rsa.PrivateK
...
-		}()
-		if b.RepoDir == "" {
-			cmd := exec.Command("m4", "-D", "UPSTREAM_URL="+b.UpstreamURL, b.DNFTemplate)
-			cmd.Stdout = outfile
-			if err = cmd.Run(); err != nil {
-				helpers.PrintError(err)
-				return err
-			}
-		} else {
-			cmd := exec.Command("m4", "-D", "MIXER_REPO",
```
matthewrsj: Glad to see this going away. We can remove `m4` from the mixer bundle once this is released.

builder/builder.go
kevincwells: I didn't go that route 'cause it seemed overkill for a single value, but I have no real objection to it and it does make it easier to configure in the future.

Lemme give it a shot.

kevincwells:
@matthewrsj  Redone with a template-based solution.

I like this better in general, as it also lets me have only one place that writes to the file -- it either writes the full file or just the `[local]` section.

builder/builder.go
```diff
@@ -34,6 +34,7 @@ import (
 	"strconv"
 	"strings"
 	"text/tabwriter"
+	textTemplate "text/template" // "template" conflicts with crypto/x509
```
tmarcu: :disappointed:

builder/builder.go
```diff
@@ -1312,6 +1311,39 @@ func (b *Builder) createMixBundleDir() error {
 	return nil
 }
 
+type dnfConf struct {
+	UpstreamURL, RepoDir string
```
tmarcu: I like doing this in function signatures... I feel conflicted about this in a struct. :man_shrugging: looks weird :)

builder/builder.go
```diff
@@ -1312,6 +1311,39 @@ func (b *Builder) createMixBundleDir() error {
...
+	Base, Local          bool
```
tmarcu: Would be great to explain what base and local mean either here or below where you are actually setting them, in a short comment.

kevincwells:
Force-pushed rebase after #239

builder/builder.go
```diff
@@ -1410,6 +1409,41 @@ func (b *Builder) createMixBundleDir() error {
...
+[local]
+name=Local
+failovermethod=priority
+baseurl=file://{{.RepoDir}}
+enabled=1
+gpgcheck=0
+priority=1
+{{end}}`
+
 // BuildBundles will attempt to construct the bundles required by using the m4
```
icarus-sparry: Outdated comment!

kevincwells: Thank you. This also reminded me that `m4` also needed to be removed from the dependency list. Commit added to fix both.

PR #235, opened 2018-03-15
bat tests: Fix local directory names and rpm paths

kevincwells:
Note: I have a much more extensive rewrite of the bats tests coming, but this is an easily separable first step. 

----------------

The mixerlib file and the CLEANDIRS variable in all of the individual
bat Makefiles still pointed to the old "local" and "rpms" directories in
some locations.

This, made 'make batclean' not clean up directories correctly. This likely
made some of the tests not actually do what they were supposed to do, as it
was looking for rpms in 'local-rpms', but putting them in 'rpms'.

This patch also updates the url for the "json-c" source rpm, as the
previous url no longer is active.


kevincwells:
**Note**: 
With just this patch, bats tests pass, but only because the local RPMS are never actually used (the `[local]` section is not written). 

With this patch _and_ #234, bats tests **fail** if `--new-swupd` is used with `mixer build update`.  Without the `--new-swupd` flag, tests pass.

kevincwells:
Just force-pushed minor change -- fixed the rpm URL _again_.

PR #236, opened 2018-03-16
Update bats

kevincwells:
This patch has two commits:
The first commit adds tests for all of the `mixer bundle ____` commands, which previously were not being tested. 

My bash-foo is seriously lacking, so I am confident there will be many "this doesn't actually test what you think it does" or "this could be done way more efficiently by doing _____" type comments from @icarus-sparry.  I tried to put comments describing the actual concept I was trying to test, so hopefully people can tell me how to do things better.

The second commit dramatically overhauls the build tests. Please see the commit message there for more details.

Overall, this patch significantly increases code coverage, tests a few build cases that weren't actually being tested before, and reduces average run time (in my testing) by about a minute and a half. The average time for me was about 3 minutes. And that's with the older, slower chroots. Hopefully that'll go down once `--new-chroots` can be used.


bat/tests/01-bundle-commands/run.bats
```diff
@@ -0,0 +1,83 @@
...
+  [[ "$output" =~ kernel-native ]]
+  [[ "$output" =~ bootloader ]]
+}
+
+@test "Add an upstream bundle to the mix" {
+  mixer bundle add editors
+
+  run mixer bundle list
+  [[ ${#lines[@]} > 4 ]]                           # More bundles in list now
+  [[ "$output" =~ editors[[:space:]]+\(upstream ]] # "editors" bundle is from upstream
```
matthewrsj: Do we really care about the `(upstream)` tag here? My thought is that that could change, for instance we may want to say `[upstream]` in the future or something, so we should only test the functional information (`editors`)

kevincwells: I don't care that there are parens around it, but I do care that it reports being from upstream.  I didn't spend a great amount of time on this regex, but I put the paren in (instead of, say, `.*` or something) because I didn't want to match
```
editors
otherbundle   (upstream)
```

matthewrsj:
> this currently has --new-chroots removed because of the local rpm bug.

Now that we know the local rpm was wrong (src rpm) are the bat tests fixed even with --new-chroots if the rpm is a valid one?

kevincwells:
> Now that we know the local rpm was wrong (src rpm) are the bat tests fixed even with --new-chroots if the rpm is a valid one?

I believe so. Need to rebase this whole thing and add the `--new-chroots` back in.

bat/lib/mixerlib.bash
```diff
@@ -42,7 +42,7 @@ mixer-create-update() {
 }
 
 mixer-add-rpms() {
-  mkdir -p ./local ./rpms
+  mkdir -p $BATS_TEST_DIRNAME/local-yum $BATS_TEST_DIRNAME/local-rpms
```
icarus-sparry: In the spirit of "This works, but I would write it differently", I note a couple of things.

The `-p` is there to suppress error messages in case the directory exists, you don't need to create any parent directories.

You can use shell expansion to reduce the amount of typing

     mkdir -p $BATS_TEST_DIRNAME/{local-yum,local-rpm}

as the local- is common you could go further, and say `mkdir -p $BATS_TEST_DIRNAME/local-{yum,rpm}` but I wouldn't in a script as it will make it harder to use say `grep` to find where this directory is being made.

bat/lib/mixerlib.bash
```diff
@@ -69,9 +69,9 @@ remove-package() {
 }
 
 download-rpm() {
-  mkdir -p ./rpms
-  pushd rpms
-  sudo curl -O $1
+  mkdir -p $BATS_TEST_DIRNAME/local-rpms
+  pushd $BATS_TEST_DIRNAME/local-rpms
```
icarus-sparry: Again in the "I would write it this way", starting a subshell rather than using pushd/popd is reasonable.

```
download-rpm()(
    mkdir -p $BATS_TEST_DIRNAME/local-rpms
   cd $BATS_TEST_DIRNAME/local-rpms
   sudo -E curl -LO "$1"
)

As a general comment always having double quotes around things that are being expanded is good practice, it avoids splitting on white-space (technically on $IFS). Adding -E to sudo keeps http_proxy variables around.

File: bat/tests/01-create-simple-mix/Makefile [hunk 4]
```diff
@@ -3,7 +3,7 @@
 check:
 	bats ./run.bats
 
-CLEANDIRS = ./update ./test-chroot ./logs ./.repos ./bundles ./update ./mix-bundles ./clr-bundles ./local ./results ./repodata ./rpms ./upstream-bundles ./local-bundles
+CLEANDIRS = ./update ./test-chroot ./logs ./.repos ./bundles ./update ./mix-bundles ./clr-bundles ./local-yum ./results ./repodata ./local-rpms ./upstream-bundles ./local-bundles
```
icarus-sparry: Long lines can be split in a couple of different ways, depending on if you want to override them from the command line when invoking `make`. Either you can introduce backslash newline (which gets replaced with a space by make, or you can say
```
CLEANDIRS = ./update ./test-chroot
CLEANDIRS += ./logs
```
The latter although popular, means that invoking `make CLEANDIRS='fred'` will still append ./logs onto it, so care should be taken to decided is CLEANDIRS is part of the api this makefile is presenting to the world.

kevincwells:
Rebased and re-pushed. With the most recent `build bundles` efficiencies, this now runs in ~2m30s on my machine.

@icarus-sparry thanks for the comments! I was expecting at least one or two (or a dozen) about the checks I wrote in `bat/tests/01-bundle-commands/run.bats`. Are they really Icarus-approved?

bat/tests/01-bundle-commands/run.bats
```diff
@@ -0,0 +1,83 @@
...
+  [[ ${#lines[@]} == 2 ]]                      # 'list local' returns 2 results
+  [[ "$output" =~ .*foobar.* ]]                # "foobar" bundle is in output
+}
+
+@test "Remove bundle from mix" {
+  mixer bundle remove editors
+
+  [[ $(ls -1q $BATS_TEST_DIRNAME/local-bundles | wc -l) == 2 ]] # Still 2 bundles in local-bundles
+
+  ! $(mixer bundle list | grep editors)                         # "editors" no longer in mix
```
icarus-sparry: You don't want the `$(` and `)` here,

If you had editors-foo as something that is output by `mixer bundle list` then the shell would try and run a program called `editors-foo`.

(You may want a `-q` to grep to make it not print anything, potentially makes debugging harder and you are not expecting anything anyway)

bat/tests/01-bundle-commands/run.bats
```diff
@@ -0,0 +1,83 @@
...
+}
+
+@test "Validate a bundle" {
+  echo "package" >> $BATS_TEST_DIRNAME/local-bundles/foobar
+
+  mixer bundle validate foobar
+
+  run mixer bundle validate foobar --strict
```
icarus-sparry: If you just want to check that the status is non-zero, then `! bundle validate foobar --strict` avoids the need for a `run`. If you are actually looking for a return code of 1 then this is perfect, but I think could do with a comment.

bat/tests/01-bundle-commands/run.bats
kevincwells: Ah right, gotcha. Fixed in forthcoming push.

bat/tests/01-bundle-commands/run.bats
kevincwells: Gotcha. And you guessed right -- our docs only state it as "non-zero", so no need for explicitly checking for 1.

Fixed in forthcoming push.

kevincwells:
Update pushed.

PR #239, opened 2018-03-20
Process package-scope bundles from package lists

matthewrsj:
This patch adds the capability to add package-scope bundles to your mix
either from upstream `packages` file or from a `local-packages` file in
your mix directory. These files list one package per line like so:
```
helloworld
vim
emacs
```
The bundles created from these packages will be named after the package
itself and will only have that one package defined (plus filesystem as a
hard-coded depencency). The bundle will have no includes. A bundle
definition for the `helloworld` package will look like this:
```
Name: helloworld
Filename: /path/to/{local-packages,packages}
DirectPackages: [helloworld, filesystem]
```
These packages can be added as bundles simply by running
```
mixer bundle add helloworld
```
and mixer will look for the definition in the following order:
1. local-bundles directory
2. local-packages file
3. upstream-bundles directory
4. packages file (upstream)

and will take the first definition it finds. The rest of mixer will see
this package as just another bundle and update content will be created
in the same manner as all other bundles.

The `validate-swupd.sh` script was added to define a `helloworld` package,
`bundle add` it, and test it with the client.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -63,23 +63,24 @@ type Builder struct {
 	BuildScript string
 	BuildConf   string
 
-	BundleDir       string
-	Cert            string
-	Format          string
-	LocalBundleDir  string
-	MixVer          string
-	MixVerFile      string
-	MixBundlesFile  string
-	RepoDir         string
-	RPMDir          string
-	StateDir        string
-	UpstreamURL     string
-	UpstreamURLFile string
-	UpstreamVer     string
-	UpstreamVerFile string
-	VersionDir      string
-	DNFConf         string
-	DNFTemplate     string
+	BundleDir         string
+	Cert              string
+	Format            string
+	LocalBundleDir    string
+	MixVer            string
+	MixVerFile        string
+	MixBundlesFile    string
+	LocalPackagesFile string
+	RepoDir           string
+	RPMDir            string
+	StateDir          string
+	UpstreamURL       string
+	UpstreamURLFile   string
+	UpstreamVer       string
+	UpstreamVerFile   string
+	VersionDir        string
+	DNFConf           string
+	DNFTemplate       string
 
 	Signing int
 	Bump    int
@@ -93,16 +94,20 @@ type Builder struct {
 	UpstreamVerUint32 uint32
 }
 
+var localPackages = make(map[string]bool)
+var upstreamPackages = make(map[string]bool)
+
 // New will return a new instance of Builder with some predetermined sane
 // default values.
 func New() *Builder {
 	return &Builder{
-		BuildScript:     "bundle-chroot-builder.py",
-		DNFTemplate:     "/usr/share/defaults/mixer/yum.conf.in",
-		UpstreamURLFile: "upstreamurl",
-		UpstreamVerFile: "upstreamversion",
-		MixBundlesFile:  "mixbundles",
-		MixVerFile:      "mixversion",
+		BuildScript:       "bundle-chroot-builder.py",
+		DNFTemplate:       "/usr/share/defaults/mixer/yum.conf.in",
+		UpstreamURLFile:   "upstreamurl",
+		UpstreamVerFile:   "upstreamversion",
+		MixBundlesFile:    "mixbundles",
+		LocalPackagesFile: "local-packages",
+		MixVerFile:        "mixversion",
 
 		Signing: 1,
 		Bump:    0,
@@ -514,6 +519,10 @@ func getUpstreamBundlesPath(ver string) string {
 	return filepath.Join(upstreamBundlesBaseDir, fmt.Sprintf(upstreamBundlesVerDirFmt, ver), upstreamBundlesBundleDir)
 }
 
+func (b *Builder) getLocalPackagesPath() string {
+	return filepath.Join(b.VersionDir, b.LocalPackagesFile)
+}
+
 func (b *Builder) getUpstreamBundles(ver string, prune bool) error {
 	if Offline {
 		return nil
@@ -568,9 +577,25 @@ func (b *Builder) getUpstreamBundles(ver string, prune bool) error {
 	return errors.Wrapf(os.Remove(tmptarfile), "Failed to remove temp bundle archive: %s", tmptarfile)
 }
 
+func setPackagesList(source *map[string]bool, filename string) error {
+	var err error
+	if len(*source) > 0 {
+		return nil
+	}
+	if _, err = os.Stat(filename); os.IsNotExist(err) {
+		return nil
+	}
+
+	*source, err = parsePackageBundleFile(filename)
+	return err
+}
+
 // getBundlePath returns the path to the bundle definition file for a given
-// bundle name, or error if it cannot be found. Looks first in local-bundles,
-// then upstream-bundles.
+// bundle name, or error if it cannot be found. Looks in the following order:
+// local-bundles/
+// local-packages
+// upstream-bundles/clr-bundles-<ver>/bundles/
+// upstream-bundles/clr-bundles-<ver>/packages
 func (b *Builder) getBundlePath(bundle string) (string, error) {
 	// Check local-bundles
 	path := filepath.Join(b.LocalBundleDir, bundle)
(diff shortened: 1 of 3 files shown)


matthewrsj:
bat tests, how they are currently implemented, do pass.

builder/builder.go
```diff
@@ -608,17 +648,40 @@ func getBundleSetKeysSorted(set bundleSet) []string {
 	return keys
 }
 
+// isLocalPackagePath checks if path is a local-packages definition file
+func isLocalPackagePath(path string) bool {
+	return strings.HasSuffix(path, "/local-packages")
```
kevincwells: I'd argue that this should be configurable in `builder.conf` alongside `LOCAL_BUNDLE_DIR`, presumably as `LOCAL_PACKAGE_LIST` or the like, and parsed into `b.LocalPackagesFile` (to mirror `b.MixBundlesFile`).

builder/builder.go
```diff
@@ -569,27 +572,64 @@ func (b *Builder) getUpstreamBundles(ver string, prune bool) error {
...
+	// Check local-packages
+	path = filepath.Join(pwd, "local-packages")
+	if len(localPackages) == 0 {
+		if _, err := os.Stat(path); err == nil {
+			localPackages, err = parsePackageBundleFile(path)
+			if err != nil {
+				return "", err
+			}
+		}
+	}
```
kevincwells: You have three places (here, L614, and L788) with these same 8 lines. I'd pull that out into a standalone `getPackageList(filename string) (map[string]bool, err)` function.

builder/builder.go
```diff
@@ -774,6 +860,16 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool,
 		if err != nil {
 			return errors.Wrapf(err, "Failed to read local bundles dir: %s", b.LocalBundleDir)
 		}
+		// handle packages defined in local-packages, if it exists
+		pwd, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		localPackagesFile := filepath.Join(pwd, "local-packages")
```
kevincwells: On top of making it configurable, pulling it out into `b.LocalPackagesFile` also lets us not have this file name hardcoded in multiple places.

builder/builder.go
```diff
@@ -719,6 +782,29 @@ func (b *Builder) getFullBundleSet(bundles bundleSet) (bundleSet, error) {
 	return set, nil
 }
 
+func populateSetFromPackages(source map[string]bool, dest map[string]*bundle, filename string) error {
```
kevincwells: While they're functionally equivalent, the type should be a `bundleSet` for `dest`.

builder/bundleset.go
```diff
@@ -221,6 +221,49 @@ func validateBundle(b *bundle) error {
...
+// names to be converted to one-package bundles (pundles). Returns a list of
+// pointers to bundle objects created from this list.
+func parsePackageBundleFile(filename string) (map[string]bool, error) {
+	packages := make(map[string]bool)
+	contents, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return nil, err
+	}
+
+	pundleList := strings.Split(string(contents), "\n")
```
kevincwells: Optional: If you wanted, you could use `helpers.ReadFileAndSplit()` to get these same 5 lines.

..but you already have them written.... so..

builder/builder.go
matthewrsj: I'm hesitant to make configuration changes with #237 still outstanding...

kevincwells: Fair point. But if you don't do it not, we'll just have to do it later. People loves them their puttin' stuff outside the workdir.

builder/builder.go
matthewrsj: good call, will do.

builder/builder.go
matthewrsj: err, if I do it now, #237 will have to change OR I will have to redo the changes after #237 is merged. I'd rather make an issue to move this to a config. I'll even assign it to myself if it'll make you happy.

kevincwells: Fine by me. If you do it before this goes into a real release, no one will even notice. Or probably no one will notice either way, as this is a brand new feature no one is using yet.

Well, Bun will notice ;)

builder/builder.go
kevincwells: @matthewrsj - Even if you don't make this configurable yet, you could still do the `b.LocalPackagesFile` part so the name isn't hard-coded multiple places.

matthewrsj: ok, I'll do that. I didn't do that because of how long the builder struct already was, but it'll make it easier to make configurable in the future.

builder/bundleset.go
matthewrsj: *sigh* fine...

matthewrsj: Actually, never mind. bundleset doesn't currently import helpers and I don't see the point in doing it for a simple file read/line splitting function.

kevincwells: Good answer.

builder/builder.go
matthewrsj: Done

builder/builder.go
matthewrsj: Done

builder/builder.go
matthewrsj: Done

builder/builder.go
kevincwells: Did you skip this one? I see the one in `isLocalPackagePath` got converted, but this one still shows as "local-path"

validate-swupd.sh
```diff
@@ -50,6 +50,13 @@ for i in $(seq $1 10 $2); do
 		popd
 		exit 1
 	fi
+	echo "helloworld" >> local-packages
+	mixer bundle add helloworld
+	if [[ $? -ne 0 ]]; then
+		echo "failed to add mix bundles"
+		popd
```
icarus-sparry: No point in changing directory via popd if you are going to exit.

builder/bundleset.go
```diff
@@ -221,6 +221,49 @@ func validateBundle(b *bundle) error {
 	return nil
 }
 
+// parsePundleFile parses a file at filename containing a flat list of package
+// names to be converted to one-package bundles (pundles). Returns a list of
+// pointers to bundle objects created from this list.
+func parsePackageBundleFile(filename string) (map[string]bool, error) {
```
icarus-sparry: Comment and function do not agree, this returns a map of strings, not a list of pointers.

builder/builder.go
```diff
@@ -608,17 +655,40 @@ func getBundleSetKeysSorted(set bundleSet) []string {
...
 		return nil, err
 	}
-	bundle, err := parseBundleFile(path)
-	if err != nil {
-		return nil, err
-	}
-	if err = validateBundleFilename(path); err != nil {
-		return nil, err
+	if b.isPathToPackageFile(path) {
+		bundle, err = newBundleFromPackage(name, path)
```
icarus-sparry: just

       return newBundleFromPackage(name, path)

then remove the else.

builder/bundleset.go
```diff
@@ -221,6 +221,49 @@ func validateBundle(b *bundle) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	pundleList := strings.Split(string(contents), "\n")
+	for _, p := range pundleList {
+		if len(p) == 0 {
+			continue
+		}
+
```
icarus-sparry: Is it worth adding 'if p[0] == '#' { continue ; }' so you can have commented out pundles in your pundle file, or even real comments?

validate-swupd.sh
matthewrsj: Done

builder/bundleset.go
matthewrsj: Fixed.

builder/builder.go
matthewrsj: Done

builder/bundleset.go
matthewrsj: Yep, done.

builder/builder.go
matthewrsj: Oops, done.

matthewrsj:
All comments addressed.

kevincwells:
I +1'd the changes you made to my comments, but now it's not passing Travis. :frowning_face:

matthewrsj:
Ah, I tried to proactively update this for #240 but got ahead of myself. I reverted that and I'll update whichever PR gets merged last.

PR #240, opened 2018-03-21
Perform all DNF transactions in bulk

matthewrsj:
When building bundles perform all DNF transactions in bulk. There are
two reasons to do this.

1. DNF is too smart for its own good. It will resolve dependencies
   differently depending on the packages listed in the transaction.
   Therefore we must perform these transactions with all relevant
   packages listed for the bundle being processed.
2. DNF transactions take time. When performing much fewer transactions
   in bulk the bundle builder can execute in a shorter time.

Certain manifest validation tools also require a complete dep-solved
list of all packages and included packages for a bundle to validate the
bundle manifest. For this reason use the bundle.AllPackages list and
leave the bundle.DirectPackages list as an informational list of just
the direct packages specified in the bundle definition file.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -82,24 +82,6 @@ func readBuildBundlesConfig(path string) (*buildBundlesConfig, error) {
 	return cfg, nil
 }
 
-func updateBundlePackages(set bundleSet, packages *sync.Map) {
-	for _, b := range set {
-		for k := range b.DirectPackages {
-			deps, ok := packages.Load(k)
-			if !ok {
-				continue
-			}
-
-			for _, d := range deps.([]string) {
-				d = strings.Trim(d, " ")
-				if len(d) > 0 {
-					b.DirectPackages[d] = true
-				}
-			}
-		}
-	}
-}
-
 var bannedPaths = [...]string{
 	"/var/lib/",
 	"/var/cache/",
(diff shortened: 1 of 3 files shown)


matthewrsj:
Looks like this causes `build bundles` to run a little over three times faster than it runs without this PR.

builder/bundles.go
```diff
@@ -212,18 +194,21 @@ func addUpdateBundleSpecialFiles(b *Builder, bundle *bundle) {
...
+
+	queryString := merge(packagerCmd, "repoquery", "-l", "--quiet")
+	for p := range bundle.AllPackages {
+		queryString = append(queryString, p)
+	}
+	outBuf, err := helpers.RunCommandOutput(queryString[0], queryString[1:]...)
+	if err != nil {
+		return err
+	}
+	for _, f := range strings.Split(outBuf.String(), "\n") {
```
icarus-sparry: It would be interesting to see the profile for this, The idea of generate a long string, then break it apart into an array of slices, and eventually discard it all, rather than pulling apart the returned string a bit at a time.

builder/bundles.go
```diff
@@ -212,18 +194,21 @@ func addUpdateBundleSpecialFiles(b *Builder, bundle *bundle) {
...
-		if err != nil {
-			return err
-		}
-		for _, f := range strings.Split(outBuf.String(), "\n") {
-			if len(f) > 0 {
-				addFileAndPath(bundle.Files, resolveFileName(f))
-			}
+
+	queryString := merge(packagerCmd, "repoquery", "-l", "--quiet")
+	for p := range bundle.AllPackages {
```
icarus-sparry: Here we are having to convert the map to an array again. We probably do this far too often. Might be worth keeping parallel data structures.

PR #247, opened 2018-03-28
[AFB] Format Bump Support

tmarcu:
Begin adding support for automating format bumps with mixer.

### DIFF
--- a/builder/format.go
+++ b/builder/format.go
@@ -1,0 +1,132 @@
+// Copyright © 2017 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package builder
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/clearlinux/mixer-tools/helpers"
+	"github.com/pkg/errors"
+)
+
+// UpdateFormatVersion updates the builder.conf file with a new format version
+func (b *Builder) UpdateFormatVersion(version string) error {
+	b.Config.Swupd.Format = version
+
+	newver := "${1}" + b.Config.Swupd.Format
+	var re = regexp.MustCompile(`(FORMAT=)[0-9]*`)
+
+	builderData, err := ioutil.ReadFile(b.BuildConf)
+	if err != nil {
+		return errors.Wrap(err, "Failed to read builder.conf")
+	}
+
+	builderEdit := re.ReplaceAllString(string(builderData), newver)
+
+	var filename string
+	if UseNewConfig {
+		filename, err = GetConfigPath("")
+		if err != nil {
+			return err
+		}
+
+		return b.Config.SaveConfig(filename)
+	}
+
+	builderOut := []byte(builderEdit)
+	if err = ioutil.WriteFile(b.BuildConf, builderOut, 0644); err != nil {
+		return errors.Wrap(err, "Failed to write new builder.conf")
+	}
+
+	return nil
+}
+
+// CopyFullGroupsINI copies the initial ini file which has ALL bundle definitions
+func (b *Builder) CopyFullGroupsINI() error {
+	return helpers.CopyFile(filepath.Join(b.Config.Builder.ServerStateDir, "full_groups.ini"), filepath.Join(b.Config.Builder.ServerStateDir, "groups.ini"))
+}
+
+// CopyTrimmedGroupsINI copies the new ini made with deleted bundles removed
+func (b *Builder) CopyTrimmedGroupsINI() error {
+	return helpers.CopyFile(filepath.Join(b.Config.Builder.ServerStateDir, "trimmed_groups.ini"), filepath.Join(b.Config.Builder.ServerStateDir, "groups.ini"))
+}
+
+// RevertFullGroupsINI copies back the full ini to the manifest creator accounts for deleted bundles
+func (b *Builder) RevertFullGroupsINI() error {
+	return helpers.CopyFile(filepath.Join(b.Config.Builder.ServerStateDir, "groups.ini"), filepath.Join(b.Config.Builder.ServerStateDir, "full_groups.ini"))
+}
+
+// RevertTrimmedGroupsINI copies back the trimmed INI so manifests are not created anymore for deleted bundles in new format
+func (b *Builder) RevertTrimmedGroupsINI() error {
+	return helpers.CopyFile(filepath.Join(b.Config.Builder.ServerStateDir, "groups.ini"), filepath.Join(b.Config.Builder.ServerStateDir, "trimmed_groups.ini"))
+}
+
+func (b *Builder) getLastBuildVersion() (string, error) {
+	var lastVer []byte
+	var err error
+
+	filename := filepath.Join(b.Config.Builder.ServerStateDir, "image/LAST_VER")
+	// Likely the first build
+	if lastVer, err = ioutil.ReadFile(filename); os.IsNotExist(err) {
+		return "", nil
+	} else if err != nil {
+		return "", errors.Wrap(err, "Cannot find last built version")
+	}
+	data := string(lastVer)
+	ver := strings.Split(data, "\n")
+
+	return ver[0], nil
+}
+
+// CheckBumpNeeded returns nil if it successfully deduces there is no format
+// bump boundary being crossed.
+func (b *Builder) CheckBumpNeeded() (bool, error) {
+	version, err := b.getLastBuildVersion()
+	if err != nil {
+		return false, err
+	}
+	// Check what format our last built version is part of
+	oldVer, err := b.DownloadFileFromUpstream(filepath.Join("/update", version, "format"))
+	if err != nil {
+		return false, errors.Wrapf(err, "Could not read format version from %s", b.UpstreamURL)
+	}
+	// Check what format our to-be-built version is part of
+	newVer, err := b.DownloadFileFromUpstream(filepath.Join("/update", b.MixVer, "format"))
+	if err != nil {
+		return false, errors.Wrapf(err, "Could not read format version from %s", b.UpstreamURL)
+	}
+
+	// Check both formats are real numbers
+	oldFmt, err := strconv.ParseUint(string(oldVer), 10, 32)
+	if err != nil {
+		return false, errors.New("Old format is not a number")
+	}
+	newFmt, err := strconv.ParseUint(string(newVer), 10, 32)
+	if err != nil {
+		return false, errors.New("Old format is not a number")
+	}
+
+	// We always need to perform a format bump if these are not equal
+	if oldFmt != newFmt {
+		return true, nil
+	}
+
+	return false, nil
+}
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -18,7 +18,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"runtime"
+	"strconv"
 
 	"github.com/clearlinux/mixer-tools/builder"
 	"github.com/clearlinux/mixer-tools/helpers"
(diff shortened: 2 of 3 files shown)


tmarcu:
Make batcheck passes this as well.

mixer/cmd/build.go
```diff
@@ -110,6 +112,144 @@ var buildBundlesCmd = &cobra.Command{
...
+		// same content
+		prevVersion, err := strconv.Atoi(b.MixVer)
+		if err != nil {
+			fail(err)
+		}
+		prevVersion -= 10
+		source := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer)
+		dest := filepath.Join(b.Config.Builder.ServerStateDir, "image", strconv.Itoa(prevVersion))
+		fmt.Println(" Copying +20 chroots to +10 chroots")
+		if err = helpers.RunCommandSilent("cp", "-al", source, dest); err != nil {
```
kevincwells: It's easy to do an exec like this, but could you achieve the same thing natively with `os.Link()` in an `ioutil.ReadDir()` loop? Or would you be losing the effects of the `--archive` you're passing to `cp`?

mixer/cmd/build.go
```diff
@@ -110,6 +112,144 @@ var buildBundlesCmd = &cobra.Command{
...
+		prevFormat--
+		if err = builder.UpdateFormatVersion(b, strconv.Itoa(prevFormat)); err != nil {
+			fail(err)
+		}
+		// Set mixversion to the +10 since we have used +20 up to this point
+		if err = b.DecrementMixVer(); err != nil {
+			fail(err)
+		}
+		// Re-read builder.conf after updating format/mixver
+		b, err = builder.NewFromConfig(config)
```
kevincwells: Same comment as above about `b.ReadVersions()`

builder/format.go
```diff
@@ -0,0 +1,66 @@
...
+	builderOut := []byte(builderEdit)
+	if err = ioutil.WriteFile(b.BuildConf, builderOut, 0644); err != nil {
+		return errors.Wrap(err, "Failed to write new builder.conf")
+	}
+
+	return nil
+}
+
+// CopyFullGroupsINI copies the initial ini file which has ALL bundle definitions
+func CopyFullGroupsINI(b *Builder) error {
```
kevincwells: I haven't thought carefully about how you're using these to see if it needs to be this way, so out of curiosity, why are all of these `func Function(b *Builder)` and thus called as `builder.Function(b)` down below, rather than `func (b *Builder) Function()` and `b.Function()`?

mixer/cmd/build.go
```diff
@@ -110,6 +112,144 @@ var buildBundlesCmd = &cobra.Command{
...
+			fail(err)
+		}
+		// Incremement regardless of flag because +20 must be the next version
+		if err = b.UpdateMixVer(); err != nil {
+			failf("Couldn't update Mix Version")
+		}
+	},
+}
+
+var buildNewFormatCmd = &cobra.Command{
```
kevincwells: What happens if you call this without having called `mixer build format-old` first?

mixer/cmd/build.go
```diff
@@ -110,6 +112,144 @@ var buildBundlesCmd = &cobra.Command{
 	},
 }
 
+// buildOldFormatCmd is used to build the final version in the current format (the +10)
+// and ready the mix state for the new (+20) version to be built by a (possibly) newer mixer
+var buildOldFormatCmd = &cobra.Command{
+	Use:   "format-old",
```
kevincwells: The `mixer build format-old` and `mixer build format-new` names aren't particularly intuitive to me. Maybe "last", like you use in the function descriptions?

mixer/cmd/build.go
```diff
@@ -110,6 +112,144 @@ var buildBundlesCmd = &cobra.Command{
...
+	Short: "Build the last version in the current format for the format bump",
+	Long:  `Build the last version in the current format for the format bump`,
```
kevincwells: Not necessary yet, but before this gets released we really could use a good long-format description for this and `mixer build format-new`

builder/format.go
```diff
@@ -64,3 +68,63 @@ func RevertTrimmedGroupsINI(b *Builder) error {
 	return helpers.CopyFile(filepath.Join(b.Config.Builder.ServerStateDir, "groups.ini"), filepath.Join(b.Config.Builder.ServerStateDir, "trimmed_groups.ini"))
 }
 
+func getLastBuildVersion(b *Builder) (string, error) {
+	// Likely the first build or some weirdness that requires us to get the container image and build it anyway
```
kevincwells: Does this comment refer to the error case below at L77?

builder/format.go
```diff
@@ -64,3 +68,63 @@ func RevertTrimmedGroupsINI(b *Builder) error {
...
+
+// CheckBumpNeeded returns nil if it successfully deduces there is no format
+// bump boundary being crossed.
+func CheckBumpNeeded(b *Builder) (bool, error) {
+	version, err := getLastBuildVersion(b)
+	if err != nil {
+		return false, err
+	}
+	// Check what format our last built version is part of
+	versionURL := fmt.Sprintf("https://download.clearlinux.org/update/%s/format", version)
```
kevincwells: You could rewrite this to use `b.getUpstreamFormatRange(version string)`, a) because it already exists and gets this value, and b) so you're not hardcoding the upstream url.

If you don't want use that (because it gets the other version values too and you're not using them), there's also the `b.DownloadFileFromUpstream` for times you need to grab individual files like this, which also lets you correctly use the `b.UpstreamURL` value automatically.

That said, `UpdateVersions` does the same conceptual check by doing one call to `format, first, latest, err := b.getUpstreamFormatRange(b.UpstreamVer)`, and then checking if the new format is different with `case nextUpstream < first, nextUpstream > latest:`, which could save you the second call.

builder/format.go
```diff
@@ -64,3 +68,63 @@ func RevertTrimmedGroupsINI(b *Builder) error {
...
+	resp, err = http.Get(versionURL)
+	if err != nil {
+		return false, errors.Wrapf(err, "Failed to http.Get() %s", versionURL)
+	}
+	newVer, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return false, errors.Wrapf(err, "Could not read format version from %s", versionURL)
+	}
+
+	// Check both formats are real numbers
```
kevincwells: Is this actually a concern? Do we have released versions with bad format values?

Also, elsewhere we use the `parseUint32` in `builder` to parse versions as `uint32`. If you don't need these to be `uint64`, you could just reuse that function.

builder/format.go
```diff
@@ -64,3 +68,63 @@ func RevertTrimmedGroupsINI(b *Builder) error {
...
+	}
+	data := string(lastVer)
+	ver := strings.Split(data, "\n")
+
+	return ver[0], nil
+}
+
+// CheckBumpNeeded returns nil if it successfully deduces there is no format
+// bump boundary being crossed.
+func CheckBumpNeeded(b *Builder) (bool, error) {
```
kevincwells: This and `b.UpdateVersions` do conceptually a very similar thing. Probably would be best to rewrite these to share some of the overlapping bits.

mixer/cmd/build.go
icarus-sparry: cp handles cross device copies. Of course we don't usually have extra disks mounted here, but we could.

A os.Link() would certainly be much faster.

tmarcu: The cp -al is what is currently done in the format bump flow and as icarus pointed out, handles copying over device boundaries better.

builder/format.go
tmarcu: Yes it does.. had to fix some var shadowing and didn't move the comment.

builder/format.go
tmarcu: Sure, I can consolidate and use the other.

builder/format.go
tmarcu: If someone grabs a url like 20130 which doesn't actually exist since it was aborted it will correctly download you a 404 page, which is then compared to a (possibly) real number - in my testing at least... that's what happened.

mixer/cmd/build.go
```diff
@@ -110,6 +112,144 @@ var buildBundlesCmd = &cobra.Command{
 	},
 }
 
+// buildOldFormatCmd is used to build the final version in the current format (the +10)
+// and ready the mix state for the new (+20) version to be built by a (possibly) newer mixer
+var buildOldFormatCmd = &cobra.Command{
```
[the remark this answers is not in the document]
tmarcu: Perhaps, this is trivial to move out to builder either way.

mixer/cmd/build.go
tmarcu: Bad stuff, you don't call this without calling the other :) just as you don't call build update before you call build bundles.

mixer/cmd/build.go
tmarcu: I had a long discussion with myself about how to name these and then gave up and left it as is to finish the rest of the code. I am :100:  looking for suggestions on how to name these better.

mixer/cmd/build.go
kevincwells: We could add a check for the `upstreamversion.bump` file that we discussed and error out early if this doesn't exist.

mixer/cmd/build.go
```diff
@@ -110,6 +112,144 @@ var buildBundlesCmd = &cobra.Command{
...
+
+		// Set format to format+1 so that the format file inserted into the
+		// update content is the new one
+		if err = builder.UpdateFormatVersion(b, buildFlags.format); err != nil {
+			fail(err)
+		}
+		setWorkers(b)
+
+		// Re-read builder.conf after updating format/mixver
+		b, err = builder.NewFromConfig(config)
```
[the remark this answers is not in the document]
tmarcu: It won't get the right version if this is removed :/

mixer/cmd/build.go
kevincwells: Am I wrong about it working here too, or did this get missed?

builder/format.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func getLastBuildVersion(b *Builder) (string, error) {
+	var lastVer []byte
+	var err error
+
+	filename := filepath.Join(b.Config.Builder.ServerStateDir, "image/LAST_VER")
+	// Likely the first build or some weirdness that requires us to get the container image and build it anyway
+	if lastVer, err = ioutil.ReadFile(filename); err != nil {
+		return "", errors.Wrap(err, "Cannot find last built version, must get container image")
```
kevincwells: So what does this comment/error message actually mean? Doesn't this case _always_ mean it's the first build? Is it actually possible for a build to succeed without generating this file? And I'm not following what difference running in a container would make.

builder/format.go
```diff
@@ -0,0 +1,129 @@
...
+
+	return ver[0], nil
+}
+
+// CheckBumpNeeded returns nil if it successfully deduces there is no format
+// bump boundary being crossed.
+func CheckBumpNeeded(b *Builder) (bool, error) {
+	version, err := getLastBuildVersion(b)
+	if err != nil {
+		return false, err
```
kevincwells: This needs to just `return false, nil` here, not `err`, because a first-time build isn't a failure case. It's theoretically possible that something else would cause an error here (the file not being readable??), but if this returns an error, whatever code calls `CheckBumpNeeded` will fail on first time builds.

builder/format.go
kevincwells: So was there a reason for this? Since you're passing a builder object to all of them, you must have a builder object already by the time you call them; why aren't they just methods on the builder object, then?

builder/format.go
tmarcu: It shouldn't be possible for it to succeed without this file existing. It could mean someone deleted it or broke it too.
However, running a build with this file not existing is not an error case, it will re-create the file and take it as a "build from 0" build and work successfully, which means the container creating the build will succeed in making an update.

builder/format.go
tmarcu: I think the code should check whether it returns true or false, not err, to decide to do a bump. If it returns false, err -> then no bump is needed but we need to get a container regardless.

builder/format.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func getLastBuildVersion(b *Builder) (string, error) {
+	var lastVer []byte
+	var err error
+
+	filename := filepath.Join(b.Config.Builder.ServerStateDir, "image/LAST_VER")
+	// Likely the first build
+	if lastVer, err = ioutil.ReadFile(filename); os.IsNotExist(err) {
+		return "", errors.Wrap(err, "Cannot find last built version, must get container image")
```
kevincwells: Wait, we need the opposite, right? `os.IsNotExist(err)` will tell us that it _is_ a "file does not exist" error. So we need something like

```golang
func getLastBuildVersion(b *Builder) (string, error) {
	filename := filepath.Join(b.Config.Builder.ServerStateDir, "image/LAST_VER")
	// Likely the first build
	lastVer, err := ioutil.ReadFile(filename)
	if os.IsNotExist(err) {
		return "", nil
	} else if err != nil {
		return "", errors.Wrap(err, "Cannot find last built version, must get container image")
	}
	return strings.Split(string(lastVer), "\n")[0], nil
}
```

Either way, I still don't understand the container part of the error message -- I'm maybe missing something, 'cause this seems tangential to whether or not it's run natively.

builder/format.go
```diff
@@ -0,0 +1,132 @@
...
+
+	return ver[0], nil
+}
+
+// CheckBumpNeeded returns nil if it successfully deduces there is no format
+// bump boundary being crossed.
+func (b *Builder) CheckBumpNeeded() (bool, error) {
+	version, err := b.getLastBuildVersion()
+	if err != nil {
+		return false, nil
```
kevincwells: Wait, sorry this is going back and forth, but now that we're only returning an error from `getlastBuildVersion` if there _is_ an error, this needs to go back to `return false, err`.

kevincwells:
I'm marking this as Approve Changes with the assumption that you will:
1. Fix the last thing I just commented on in `CheckBumpNeeded`
1. Squash (at least the last commit)

With that, approved and can merge.

PR #248, opened 2018-03-28
[AFB] Dockerize mixer.

kevincwells:
**DO NOT MERGE YET**
- This has a test command (`mixer build docker [command as args]`) that needs to be removed
- This will currently fail to run unless you've already downloaded the test docker image content and put it in the right directory, as it's not actually being published yet.

The key command here is `RunCommandInContainer(cmd []string) error`

I intend to squash these commits before merging.

------------------

This patch Dockerizes mixer. A new '--native' flag is added that, if
passed, will make mixer run on the host system the way it always did.
This defaults to false.

Before every command runs, if '--native' is false, mixer now pulls an
image file published by upstream that contains the core Clear Linux OS and
mixer toolchain for the version of Clear Linux defined in the upstreamversion
file, which mixer than uses to build a Docker image. Mixer then re-runs the
original mixer command within a container of that Docker image.

This ensures that you are always using a version of the mixer toolchain
compatible with the upstream version you are building off of. This is
required for building across format boundaries.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -45,7 +45,7 @@ const Version = "4.2.1"
 
 // Native controls whether mixer runs the command on the native machine or in a
 // container.
-var Native = false
+var Native = true
 
 // UseNewSwupdServer controls whether to use the new implementation of
 // swupd-server (package swupd) when possible. This is an experimental feature.
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -43,6 +43,10 @@ import (
 // Version of Mixer. Also used by the Makefile for releases.
 const Version = "4.2.1"
 
+// Native controls whether mixer runs the command on the native machine or in a
+// container.
+var Native = false
+
 // UseNewSwupdServer controls whether to use the new implementation of
 // swupd-server (package swupd) when possible. This is an experimental feature.
 var UseNewSwupdServer = false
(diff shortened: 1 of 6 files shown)


builder/builder.go
```diff
@@ -160,42 +164,90 @@ func (b *Builder) initDirs() error {
...
+		return err
 	}
 
-	body, err := ioutil.ReadAll(resp.Body)
+	return nil
+}
+
+// DownloadFileFromUpstreamAsString will download a file from the Upstream URL
+// joined with the passed subpath. It will trim spaces from the result.
+func (b *Builder) DownloadFileFromUpstreamAsString(subpath string) (string, error) {
```
tmarcu: Will have to update my code with this change, but I think it's a fine change.

builder/builder.go
```diff
@@ -2091,14 +2143,14 @@ func writeMetaFiles(path, format, version string) error {
 	return ioutil.WriteFile(filepath.Join(path, "mixer-src-version"), []byte(version), 0644)
 }
 
-func (b *Builder) getUpstreamFormatRange() (format string, first, latest uint32, err error) {
-	format, err = b.DownloadFileFromUpstream(fmt.Sprintf("update/%d/format", b.UpstreamVerUint32))
+func (b *Builder) getUpstreamFormatRange(version string) (format string, first, latest uint32, err error) {
+	format, err = b.DownloadFileFromUpstreamAsString(fmt.Sprintf("update/%s/format", version))
```
tmarcu: I think it'd be useful to keep the original functionality and check if version == "", then use b.UpstreamVerUint32 by default.

builder/docker.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	mounts := b.getDockerMounts()
+	for _, path := range mounts {
+		dockerCmd = append(dockerCmd, "-v", fmt.Sprintf("%s:%s", path, path))
+	}
+
+	dockerCmd = append(dockerCmd, getDockerImageName(format))
+	dockerCmd = append(dockerCmd, cmd[1:]...)
+	// TODO: Figure out chicken-egg problem of when to activate this flag. Until
```
tmarcu: Update this comment once we get an answer from DevOps about what images we can have built - always latest/out of cycle latest.

mixer/cmd/root.go
```diff
@@ -62,6 +63,39 @@ var RootCmd = &cobra.Command{
...
+			b, err := builder.NewFromConfig(config)
+			if err != nil {
+				fail(err)
+			}
+			if err := b.RunCommandInContainer(reconstructCommand(cmd, args)); err != nil {
+				fail(err)
+			}
+			// Cancel native run
+			cmd.RunE = nil
+			cmd.Run = func(cmd *cobra.Command, args []string) {} // No-op
```
tmarcu: 😃

builder/docker.go
```diff
@@ -107,19 +107,29 @@ func getDockerImageName(format string) string {
...
+	// TODO: Look into Docker Content Trust, or any other mechanism for verifying
+	// the validity of the docker image
+	cmd := []string{
+		"docker",
+		"images",
+		"-q", getDockerImageName(format),
+	}
+	output, err := helpers.RunCommandOutput(cmd[0], cmd[1:]...)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to check for docker image %q", getDockerImageName(format))
```
tmarcu: Failed check or failed to check here?

builder/docker.go
```diff
@@ -59,17 +61,32 @@ func (b *Builder) GetHostAndUpstreamFormats(upstreamVer string) (string, string,
 	return string(hostFormat), upstreamFormat, nil
 }
 
+func getCheckSum(filename string) string {
```
tmarcu: I'm not sure how in scope this can be yet, but I think the 'correct' solution here is to get an official signature for the content that we can check with our ClearLinuxRoot.pem, since this isn't part of any manifest. Alternatively, we could check that the mixer running in the container has the same hash as the manifest it shows up in.

builder/builder.go
```diff
@@ -2244,3 +2244,53 @@ Updated upstream: %d (format: %s)
...
+	}
+	format, first, latest, err := b.getUpstreamFormatRange(lastBuildVer)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("The upstream version for this build (%s) is outside the format range of your last mix "+
+		"(format %s, upstream versions %d to %d). This build cannot be done until you complete a "+
+		"format-bump build. Please run the following two commands to complete the format bump:\nmixer "+
+		"build format-old\nmixer build format-new\nOnce these have completed, you can re-run this build.\n",
```
tmarcu: I will update my code so this is accurate, currently the correct command is build format-old --format <currFmt + 1>

tmarcu: Although I kind of like it being explicit as well..

builder/docker.go
kevincwells: You'll have to walk me through how that stuff works, but I assume you're talking about verifying the `.sig` files that are published alongside the hashes?

builder/builder.go
kevincwells: Hmm, I'm not sure I understand the benefit yet. The original functionality was not having to pass anything, and it only knew how to work on the "current" version. If the calling sites have to pass something either way, I don't see why it's bad for them to pass the version they want. (I've also just never been a fan of passing empty string as an argument to things as a way of imitating optional fields.)

builder/docker.go
tmarcu: Yes, that ^ :)

builder/builder.go
tmarcu: Ok I can agree with that answer.

mixer/cmd/root.go
kevincwells: They so far haven't responded to my PR (https://github.com/spf13/cobra/pull/657) to make this noop hack unnecessary.

builder/docker.go
kevincwells: Fixed as part of rebase.

builder/docker.go
```diff
@@ -84,6 +84,7 @@ func (b *Builder) fetchDockerBase(ver string, baseDir string) error {
 
 const dockerfile = `FROM scratch
 ADD mixer.tar.xz /
+ENV LC_ALL="en_US.UTF-8"
```
tmarcu: +💯

PR #253, opened 2018-04-05
Deprecate and hide --new-swupd and --new-chroots flags

matthewrsj:
Deprecate these flags as the new implementation has been heavily tested and can now be default.

bat tests pass.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -43,10 +43,6 @@ import (
 // Version of Mixer. Also used by the Makefile for releases.
 const Version = "4.2.1"
 
-// UseNewSwupdServer controls whether to use the new implementation of
-// swupd-server (package swupd) when possible. This is an experimental feature.
-var UseNewSwupdServer = false
-
 // Offline controls whether mixer attempts to automatically cache upstream
 // bundles. In offline mode, all necessary bundles must exist in local-bundles.
 var Offline = false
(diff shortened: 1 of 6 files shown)


matthewrsj:
This patch will allow us to remove the following packages from the `mixer` bundle in Clear Linux:
* bundle-chroot-builder
* m4
* swupd-server
* urlgrabber
* yum
* yum-metadata-parser
* yum-utils

tmarcu:
Ok, testing this with mix-bundles remaining and new options continues to work and print a warning that LOCAL_BUNDLE_DIR is not defined.

kevincwells:
> This patch allows us to remove ... bundle-chroot-builder

As we discussed, we can't do that until we move fully to the new config, _unless_ we copy the default `builder.conf` to the `mixer` package, change our `Makefile` to install it, and change the legacy config code to pull it from `/usr/share/defaults/mixer/builder.conf`. I think this is the right move, 'cause keeping a dependency on `BCB` just for that file is silly, and it's an easy change that doesn't need to wait til `--new-config` goes away.

kevincwells:
>  ... people may be running with an old builder.conf and using mix-bundles still

If people are using a post-4.0 version of mixer, the `mix-bundles` directory is already auto-generated (and wiped clean) for every build (with the legacy BCB). So while people might be used to seeing it, they've already had to switch to using the `mixbundles` list and `local-bundles` directories

mixer/cmd/build.go
```diff
@@ -279,7 +270,10 @@ func init() {
 	RootCmd.AddCommand(buildCmd)
 
 	buildBundlesCmd.Flags().BoolVar(&buildFlags.noSigning, "no-signing", false, "Do not generate a certificate to sign the Manifest.MoM")
-	buildBundlesCmd.Flags().BoolVar(&builder.UseNewChrootBuilder, "new-chroots", false, "EXPERIMENTAL: Use new implementation of build chroots")
+	unusedBoolFlag := false
+	buildBundlesCmd.Flags().BoolVar(&unusedBoolFlag, "new-chroots", false, "")
+	_ = buildBundlesCmd.Flags().MarkHidden("new-chroots")
+	_ = buildBundlesCmd.Flags().MarkDeprecated("new-chroots", "this flag is no longer used")
```
kevincwells: I would prefer if these were descriptive of why. It's already going to report the flag as "deprecated", so just saying it's no longer used feels redundant. Perhaps `"New chroot functionality is now the standard behavior. This flag is obsolete and will be removed."`, or something similar.

mixer/cmd/root.go
```diff
@@ -141,7 +141,10 @@ func init() {
 	_ = RootCmd.PersistentFlags().MarkHidden("cpu-profile")
 
 	// TODO: Remove this once we migrate to new implementation.
-	RootCmd.PersistentFlags().BoolVar(&builder.UseNewSwupdServer, "new-swupd", false, "EXPERIMENTAL: Use new implementation of swupd-server when possible")
+	unusedBoolFlag := false
+	RootCmd.PersistentFlags().BoolVar(&unusedBoolFlag, "new-swupd", false, "")
+	_ = RootCmd.Flags().MarkHidden("new-swupd")
+	_ = RootCmd.Flags().MarkDeprecated("new-swupd", "this flag is no longer used")
```
kevincwells: Same comment about deprecation message.

builder/builder.go
```diff
@@ -1445,71 +1397,34 @@ func (b *Builder) BuildBundles(template *x509.Certificate, privkey *rsa.PrivateK
...
-			err = helpers.CopyFile(chrootcert, b.Config.Builder.Cert)
-			if err != nil {
-				helpers.PrintError(err)
-				return err
-			}
-		}
 	}
 
-	// TODO: Remove all the files-* entries since they're now copied into the noship dir
-	// do code stuff here
```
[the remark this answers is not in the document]
matthewrsj: we don't have files-* entries any more

PR #255, opened 2018-04-06
[AFB] Check access permissions on dirs before mounting

kevincwells:
This patch adds an access check on each directory in the config file
before mounting it to the container. This prevents the user from
accidentally (or surreptitiously) granting the container access to
directories they shouldn't.

The config parsing for directory extraction code is also refactored to
provide better results and more accurate error messages.

This patch also adds unit tests for directory access checks, directory extraction,
and mount path list reduction.

For these tests to not fail inside the Docker container on Travis, they
need to not be run as root. This patch thus also creates a non-root user
(but with wheelnopw membership), and runs as this user instead.

### DIFF
--- a/builder/docker.go
+++ b/builder/docker.go
@@ -179,50 +179,81 @@ func reduceDockerMounts(paths []string) []string {
 	return paths
 }
 
+// canAccess checks whether mixer has read/write access to the given directory.
+// A nil error means success.
+func canAccess(dir string) error {
+	// Check read access
+	if _, err := ioutil.ReadDir(dir); err != nil {
+		return err
+	}
+	// Check write acess
+	f, err := ioutil.TempFile(dir, "")
+	if err != nil {
+		return fmt.Errorf("open %s: permission denied", dir)
+	}
+	defer func() {
+		_ = os.Remove(f.Name())
+		_ = f.Close()
+	}()
+
+	return nil
+}
+
+// getPathDir returns the directory portion of a config file path. If the path
+// is already to a directory, the path is returned unchanged. If it is instead
+// to a file or does not exist, its parent is returned. This is needed because
+// some values in the config are paths to files or directories that get created
+// by the commands, but their parent directory exists and needs to be mounted.
+func getDirFromConfigPath(path string) (string, error) {
+	f, err := os.Stat(path)
+	if os.IsNotExist(err) || (err == nil && !f.Mode().IsDir()) {
+		path = filepath.Dir(path)
+		f, err = os.Stat(path)
+	}
+	if err != nil {
+		return "", err
+	}
+
+	return path, nil
+}
+
+// addConfigFieldPaths loops through each field in a config section, verifying
+// and adding its value to the mounts slice.
+func addConfigFieldPaths(config reflect.Value, mounts *[]string) error {
+	for i := 0; i < config.NumField(); i++ {
+		path, err := getDirFromConfigPath(config.Field(i).String())
+		if err != nil {
+			return err
+		}
+		if !strings.HasPrefix(path, "/") { // filepath.Dir can return "."
+			continue
+		}
+		if err := canAccess(path); err != nil {
+			return err
+		}
+		*mounts = append(*mounts, path)
+	}
+	return nil
+}
+
 // getDockerMounts returns a minimal list of all directories in the config that
 // need to be mounted inside the container. Only the "Buiilder" and "Mixer"
 // sections of the conf are parsed.
-func (b *Builder) getDockerMounts() []string {
-	// Returns the longest substring of path that is the path to a directory.
-	var getMaxPath func(path string) string
-	getMaxPath = func(path string) string {
-		f, err := os.Stat(path)
-		if os.IsNotExist(err) {
-			// Try again on parent. This happens because some values in config
-			// are paths to files that get created by the commands, but their
-			// parent directory exists and needs to be mounted.
-			path = filepath.Dir(path)
-			return getMaxPath(path)
-		} else if err != nil {
-			return ""
-		}
-		if f.Mode().IsDir() {
-			return path
-		}
-		return filepath.Dir(path)
-	}
-
+func (b *Builder) getDockerMounts() ([]string, error) {
 	wd, _ := os.Getwd()
 	mounts := []string{wd}
 
-	config := reflect.ValueOf(b.Config.Builder)
-	for i := 0; i < config.NumField(); i++ {
-		field := getMaxPath(config.Field(i).String())
-		if !strings.HasPrefix(field, "/") {
-			continue
-		}
-		mounts = append(mounts, field)
-	}
-	config = reflect.ValueOf(b.Config.Mixer)
-	for i := 0; i < config.NumField(); i++ {
-		field := getMaxPath(config.Field(i).String())
-		if !strings.HasPrefix(field, "/") {
-			continue
-		}
-		mounts = append(mounts, field)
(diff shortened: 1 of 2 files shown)


kevincwells:
Example output:

```
user@machine ~/mix $ mixer build bundles --native=false
Running command in container: "mixer build bundles"
ERROR: Failed to extract mountable directories from config: Cannot access "/usr/lib": permission denied
user@machine ~/mix $ vim builder.conf 
user@machine ~/mix $ mixer build bundles --native=false
Running command in container: "mixer build bundles"
ERROR: Failed to extract mountable directories from config: stat /usr/lib/foo: no such file or directory
```

builder/docker.go
```diff
@@ -179,50 +179,79 @@ func reduceDockerMounts(paths []string) []string {
...
 		}
-		return filepath.Dir(path)
+
+		return path, nil
+	}
+
+	// Loops through each field in a config section, verifying and adding its
+	// value to the mount slice
+	var addFieldPaths func(config reflect.Value, mounts []string) ([]string, error)
+	addFieldPaths = func(config reflect.Value, mounts []string) ([]string, error) {
```
matthewrsj: As mentioned on IRC, you can combine the above two lines

builder/docker.go
```diff
@@ -179,50 +179,79 @@ func reduceDockerMounts(paths []string) []string {
...
 // sections of the conf are parsed.
-func (b *Builder) getDockerMounts() []string {
-	// Returns the longest substring of path that is the path to a directory.
-	var getMaxPath func(path string) string
-	getMaxPath = func(path string) string {
+func (b *Builder) getDockerMounts() ([]string, error) {
+	// getPathDir returns the path if it is a directory, or its parent if it is
+	// a file or does not exist
+	var getPathDir func(path string) (string, error)
+	getPathDir = func(path string) (string, error) {
```
matthewrsj: As mentioned on IRC, you can combine the above two lines

kevincwells:
> Code looks good, but this logic definitely needs some testing.

Agreed. I'll update the PR with unit tests on Monday.

kevincwells:
@matthewrsj @tmarcu 
Very mildly updated the first commit (added the `ioutil.ReadDir` check to `canAccess`, stopped passing the mount list by copy, and made some error output more consistent), but it's almost entirely the same. 

The big news is the new second commit, adding unit tests for:
- read/write access check on directories
- directory from config path extraction
- mount list reduction

kevincwells:
@matthewrsj Jeez, that took a bit.

To get the new `canAccess` checks to work, the Travis container needs to be running as a non-root user. This most recent version updates the Travis `Dockerfile` to add a new `clr` user, and updates the `COPY` command so that that user has access to all of the directories it needs to run the tests.

builder/docker_test.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Fatalf("couldn't create temporary directory to write test cases: %s", err)
+	}
+	defer func() {
+		_ = os.RemoveAll(testDir)
+	}()
+
+	// Set up the dirs/files for testing
+	path := filepath.Join(testDir, "some/path/to/directory")
+	if err = os.MkdirAll(path, 0777); err != nil {
+		t.Errorf("Failed to create test directory: %q\n%s", path, err)
```
matthewrsj: Should be `Fatal` no? IMO test setup failures should be fatal, while actual test failures can be just `Error`s

builder/docker_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	path = filepath.Join(testDir, "some/path/to/file") // "some/path/to" already exists from previous command
+	f, err := os.Create(path)
+	if err != nil {
+		t.Errorf("Failed to create test file: %q\n%s", path, err)
```
matthewrsj: `Errorf` -> `Fatalf`

builder/docker_test.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Fatalf("couldn't create temporary directory to write test cases: %s", err)
+	}
+	defer func() {
+		_ = os.RemoveAll(testDir)
+	}()
+
+	// Set up the dirs/files for testing
+	path := filepath.Join(testDir, "shouldwork")
+	if err := os.Mkdir(path, 0777); err != nil {
+		t.Errorf("Failed to create test directory: %q\n%s", path, err)
```
matthewrsj: `Errorf` -> `Fatalf`

builder/docker_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	path = filepath.Join(testDir, "noread")
+	if err := os.Mkdir(path, 0333); err != nil {
+		t.Errorf("Failed to create test directory: %q\n%s", path, err)
```
matthewrsj: `Errorf` -> `Fatalf`

builder/docker_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	path = filepath.Join(testDir, "nowrite")
+	if err := os.Mkdir(path, 0555); err != nil {
+		t.Errorf("Failed to create test directory: %q\n%s", path, err)
```
matthewrsj: `Errorf` -> `Fatalf`

builder/docker_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	path = filepath.Join(testDir, "noreadwrite")
+	if err := os.Mkdir(path, 0111); err != nil {
+		t.Errorf("Failed to create test directory: %q\n%s", path, err)
```
matthewrsj: `Errorf` -> `Fatalf`

builder/docker_test.go
```diff
@@ -0,0 +1,170 @@
...
+		Filepath   string
+		ShouldFail bool
+	}{
+		{Filepath: "shouldwork"},
+
+		// Error cases.
+		{Filepath: "noread", ShouldFail: true},
+		{Filepath: "nowrite", ShouldFail: true},
+		{Filepath: "noreadwrite", ShouldFail: true},
+		{Filepath: "/usr/var", ShouldFail: true},
```
matthewrsj: We expect this to fail because we expect the user to not have access to `/usr/var`, yet you are already testing this case with the `noreadwrite` test above, and not relying on expected test-runner permissions. I'd remove this case, then remove the `!strings.HasPrefix(fullPath, "/")` check below.

builder/docker_test.go
```diff
@@ -0,0 +1,170 @@
...
+		t.Errorf("Failed to create test directory: %q\n%s", path, err)
+	}
+	path = filepath.Join(testDir, "noreadwrite")
+	if err := os.Mkdir(path, 0111); err != nil {
+		t.Errorf("Failed to create test directory: %q\n%s", path, err)
+	}
+
+	for _, tt := range tests {
+		fullPath := tt.Filepath
+		if !strings.HasPrefix(fullPath, "/") {
```
matthewrsj: As mentioned above I'd just drop the `/usr/var` check and then drop the `fullPath` variable and checks.

builder/docker_test.go
kevincwells: Good point -- I agree. I copied the bundleset tests as my starting point, so at least a couple other places are doing it this way. I won't change them as part of this, but that'd be a good cleanup task down the line.

builder/docker_test.go
kevincwells: True. I was originally thinking that it is very subtly a different failure, 'cause it's being denied from a group access failure, but it's not really proving anything new. I'll pull it.

PR #256, opened 2018-04-06
 Remove DNF state files from full chroot

matthewrsj:
When building the full chroot DNF lays down some state directories/files
which end up in the Manifest.full later on just by virtue of them being
there. These files should be removed from the chroot when all package
operations are completed.

Also
The fullDir path does not change for each bundle. Move it outside the
bundle install loop so it is not reassigned every iteration.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -540,11 +540,11 @@ func buildFullChroot(cfg *buildBundlesConfig, b *Builder, set *bundleSet, packag
 	}
 	fmt.Println("Installing all bundles to full chroot")
 	totalBundles := len(*set)
+	fullDir := filepath.Join(buildVersionDir, "full")
 	i := 0
 	for _, bundle := range *set {
 		i++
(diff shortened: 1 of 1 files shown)


matthewrsj:
@bryteise I know you are watching.

builder/bundles.go
```diff
@@ -511,6 +511,27 @@ func clearDNFCache(packagerCmd []string) error {
 	return helpers.RunCommandSilent(args[0], args[1:]...)
 }
 
+func rmDNFStatePaths(fullDir string) {
+	dnfStatePaths := []string{
+		"/var/lib/cache/yum",
+		"/var/cache/yum",
+		"/var/lib/dnf",
+		"/var/lib/rpm",
```
bryteise: /var/cache/dnf could very well appear in this list if a slightly different config was used (like the one I use for my dnf config).

bryteise: I'm a little surprised various yum/dnf log files aren't in this list.

matthewrsj: added

PR #257, opened 2018-04-11
Update Travis to use clearlinux/mixer-ci

kevincwells:
This patch updates the Travis Dockerfile to build off the new
clearlinux/mixer-ci image.

Previously, the Dockerfile built off the base clearlinux:latest image,
which meant it spent quite some time (~7 min) just building the Docker
image (swup bundl-add'ing the packages it needed) before it could build
and test mixer. This patch eliminates nearly all of that time, as the
Docker build only needs to download an image and copy the mixer files
over.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,4 @@
-FROM clearlinux:latest
-
-ENV GOPATH /home/gopath
-ENV PATH="/home/gopath/bin:${PATH}"
-COPY . / /home/gopath/src/github.com/clearlinux/mixer-tools/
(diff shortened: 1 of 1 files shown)


matthewrsj:
+1 for the changes.

matthewrsj:
Y'know, this PR might make it fast enough to actually run `make batcheck` in Travis again. Should test when the image is available.

kevincwells:
>Y'know, this PR might make it fast enough to actually run make batcheck in Travis again. Should test when the image is available.

That's the plan! This *should* be way faster than before, so this + batcheck will likely be shorter than it currently is without batcheck.

PR #258, opened 2018-04-12
build delta-packs: do not error when --to matches --from

matthewrsj:
The use case for this is when you want to always allow scripted calls to
build delta-packs to create a delta-pack from the min-version, which
might match the current version for a full min-version. This is
conceptually equivalent to finding 0 previous versions. Print the
information and return nil.


matthewrsj:
Passes bat tests

matthewrsj:
This PR will need a release when merged.

Issue #260, opened 2018-04-13
'mixer build bundles' fails with local and upstream package collisions.


kevincwells:
# Issue
Mixer fails to build update content if local and upstream rpm repositories have a package name collision with different content.

# Cause
`mixer build bundles` now relies on the output of `dnf repoquery -l` to get the contents of a given package and `dnf install` to put the contents of that package into the `full` chroot. `dnf repoquery` does not utilize the repository prioritization resolver that `dnf install` does, and instead lists the contents of every matching package in every repository. This results in the `<bundle>-info` file containing the file lists of both the upstream and the local version of the rpm, but the `full` chroot only containing the actual files from the (higher-priority) local rpm. `mixer build bundles` then fails as it attempts to resolve missing files.

# Proposed Solution(s)

- Further parse the output of `dnf install` to attain not just the list of package names, but also their version and release values. Use the full N-V-R value in the `dnf reqoquery -l`, so only one repository will return a match.
  - This has the limitation of still not allowing a full N-V-R collision between repositories. The implication is that if you want to replace an upstream package with a different version _of the same version number_, you **must** change the release number.
- Further parse the output of the `dnf install` to attain not just the list of package names, but also the repository they came from. Group packages by repository, and perform one batched `dnf repoquery -l --repo` per repo.
  - This is a more robust solution (full N-V-R collisions will be resolved correctly), but requires a much more significant refactor of the code.

kevincwells:
@matthewrsj - does this adequately reflect your understanding of the conversation we just had? Writing this down in case I can't fix this before I leave.

icarus-sparry:
Another solution is to implement the "zero root" approach and make the selection of the rpm's better (i.e. don't do `dnf install` at all, and make the virtual `dnf repoquery` know about priorities).

PR #261, opened 2018-04-14
Fix local package name collision bug

kevincwells:
Fixes https://github.com/clearlinux/mixer-tools/issues/260.
~~This passes my preliminary local testing, but I need to do additional testing with `validate-swupd.sh`.~~

------
Previously, mixer failed to build update content if local and upstream
repositories had a package name collision with different content.

'mixer build bundles' relies on the output of 'dnf repoquery -l' to get
the contents of a given package and 'dnf install' to put the contents
of that package into the full chroot. 'dnf repoquery' does not utilize
the repository prioritization resolver that 'dnf install' does, and
instead lists the contents of every matching package in every repository.
This results in the <bundle>-info file containing the file lists of both
the upstream and the local version of the rpm, but the full chroot only
containing the actual files from the (higher-priority) local rpm. 'mixer
build bundles' then fails as it attempts to resolve missing files.

This patch fixes this by separating the packages in each bundle based on
the repository they come from according to the 'dnf install' resolution.
For each bundle, a separate 'dnf repoquery -l --repo' command is made
for each repository, guaranteeing that only the correct repopository is
queried for a given package.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -192,20 +192,22 @@ func addUpdateBundleSpecialFiles(b *Builder, bundle *bundle) {
 	addFileAndPath(bundle.Files, filesToAdd...)
 }
 
-func resolveFilesForBundle(bundle *bundle, packagerCmd []string) error {
+func resolveFilesForBundle(bundle *bundle, repoPkgs map[string][]string, packagerCmd []string) error {
 	bundle.Files = make(map[string]bool)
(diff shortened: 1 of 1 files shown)


matthewrsj:
I'm running the validate-swupd.sh script on my own system right now to test this change. I'll compare manifests against upstream when that's done.

PR #263, opened 2018-04-20
builder: fix panic when trying to build bundles with invalid input

cmarcelo:
Sometimes dnf output will not contain the "Installing:\n" pattern,
e.g. if there is a package that doesn't exist in a bundle, the output
looks like

    Last metadata expiration check: 0:00:06 ago on Thu 19 Apr 2018 05:09:16 PM PDT.
    No match for argument: libc

and mixer would crash because the Split resulting slice wouldn't have
a second element.

The actual error will be caught later when the install of the "full
chroot" happens, so it is fine for now just ignore it.

Ideally the goroutine handling should be similar to fullfiles, so the
workers have a way to indicate that an error occurred and the program
can finish early. Added that as a TODO.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -307,9 +307,15 @@ func resolveFiles(numWorkers int, set bundleSet, bundleRepoPkgs map[string]map[s
 func parseNoopInstall(installOut string) map[string][]string {
 	repoPkgs := make(map[string][]string)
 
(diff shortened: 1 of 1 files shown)


cmarcelo:
Initial report comes from IRC, the output is in https://michal.pastebin.mozilla.org/9083475.

The double panic is another bug, already fixed in go 1.10.1.

PR #265, opened 2018-04-25
Add /usr/src/linux- to setBootFromPathname

### LINKED ISSUE #264: Need to support storing kernel source under /usr/src
https://github.com/clearlinux/mixer-tools/blob/c13b0dbf505eb750ec7d507f57be258274826a2e/swupd/heuristics.go#L62

Kernel source (e.g. to support building out-of-tree modules) is typically stored beneath /usr/src, but this path is specifically disallowed.

Example path: /usr/src/linux-4.16.4-556.native/

bwarden:
In setStateFromPathname(), paths with /usr/src/ are assigned
ModifierState, so excluded from being written. Kernel source (for
building out-of-tree modules) is intended to be stored under
/usr/src/linux-<release>, so is flagged by this rule. Add a rule in
setBootFromPathname to set ModifierBoot for these paths so that they're
handled the same as kernel modules (/usr/lib/modules) and other kernel
artifacts (/usr/lib/kernel).

Fixes #264

**Probably requires additional support in cbm to ensure cleanup along with kernel image**

### DIFF
--- a/swupd/heuristics.go
+++ b/swupd/heuristics.go
@@ -101,6 +101,7 @@ func (f *File) setBootFromPathname() {
 		"/usr/lib/kernel/",
 		"/usr/lib/gummiboot",
 		"/usr/bin/gummiboot",
+		"/usr/src/linux-",
 	}
 
 	for _, path := range bootPaths {


matthewrsj:
@bwarden have we solved this problem for a different directory already? Look here: https://github.com/clearlinux/swupd-client/commit/3afa5220f934d11a8435d5666e9ddeb211f56dc8

bwarden:
Yeah, looks like that approach could work. Shall I submit a PR over there instead?

matthewrsj:
> Shall I submit a PR over there instead?

Submit one over there *as well*.

But we should figure out if the kernel sources are going under `/usr/src/linux-*` or `/usr/src/kernel` first

matthewrsj:
marking as blocked by a necessary CBM improvement - CBM needs to be taught to clean these up so they don't get left behind on client systems.

bwarden:
Looks like this is no longer needed as I've moved the files under /usr/lib/modules

PR #266, opened 2018-04-26
Detect renames for pack creation, not manifest creation

matthewrsj:
Instead of performing rename detection for manifest creation between the
current and previous version, perform rename detection for every delta
pack that is created, thus maximizing the number of deltas created for
each pack.

This removes the now-unused rename flag from the manifest. Now that
swupd-client optimistically applies all deltas in a pack it no longer
relies on a rename flag to tell it if a delta is available.

### DIFF
--- a/swupd/packs_test.go
+++ b/swupd/packs_test.go
@@ -255,6 +255,64 @@ func TestCreatePackZeroPacks(t *testing.T) {
 	mustValidateZeroPack(t, ts.path("www/20/Manifest.shells"), ts.path("www/20/pack-shells-from-0.tar"))
 }
 
+func TestCreatePackNonConsecutiveDeltas(t *testing.T) {
+	ts := newTestSwupd(t, "create-pack-ncd")
+	defer ts.cleanup()
+
+	ts.Bundles = []string{"os-core", "contents"}
+
+	contents := strings.Repeat("large", 1000)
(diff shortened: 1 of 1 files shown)


swupd/packs_test.go
```diff
@@ -255,6 +255,64 @@ func TestCreatePackZeroPacks(t *testing.T) {
...
+	ts.Bundles = []string{"os-core", "contents"}
+
+	contents := strings.Repeat("large", 1000)
+	if len(contents) < minimumSizeToMakeDeltaInBytes {
+		t.Fatal("test content size is invalid")
+	}
+
+	ts.addFile(10, "contents", "/A", contents+"A")
+	ts.addFile(10, "contents", "/B", contents+"B")
+	ts.addFile(10, "contents", "/C", contents+"C")
```
cmarcelo: Optional: use separate contentsA, contentsB and contentsC (with different contents) to reduce the chance in the future we get deltas other the ones we expect if the delta picking logic change drastically.

matthewrsj: I'm going to call this good for now, since the test explicitly checks for the deltas it expects. If we change the delta picking logic drastically this test will fail and the developer can update it.

matthewrsj:
@icarus-sparry ping

PR #267, opened 2018-05-01
Add add-repo feature to mixer

matthewrsj:
The add-repo subcommand adds a remote repo configuration to the DNF conf
defined in the builder.conf. This allows users to configure mixer to use
remote RPM repositories when building mixes instead of being limited to
just local RPMs and upstream Clear Linux RPMs.

This feature is an important pre-requisite for enabling PPA-style repos
on the client side.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -37,6 +37,7 @@ import (
 
 	"github.com/clearlinux/mixer-tools/helpers"
 	"github.com/clearlinux/mixer-tools/swupd"
+	"github.com/go-ini/ini"
 	"github.com/pkg/errors"
 )
 
@@ -1317,24 +1318,7 @@ gpgcheck=0
 priority=1
 {{end}}`
 
-// BuildBundles will attempt to construct the bundles required by generating a
-// DNF configuration file, then resolving all files for each bundle using dnf
-// resolve and no-op installs. One full chroot is created from this step with
-// the file contents of all bundles.
-func (b *Builder) BuildBundles(template *x509.Certificate, privkey *rsa.PrivateKey, signflag bool) error {
-	// Fetch upstream bundle files if needed
-	if err := b.getUpstreamBundles(b.UpstreamVer, true); err != nil {
-		return err
-	}
-
-	// Generate the dnf config file if it does not exist.
-	// This takes the template and adds the relevant local rpm repo path if needed
-	fmt.Println("Building bundles...")
-
-	timer := &stopWatch{w: os.Stdout}
-	defer timer.WriteSummary(os.Stdout)
-
-	timer.Start("BUILD BUNDLES")
+func (b *Builder) newDNFConfIfNeeded() error {
 	conf := dnfConf{
 		UpstreamURL: b.UpstreamURL,
 		RepoDir:     b.Config.Mixer.LocalRepoDir,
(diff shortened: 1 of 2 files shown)


builder/builder.go
```diff
@@ -1640,6 +1647,64 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+		helpers.PrintError(err)
+		return err
+	}
+
+	read, err := ioutil.ReadFile(b.Config.Builder.DNFConf)
+	if err != nil {
+		return err
+	}
+
+	if strings.Contains(string(read), fmt.Sprintf("[%s]", repo.RepoName)) {
```
phmccarty: Since the dnf conf is an INI file, I would prefer to parse the INI and see if the appropriate section exists.

builder/builder.go
```diff
@@ -1640,6 +1647,64 @@ func (b *Builder) BuildImage(format string, template string) error {
 	return imagecmd.Run()
 }
 
+type dnfRepoConf struct {
+	RepoName, RepoURL string
+}
+
+const dnfConfRepoTemplate = `
```
phmccarty: Really minor issue, but perhaps add a blank line at the start of this template so that the INI sections are more clearly separated?

builder/builder.go
```diff
@@ -1317,24 +1317,7 @@ gpgcheck=0
...
-
-	// Generate the dnf config file if it does not exist.
-	// This takes the template and adds the relevant local rpm repo path if needed
-	fmt.Println("Building bundles...")
-
-	timer := &stopWatch{w: os.Stdout}
-	defer timer.WriteSummary(os.Stdout)
-
-	timer.Start("BUILD BUNDLES")
+func (b *Builder) newConfIfNeeded() error {
```
rchiossi: I think it would be better to call this `newDnfConfIfNeeded()`. We have too many things called `conf` in mixer, it helps to avoid confusion if things a a bit better defined.

mixer/cmd/rpms.go
```diff
@@ -15,6 +15,8 @@
 package cmd
 
 import (
+	"errors"
```
rchiossi: is this the `errors` import you wanna use? Other parts of the code aew using `https://github.com/pkg/errors` instead.

builder/builder.go
phmccarty: Because `[arbitrary-string]` could appear within a quoted string, and by parsing the file as INI, you would not have to worry about this possibility.

matthewrsj: T_T fine

builder/builder.go
matthewrsj: agree.

mixer/cmd/rpms.go
matthewrsj: If you do a grep for \"errors\" vs pkg/errors\" in the repo you'll see we are very inconsistent. I grabbed the one in the standard library because it provided the functionality I need.

builder/builder.go
matthewrsj: It adds a newline at the beginning and end of the template already. Therefore if you keep adding repos there will always be a blank line between the sections. The other base template also adds a trailing newline. I'd rather not have two newlines between the sections.

matthewrsj:
Comments addressed and changes pushed.

mixer/cmd/rpms.go
rchiossi: Indeed. Yet another thing to add to our todo list

builder/builder.go
phmccarty: Okay, I agree.

PR #273, opened 2018-05-02
Ignore filenames starting with '.'

### LINKED ISSUE #269: Allow LOCAL_BUNDLE_DIR ('local-bundles') to be a repository
My mix custom bundles are kept on its own git repository. It is very convenient to just git clone that repository on LOCAL_BUNDLE_DIR for mixing. It used to work just fine when 'mix-bundles' was the bundles folder.

Now if I try to do that I get the following errors:
```shell
[murilo@clearlinux mix]$ mixer bundle list
ERROR: read /tmp/mix/local-bundles/.git: is a directory
[murilo@clearlinux mix]$ mixer bundle add --all-local
ERROR: Failed to read local bundles dir: /tmp/mix/local-bundles: read /tmp/mix/local-bundles/.git: is a directory
```

icarus-sparry:
Previously the code was using ioutils.ReadDir which does a stat call
on every file, but none of the code required the results of the stat
calls, just the names.

In addition programs such as git would like to hold information in the
directories. The usual unix way to do this is to have filenames
beginning with '.', and these are ignored by default by things such as
shell globbing and the ls program.

fixes #269

Signed-off-by: Icarus Sparry <icarus.w.sparry@intel.com>

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -411,16 +411,16 @@ func (b *Builder) getUpstreamBundles(ver string, prune bool) error {
 
 	// Clear out other bundle dirs if needed
 	if prune {
-		files, err := ioutil.ReadDir(upstreamBundlesBaseDir)
+		files, err := helpers.ListVisibleFiles(upstreamBundlesBaseDir)
 		if err != nil {
 			return errors.Wrap(err, "Failed to read bundles for pruning.")
 		}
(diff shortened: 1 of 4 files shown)


matthewrsj:
I think we want them sorted. This list is used to print the bundles to the user.

PR #277, opened 2018-05-04
Declare local error for fullfile creation

matthewrsj:
Instead of using the same error variable across all workers declare a
new error variable for each goroutine. This prevents race conditions
where multiple routines are using the same error variable to make
decisions and failures are missed.

### DIFF
--- a/swupd/fullfiles.go
+++ b/swupd/fullfiles.go
@@ -66,6 +66,7 @@ func CreateFullfiles(m *Manifest, chrootDir, outputDir string, numWorkers int) (
 
 	taskRunner := func(info *FullfilesInfo) {
 		for f := range taskCh {
+			var tErr error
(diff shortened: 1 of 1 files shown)


matthewrsj:
CC: @cmarcelo

cmarcelo:
LGTM. Could you take a look at the other workers that might have a similar bug?

Optional: tErr is... ugly. But our conventions don’t let us shadow err, so I guess worth looking later into moving the worker to its own toplevel function and see if that looks better.

matthewrsj:
> so I guess worth looking later into moving the worker to its own toplevel function and see if that looks better.

Agree, we were just talking about this in IRC

icarus-sparry:
I tried adding `err := err` in the inline function, which shuts up vet, but then makes ineffassign complain. :sob:

PR #278, opened 2018-05-04
[DO NOT MERGE] -ETOOMANYCONFIGS

rchiossi:
This is the first 100% test coverage version of the config unification patch.

The first big change with this patch is the move of config from builder to its own package. While inside builder it could not be used inside swupd since that would generate a cyclic dependency. Moving it outside allows for reuse in both parts of the code.

Another change is the rename of a bunch of stuff called config. With config now in its own package, it had conflicts with the config object defined inside swupd and the config global variable used to hold the config file name in the cmd package. Both are now renamed properly.

Note that most of this patch is focused on removing the extra config files (namely mixversion, upstreamversion, upstreamurl and server.ini) and not in reworking the communication API between builder and swupd. That part still need some work to remove redundant parameter passing. Since that rework is also a big patch, I decided to split the PR and send that on top of this one.

Also, note that while test coverage is 100%, that does not include use with --new-config, so there might be crashes that sliped through my review. Adding --new-config to tests is something that I'm evaluating atm and wasn't included mostly because I don't wanna duplicate all tests just to remove then later when new config becomes default.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -1525,7 +1525,7 @@ func (b *Builder) buildUpdateContent(timer *stopWatch, mixVersion uint32, minVer
 		return errors.Wrapf(err, "failed to write update metadata files")
 	}
 	timer.Start("CREATE MANIFESTS")
-	mom, err := swupd.CreateManifests(mixVersion, minVersion, uint(format), b.Config.Builder.ServerStateDir)
+	mom, err := swupd.CreateManifests(b.Config, mixVersion, minVersion, uint(format), b.Config.Builder.ServerStateDir)
 	if err != nil {
 		return errors.Wrapf(err, "failed to create update metadata")
 	}
@@ -1607,7 +1607,7 @@ func (b *Builder) buildUpdateContent(timer *stopWatch, mixVersion uint32, minVer
 		fmt.Printf("Creating zero pack for %s to version %d\n", name, version)
 
 		var info *swupd.PackInfo
-		info, err = swupd.CreatePack(name, 0, version, outputDir, bundleDir, 0)
+		info, err = swupd.CreatePack(b.Config, name, 0, version, outputDir, bundleDir, 0)
 		if err != nil {
 			return errors.Wrapf(err, "couldn't make pack for bundle %q", name)
 		}
@@ -1944,7 +1944,7 @@ func (b *Builder) BuildDeltaPacks(from, to uint32, printReport bool) error {
 
 	bundleDir := filepath.Join(b.Config.Builder.ServerStateDir, "image")
 	fmt.Printf("Using %d workers\n", b.NumDeltaWorkers)
-	return createDeltaPacks(fromManifest, toManifest, printReport, outputDir, bundleDir, b.NumDeltaWorkers)
+	return b.createDeltaPacks(fromManifest, toManifest, printReport, outputDir, bundleDir, b.NumDeltaWorkers)
 }
 
 // BuildDeltaPacksPreviousVersions builds packs to version from up to
@@ -1988,7 +1988,7 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
 	bundleDir := filepath.Join(b.Config.Builder.ServerStateDir, "image")
 	for _, fromManifest := range previousManifests {
 		fmt.Println()
-		err = createDeltaPacks(fromManifest, toManifest, printReport, outputDir, bundleDir, b.NumDeltaWorkers)
+		err = b.createDeltaPacks(fromManifest, toManifest, printReport, outputDir, bundleDir, b.NumDeltaWorkers)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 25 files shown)


matthewrsj:
> Adding --new-config to tests is something that I'm evaluating atm and wasn't included mostly because I don't wanna duplicate all tests just to remove then later when new config becomes default.

We should definitely add --new-config to the bat tests. It is completely fine to remove the flag later. We need to be testing the --new-config option since we want to be going to that implementation as our default eventually.

builder/builder.go
```diff
@@ -1274,13 +1272,23 @@ func (b *Builder) ValidateBundles(bundles []string, lvl ValidationLevel) error {
...
 func (b *Builder) UpdateMixVer() error {
-	// Deprecate '.mixversion' --> 'mixversion'
-	if _, err := os.Stat(filepath.Join(b.Config.Builder.VersionPath, ".mixversion")); err == nil {
-		b.MixVerFile = ".mixversion"
-		fmt.Println("Warning: '.mixversion' has been deprecated. Please rename file to 'mixversion'")
+	mixVer, _ := strconv.Atoi(b.Config.Builder.MixVer)
+	b.Config.Builder.MixVer = strconv.Itoa(mixVer + 10)
+
+	if !config.UseNewConfig {
+		// Deprecate '.mixversion' --> 'mixversion'
```
matthewrsj: I've noticed that these deprecation codepaths only exist under the `!config.UseNewConfig` guard, when users go to `--new-config` we can expect them not to need the deprecation warning.

But what about when we get rid of `--new-config`? Is that the end of the deprecation period and we will stop doing those checks? Or will we have to re-introduce the deprecation codepaths to the regular codepath?

swupd/data.go
```diff
@@ -31,79 +32,32 @@ type dbgConfig struct {
 	src    string
 }
 
-type config struct {
-	stateDir  string
+type swupdData struct {
```
matthewrsj: This is a config, not data. Can you rename this appropriately?

swupd/data_test.go
```diff
@@ -5,29 +5,6 @@ import (
 	"testing"
 )
 
-func TestReadServerINI(t *testing.T) {
```
matthewrsj: You remove this test here because the `swupd` package doesn't read the config anymore. Can you add a similar one to where the config is actually being read?

matthewrsj:
Also please run `make lint` against the codebase.

swupd/data.go
rchiossi: Yeah, i know. I just dont want to call if config. That is exactly what I`ve been trying to remove all along do avoid confusion. Actually, I want to get rid of this file and keep config/config.go only, but that requires some more refactoring.

swupd/data_test.go
rchiossi: What do you mean? To create a new test file for config module? I removed this test because what it tests does not exists anymore.

builder/builder.go
rchiossi: When we remove --new-config, is this file exists there it will just be ignored. If by then the user hasn' t upgraded to the new format, that will be the least of his problems. All the rest will also not work.

rchiossi:
This PR conflicts with my own PR, but the other should go in first. Also there are some other changes to this patch that will add further modification. I'll wait for the other to get in before updating this PR.

builder/builder.go
kevincwells: I think I'm missing something. The deprecation warnings @matthewrsj is talking about are for people using the old names for the version files (e.g., `.clearurl` -> `upstreamurl`). Are we having a separate deprecation period for people using the version files at all?

I haven't read this whole PR thoroughly yet, but am I right that (if you're using `--new-config`) these files are just ignored altogether immediately? Should we have a period of time where, if these files exist, a warning is output, but the value in them is still used?

builder/builder.go
```diff
@@ -200,54 +184,65 @@ func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allU
...
 
-	// Deprecate '.clearversion' --> 'upstreamversion'
-	if _, err := os.Stat(filepath.Join(b.Config.Builder.VersionPath, ".clearversion")); err == nil {
-		b.UpstreamVerFile = ".clearversion"
-		fmt.Println("Warning: '.clearversion' has been deprecated. Please rename file to 'upstreamversion'")
-	}
-	if err := ioutil.WriteFile(filepath.Join(b.Config.Builder.VersionPath, b.UpstreamVerFile), []byte(upstreamVer), 0644); err != nil {
-		return err
+		if config.UseNewConfig {
+			if err := b.Config.SaveConfig(b.BuildConf); err != nil {
```
kevincwells: Not understanding yet why this call is inside the `if b.Config.Builder.UpstreamVer == "latest"` block. Don't you need to write back out the config either way after overwriting all the `b.Config.Builder.*` objects a few lines above?

rchiossi:
Closing as this has been completely reworked. PR #321 is the first part of the replacement PR.

PR #281, opened 2018-05-07
Re-write swupd-add-pkg into mixer

matthewrsj:
This commit rewrites swupd-add-pkg into a Go binary provided by mixer.
It uses mixer libraries to make manifest merges and manipulation easier
and is a first step in enabling external repos for mixer-integration
client-side.

Some minor changes to the swupd library were necessary to make the 'm'
flag a supported flag for integration manifests.

The usage of the new binary is slightly different from the old script.

```
mcswupd <bundle-name> <path/to/rpm>
```

Instead of requiring the user to copy the rpm to /usr/share/mix/rpms it
does the work for the user.

**mcswupd is a silly name I know, I'll change it if you want. It stands for mixer-content swupd**

### DIFF
--- a/mcswupd/main.go
+++ b/mcswupd/main.go
@@ -189,7 +189,7 @@ func main() {
 
 	flag.Usage = func() {
 		fmt.Fprintf(os.Stderr, `Usage of %s:
-	%s <bundle-name> <absolute/path/to/rpm>`, os.Args[0], os.Args[0])
+	%s <bundle-name> <path/to/rpm>`, os.Args[0], os.Args[0])
 	}
 
 	flag.Parse()
@@ -210,6 +210,12 @@ func main() {
 		failIfErr(err)
 	}
 
+	// do this before changing into the mixWS so relative paths can be used
+	err = os.Link(pkg, filepath.Join(mixWS, "local-rpms", filepath.Base(pkg)))
+	if !os.IsExist(err) {
+		failIfErr(err)
+	}
+
 	err = os.Chdir(mixWS)
 	failIfErr(err)
 
@@ -227,11 +233,6 @@ func main() {
 	pkgName := filepath.Base(strings.TrimRight(pkg, ".rpm"))
 	err = appendToFile(filepath.Join(mixWS, "local-bundles", bundle), fmt.Sprintf("%s\n", pkgName))
 	failIfErr(err)
-
-	err = os.Link(pkg, filepath.Join(mixWS, "local-rpms", filepath.Base(pkg)))
-	if !os.IsExist(err) {
-		failIfErr(err)
-	}
 
 	err = b.AddRPMList([]string{filepath.Base(pkg)})
 	failIfErr(err)
--- a/mcswupd/main.go
+++ b/mcswupd/main.go
@@ -187,6 +187,11 @@ func mergeMoMs(mixWS string, mixVer, lastVer int) error {
 func main() {
 	var err error
 
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, `Usage of %s:
+	%s <bundle-name> <absolute/path/to/rpm>`, os.Args[0], os.Args[0])
+	}
+
 	flag.Parse()
 	if len(flag.Args()) != 2 {
 		flag.Usage()
--- a/mcswupd/main.go
+++ b/mcswupd/main.go
@@ -1,0 +1,292 @@
+package main
+
+import (
+	"crypto/rsa"
+	"crypto/x509"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"runtime"
+	"strconv"
+	"strings"
+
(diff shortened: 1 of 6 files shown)


mcswupd/main.go
```diff
@@ -0,0 +1,298 @@
...
+FORMAT=1
+`
+
+func getCurrentVersion() (int, error) {
+	c, err := ioutil.ReadFile("/usr/lib/os-release")
+	if err != nil {
+		return -1, err
+	}
+
+	re := regexp.MustCompile(`VERSION_ID=(\d+)\n`)
```
icarus-sparry: Probably should change the RE to be

                       (?m)^VERSION_ID=(\d+)$

so things like this don't match

    UPSTREAM_VERSION_ID=666
    # the following matches VERSION_ID=22

swupd/files.go
```diff
@@ -77,11 +77,17 @@ var modifierBytes = map[ModifierFlag]byte{
 	ModifierBoot:   'b',
 }
 
-type frename bool
+// RenameFlag describes the third position in the flag string
+// and was used to represent a rename flag (deprecated) and is
+// now used to represent a mixer-generated manifest to be merged
+// with an upstream manifest via client-side mixer integration
+type RenameFlag uint8
```
icarus-sparry: So `RenameFlag` describes something completely different to renames! This can't possibly confuse anyone!

mcswupd/main.go
```diff
@@ -0,0 +1,298 @@
...
+	err = os.Chdir(mixWS)
+	failIfErr(err)
+
+	b, err := builder.NewFromConfig(filepath.Join(mixWS, "builder.conf"))
+	failIfErr(err)
+	err = b.InitMix(fmt.Sprintf("%d", ver), fmt.Sprintf("%d", mixVer),
+		false, false, "https://download.clearlinux.org", false)
+	failIfErr(err)
+	b.NumBundleWorkers = runtime.NumCPU()
+	b.NumFullfileWorkers = runtime.NumCPU()
```
icarus-sparry: At some stage probably ought to have parameters to set these numbers of workers.

mcswupd/main.go
```diff
@@ -0,0 +1,298 @@
...
+func buildBundles(b *builder.Builder) error {
+	if _, err := os.Stat(b.Config.Builder.Cert); os.IsNotExist(err) {
+		fmt.Println("Generating certificate for signature validation...")
+		privkey, err := helpers.CreateKeyPair()
+		if err != nil {
+			return errors.Wrap(err, "Error generating OpenSSL keypair")
+		}
+		template := helpers.CreateCertTemplate()
+		err = b.BuildBundles(template, privkey, false)
+		if err != nil {
```
icarus-sparry: Doesn't errors.Wrap return nil if passed nil? So you can replace lines 134 to 137 with

           return errors.Wrap(b.BuildBundles(template, privkey, false), "Error building bundles")

If we just declared privkey and template as variables, we could collapse this function a bit more.

mcswupd/main.go
matthewrsj: Good point, done.

mcswupd/main.go
```diff
@@ -0,0 +1,298 @@
...
+			return errors.Wrap(err, "Error building bundles")
+		}
+	}
+	return nil
+}
+
+func excludeName(man *swupd.Manifest, exclude string) {
+	for i := range man.Files {
+		if man.Files[i].Name == exclude {
+			man.Files = append(man.Files[:i], man.Files[i+1:]...)
```
icarus-sparry: Not the fastest way to filter but as we probably are only going to be excluding one name it doesn't matter.

mcswupd/main.go
```diff
@@ -0,0 +1,298 @@
...
+		return 0
+	}
+
+	return v
+}
+
+// RunCommandOutput executes the command with arguments and stores its output in
+// memory. If the command succeeds returns that output, if it fails, return err that
+// contains both the out and err streams from the execution.
+func appendToFile(filename, text string) error {
```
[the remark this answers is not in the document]
matthewrsj: oh, forgot to remove the function comment when I removed the function... sorry

swupd/files.go
matthewrsj: Yeah... this could use some clean up but I didn't want to do it in this PR.

mcswupd/main.go
matthewrsj: The parameters exist through the CLI, but not through the API unfortunately. I agree with you.

mcswupd/main.go
matthewrsj: Done, looks much better.

matthewrsj:
Added a usage string as well with the latest commit.

matthewrsj:
Added another commit to allow relative paths to the RPM argument. I'll squash these commits once the PR is approved, I just don't want to mess up @icarus-sparry's review.

matthewrsj:
Closing in favor of #284

PR #287, opened 2018-05-10
Add mixer manual pages

matthewrsj:
Add manual pages for the top-level mixer command and all subcommands.


Makefile
```diff
@@ -46,6 +46,7 @@ install: gopath
 	$(GOPATH)/bin/mixer-completion zsh --path $(DESTDIR)/usr/share/zsh/site-functions/_mixer
 	install -m 00755 pack-maker.sh $(DESTDIR)/usr/bin/mixer-pack-maker.sh
 	install -m 00755 superpack-maker.sh $(DESTDIR)/usr/bin/mixer-superpack-maker.sh
+	install -m 00644 docs/* $(DESTDIR)/usr/share/man/man1/
```
ahkok: This will install the RST files into those folders too?

ahkok: And won't work if you add `.5` section man pages, but for now it's ok.

matthewrsj: > This will install the RST files into those folders too?

Ooops good point.

> And won't work if you add .5 section man pages, but for now it's ok.

Yeah I figured I'd cross that bridge when I get there.

matthewrsj: > This will install the RST files into those folders too?

fixed.

ahkok: use `$(MANPAGES)` here

matthewrsj: ok, *that's* done now.

ahkok:
Content seems largely OK, major improvement over `void`. Nice.

matthewrsj:
> @bryteise Isn't now the perfect time to start a man page trend though?

Here you go.

PR #292, opened 2018-05-14
Add man page for mixin

matthewrsj:
Adds a man page for mixin. Should be merged after #287 and is based on that PR. The only relevant commit in this PR is the last one.


matthewrsj:
OK, this can be merged now that #287 has been merged.

PR #297, opened 2018-05-15
 delta: add size check against full compressed file

matthewrsj:
Do not keep a delta if its size is larger than or equal to the size of
the compressed full file since download size is not being optimized in
this case.

This is not a critical error so do not remove the delta if the
supporting information can not be determined. For example, this check
fails if full files were not created before creating deltas because
there is no compressed full file to check. This is the case with many of
our unit tests.


swupd/delta.go
```diff
@@ -96,6 +96,31 @@ func createDeltasFromManifests(c *config, oldManifest, newManifest *Manifest, nu
 	return deltas, nil
 }
 
+// deltaLargerThanFull returns true if the delta file is larger than the
```
tmarcu: It also returns true if it's the same size :eagle:

matthewrsj: ogoodpoint

matthewrsj:
@tmarcu updated.

PR #298, opened 2018-05-15
bat: Perform integer greater than check in test

matthewrsj:
This started failing, not sure why it was passing before but not any
more. Needs to perform an integer comparison check with -gt.


bat/tests/01-bundle-commands/run.bats
```diff
@@ -29,7 +29,7 @@ setup() {
   mixer bundle add editors
 
   run mixer bundle list
-  [[ ${#lines[@]} > 4 ]]                           # More bundles in list now
+  [[ ${#lines[@]} -gt 4 ]]                           # More bundles in list now
```
kevincwells: Can you line back up the comments?

kevincwells:
Looks like I used string comparison for integer comparison in a few other places (e.g., `[[ ${#lines[@]} == 4 ]]`).  Perhaps they should all be changed over in one fell swoop?

PR #301, opened 2018-05-17
Reoder which version builds first to simplify

tmarcu:
The +20 build can be built fully at the beginning all the way though,
removing the need to swap tooling to build the +10 (just so things are
built in chronological order). The +10 build is created with the -
possibly - old tooling only after, and reduces the amount of
back-and-forth swapping of tooling.


tmarcu:
@kevincwells just throwing this out early in case you wanted to skim, need to test it again with the swap in place now.

mixer/cmd/build.go
```diff
@@ -112,12 +112,35 @@ var buildBundlesCmd = &cobra.Command{
...
+	Run: func(cmd *cobra.Command, args []string) {
+		b, err := builder.NewFromConfig(config)
+		if err != nil {
+			fail(err)
+		}
+		setWorkers(b)
+		err = b.BuildUpdate(buildFlags.prefix, buildFlags.minVersion, buildFlags.format, buildFlags.noSigning, !buildFlags.noPublish, buildFlags.keepChroot)
+		if err != nil {
+			failf("Couldn't build update: %s", err)
+		}
```
kevincwells: Erm, wait, this is just doing a regular update build. This command needs to do the whole `b.RunCommandInContainer(cmd)` thing for both of the sub commands, right?

```go
	cmd := strings.Split("mixer build upstream-format new", " ")
	b.RunCommandInContainer(cmd)
	cmd = strings.Split("mixer build upstream-format old", " ")
	b.RuNCommandInContainer(cmd)
```

kevincwells:
If I understand correctly, this still expects `--format` to be passed to the `mixer build upstream-format new`, right? Or did that change get in and I'm just reading it wrong?

mixer/cmd/build.go
kevincwells: Wait, I just realized this is the `mixer build upstream-format` one. That means it *also* has to do the finagling of `upstreamversion` before running in container.

`mixer build format-bump` is the one that needs the above 4 lines without the extra finagling.

tmarcu:
@kevincwells this now handles the format internally so you don't need to pass it in. Your comment was correct before I pushed this new version :)

builder/format.go
```diff
@@ -39,16 +39,21 @@ func (b *Builder) UpdateFormatVersion(version string) error {
 		return errors.Wrap(err, "Failed to read builder.conf")
 	}
 
-	builderEdit := re.ReplaceAllString(string(builderData), newver)
+	var builderEdit string
+	if !UseNewConfig {
```
kevincwells: The flow of this whole function is really weird now, with a bunch of stuff being created and never used for certain paths, and back-and-forth between strings and bytes. 

I think it can be simplified to:
```go
func (b *Builder) UpdateFormatVersion(version string) error {
	b.Config.Swupd.Format = version

	if UsenewConfig {
		var config string
		if config, err = GetConfigPath(config); err != nil {
			return err
		}

		var mc MixConfig
		if err = mc.LoadConfig(config); err != nil {
			return err
		}

		return mc.SetProperty(config, "Swupd.FORMAT", version)
	}

	builderData, err := ioutil.ReadFile(b.BuildConf)
	if err != nil {
		return errors.Wrap(err, "Failed to read builder.conf")
	}

	var re = regexp.MustCompile(`(FORMAT=)[0-9]+`)
	newver := []byte("${1}" + b.Config.Swupd.Format)
	builderData = re.ReplaceAll(builderData, newver)
	
	if err = ioutil.WriteFile(b.BuildConf, builderData, 0644); err != nil {
		return errors.Wrap(err, "Failed to write new builder.conf")
	}

	return nil
}
```

Notes:
- I changed your regex from `(FORMAT=)[0-9]*` to `(FORMAT=)[0-9]+`. Feel free to change it back if you want, but I didn't think we supported an empty format value (it's listed as `required:"true"`).
- @rchiossi -- the above usage illustrates that it'd probably be good to have a version of `SetProperty` that operates on the **current config file from which the `MixConifg` object was built**, so that we could just call `b.Config.SetProperty("Swupd.FORMAT", version")`

mixer/cmd/build.go
```diff
@@ -157,13 +184,25 @@ var buildOldFormatCmd = &cobra.Command{
...
+		// TODO: Inject any extra files (certs, etc) here
+		// if err := AddBundleExtrasFiles(); err != nil {...}
+
+		ver, err := strconv.Atoi(b.MixVer)
+		if err != nil {
 			fail(err)
 		}
 
+		// Build the +20 update so we don't have to switch tooling in between
+		err = b.BuildUpdate(buildFlags.prefix, ver, buildFlags.format, buildFlags.noSigning, !buildFlags.noPublish, buildFlags.keepChroot)
```
kevincwells: Doesn't this need to be `newFormat` (or... `b.Config.Swupd.Format`) rather than `buildFlags.format`?

mixer/cmd/build.go
```diff
@@ -157,13 +184,25 @@ var buildOldFormatCmd = &cobra.Command{
...
+		if err != nil {
+			failf("Couldn't build update: %s", err)
+		}
+
+		// Reset upstreamversion if necessary
```
kevincwells: Why do this here? It's still in the +10 build down below. Should both be doing it?

I haven't thought this through very carefully yet, but I think we probably want to pull it out of both and put it in the `mixer build upstream-format` command.

tmarcu:
COMMENTS ADDRESSED :raised_hands:

mixer/cmd/build.go
```diff
@@ -117,21 +116,6 @@ var buildUpstreamFormatCmd = &cobra.Command{
 	Short: "Use to create the necessary builds to cross an upstream format",
 	Long:  `Use to create the necessary builds to cross an upstream format`,
 	Run: func(cmd *cobra.Command, args []string) {
```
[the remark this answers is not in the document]
kevincwells: It's just a placeholder for what we're about to fill in.

PR #305, opened 2018-05-18
mixin: read current version from upstreamversion

### LINKED ISSUE #303: mixin: do not always set mix version to system version * 1000
This causes the mix version to just keep multiplying by 1000 and not reflect the upstream version at all. The upstreamurl file needs to be used instead.

matthewrsj:
If the upstreamversion file exists read the current version from there.
Only read from /usr/lib/os-release if the upstreamversion file does not
exist. The os-release file is modified every time a local mix is
migrated to (version * 1000) so reading from that file every time causes
the mix version to multiply by 1000 every mix build.

Fixes #303

### DIFF
--- a/mixin/helpers.go
+++ b/mixin/helpers.go
@@ -30,7 +30,15 @@ import (
 )
 
 func getCurrentVersion() (int, error) {
-	c, err := ioutil.ReadFile("/usr/lib/os-release")
+	// if upstreamversion exists, use that
+	c, err := ioutil.ReadFile(filepath.Join(mixWS, "upstreamversion"))
+	if err == nil {
+		return strconv.Atoi(strings.TrimSpace(string(c)))
+	}
+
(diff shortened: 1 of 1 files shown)


validate-swupd.sh
```diff
@@ -43,11 +43,11 @@ for i in $(seq $1 10 $2); do
...
-	mixer bundle add --all-upstream
-	if [[ $? -ne 0 ]]; then
-		echo "failed to add mix bundles"
-		exit 1
-	fi
+	#mixer bundle add --all-upstream
+	#if [[ $? -ne 0 ]]; then
+	#	echo "failed to add mix bundles"
+	#	exit 1
+	#fi
```
rchiossi: Any reason to leave this commented out in the code?

matthewrsj: fixed

PR #307, opened 2018-05-18
swupd-inspector: initial import

cmarcelo:
swupd-inspector contains various commands that help read data of a
swupd repository: download the fullfile for a specific file, see the
Manifest for a specific version, see the differences between two
versions.

The 'diff' subcommand was one of the tools used to evaluate whether
the new swupd implementation was generating correct results compared
with the existing clear implementation.

The clientState struct is almost the same as the one used in
swupd-extract, and given the two use cases, a potential candidate to
move over to swupd package (or at least to an internal package so we
can use without being API). I'm postponing this to when we make
further updates to either of the programs.

Like swupd-extract, it does have a shortcut to handle the main Clear
Linux repository

    // Shows the content of Manifest.vim at a specific OS version
    $ swupd-inspector cat clear/22460 Manifest.vim

    // Download a fullfile for a given file
    $ swupd-inspector get clear/22460 /usr/bin/vim

    // Show the differences between two versions
    $ swupd-inspector diff clear/22450 clear/22460

    // List the different versions of a file since a given OS version
    $ swupd-inspector log clear/22450 /usr/bin/vim

In those examples, "clear/..." could be replace for a URL of a
repository (plus version) of a mix. The downloaded metadata is cached
based on the URL, allowing further queries to be cached, even when
querying mixes.

### DIFF
--- a/swupd-inspector/diff.go
+++ b/swupd-inspector/diff.go
@@ -1,0 +1,268 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"path/filepath"
+	"sort"
+
+	"github.com/clearlinux/mixer-tools/internal/client"
+	"github.com/clearlinux/mixer-tools/swupd"
+)
+
+// TODO: MoM signature verification.
+
+// Terminal colors.
+var (
+	RED   = "\x1b[31m"
+	GREEN = "\x1b[32m"
+	RESET = "\x1b[0m"
+)
+
+func runDiff(cacheDir string, args []string) {
+	var (
+		noColor bool
+		strict  bool
+	)
+
+	flagSet := flag.NewFlagSet("diff", flag.ExitOnError)
+	flagSet.BoolVar(&noColor, "no-color", false, "use no colors")
+	flagSet.BoolVar(&strict, "strict", false, "also compare version numbers")
+	flagSet.Usage = usage
+	err := flagSet.Parse(args)
+	if err != nil {
+		log.Fatalf("ERROR: %s", err)
+	}
+
+	if len(flagSet.Args()) != 2 {
+		usage()
+		return
+	}
+
+	if noColor {
+		RED = ""
+		GREEN = ""
+		RESET = ""
+	}
+
+	baseA, versionA := parseURL(flagSet.Arg(0))
+	baseB, versionB := parseURL(flagSet.Arg(1))
+
+	var stateA, stateB *client.State
+	stateDirA := filepath.Join(cacheDir, convertContentBaseToDirname(baseA))
+	stateA, err = client.NewState(stateDirA, baseA)
+	if err != nil {
+		log.Fatalf("ERROR: %s", err)
+	}
+
+	stateDirB := filepath.Join(cacheDir, convertContentBaseToDirname(baseB))
+	if stateDirB == stateDirA {
+		if baseB == baseA {
+			stateB = stateA
+		} else {
+			// This should be a rare case, if we hit this we should improve
+			// our normalization function.
+			stateDirB = stateDirB + "_other"
+			stateB, err = client.NewState(stateDirB, baseB)
+			if err != nil {
+				log.Fatalf("ERROR: %s", err)
+			}
+		}
+	} else {
+		stateB, err = client.NewState(stateDirB, baseB)
+		if err != nil {
+			log.Fatalf("ERROR: %s", err)
+		}
+	}
+
+	momA, err := stateA.GetMoM(versionA)
+	if err != nil {
+		log.Fatalf("ERROR: %s", err)
+	}
+	momB, err := stateB.GetMoM(versionB)
+	if err != nil {
+		log.Fatalf("ERROR: %s", err)
+	}
+
+	fmt.Printf(`=== Differences from A to B
+
+  A Base:            %s
+  A Version:         %s
+  A State directory: %s
+
+  B Base:            %s
+  B Version:         %s
+  B State directory: %s
+
+`, baseA, versionA, stateDirA, baseB, versionB, stateDirB)
+
+	sortFiles(momA)
+	sortFiles(momB)
+
+	fmt.Println("=== Manifest.MoM")
+
+	type bundlePair struct {
+		Name string
+		A, B *swupd.File
+	}
+	var bundles []*bundlePair
+
+	// TODO: Check all them are manifests...
+
+	walkFiles(momA.Files, momB.Files, func(a, b *swupd.File) {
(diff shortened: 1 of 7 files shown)


cmarcelo:
@matthewrsj @tmarcu this would follow the same approach we taken for `swupd-extract`: I'll be the maintainer of it, but want to keep in the same repo to avoid API breakages given the use of swupd package. I'm sure there'll be tons of improvements to be made, but I think for an initial import this is in a decent shape.

I'm open to other solutions on how we deploy this, e.g. I'm fine moving this and swupd-extract to another repository, and just vendor swupd package. Let me know what you prefer!

Also, I think would be convenient if both tools were available in Clear Linux, would be OK to patch Makefile (and the Clear package) to include them?

matthewrsj:
> I'll be the maintainer of it,

Ok, this essentially means you are still an active contributor to mixer, and I can make serious requests for you to change things about this PR, correct? If you were not an active contributor I would overlook some things if it all worked just to get the functionality in so we could modify it...

> Also, I think would be convenient if both tools were available in Clear Linux, would be OK to patch Makefile (and the Clear package) to include them?

Fine by me. The `mixer-tools` project name is starting to make sense again.

swupd-inspector/archive.go
```diff
@@ -0,0 +1,78 @@
...
+	"bytes"
+	"compress/bzip2"
+	"compress/gzip"
+	"fmt"
+	"io"
+
+	"github.com/clearlinux/mixer-tools/swupd"
+)
+
+// TODO: Evaluate where to put this, partially duplicate with code in packs.
```
matthewrsj: Partially? Can't the swupd library do this already? I would much prefer you extend the existing capabilities rather than duplicating much of the implementation into another package in the same project.

swupd-inspector/archive.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (ctr *compressedTarReader) Close() error {
+	if ctr.compressionCloser != nil {
+		return ctr.compressionCloser.Close()
+	}
+	return nil
+}
+
+// Compression algorithms have "magic" bytes in the beginning of the file to identify them.
+var (
```
matthewrsj: These are definitely already defined in the swupd package. Those should be exported if you need to use them here.

swupd-inspector/archive.go
```diff
@@ -0,0 +1,78 @@
...
+	gzipMagic  = []byte{0x1F, 0x8B}
+	xzMagic    = []byte{0xFD, '7', 'z', 'X', 'Z', 0x00}
+	bzip2Magic = []byte{'B', 'Z', 'h'}
+)
+
+// newCompressedTarReader creates a struct compatible with tar.Reader reading from uncompressed or
+// compressed input.
+func newCompressedTarReader(rs io.ReadSeeker) (*compressedTarReader, error) {
```
matthewrsj: Seriously, you copied this function, just use the one already written.

cmarcelo:
>>    I'll be the maintainer of it,
> Ok, this essentially means you are still an active contributor to mixer, and I can make serious requests for you to change things about this PR, correct?

The main intention behind being maintainer of this (and swupd-extract) is to not put burden into you folks. If you feel like its better for Clear team to take maintainership of it, I'm fine. (As long as we don't break mkosi in the case of swupd-extract :-). What do you prefer?

That said, ask for changes, we can discuss them case by case. In general I'm inclined to get it in first, and we iterate on improving it.

swupd-inspector/main.go
```diff
@@ -0,0 +1,208 @@
...
+by default, so clear/20520 refer to the same as the URL above. Other
+aliases can be defined in $HOME/.config/swupd-inspector/aliases in the
+format ALIAS=URL per line.
+`)
+}
+
+func main() {
+	log.SetFlags(0)
+
+	if len(os.Args) < 2 {
```
matthewrsj: Ok, I would *much* prefer you use `cobra` here to manage your CLI. This is not a simple CLI and using cobra will help organize it a lot better. It will also keep your usage text consistent with the other main `mixer-tools` binaries, `mixer` and `mixin`.

swupd-inspector/archive.go
cmarcelo: We don't export those functions / structs, but maybe its time to go ahead and do it (at least in an internal package). I'll fix this duplication.

swupd-inspector/main.go
cmarcelo: Will do this.

swupd-inspector/archive.go
cmarcelo: Done.

swupd-inspector/archive.go
cmarcelo: Ditto.

swupd-inspector/archive.go
cmarcelo: Done.

cmarcelo:
This PR depends on #314.

I've pushed this new version (without Cobra) to simplify review until I do the Cobra change.

swupd-inspector/main.go
cmarcelo: Done.

cmarcelo:
Rebased.

cmarcelo:
And added the Makefile changes.

matthewrsj:
I get a panic when I run

```
$GOPATH/bin/swupd-inspector diff https://download.clearlinux.org/update/22600/ https://<internal-staging-url>/update/22600/
```

cmarcelo:
> I get a panic when I run

Fixed. Thanks :-)

matthewrsj:
@cmarcelo you force-pushed so I can't tell what the fix was. Can you point it out?

swupd-inspector/diff.go
```diff
@@ -0,0 +1,268 @@
...
+			// This should be a rare case, if we hit this we should improve
+			// our normalization function.
+			stateDirB = stateDirB + "_other"
+			stateB, err = client.NewState(stateDirB, baseB)
+			if err != nil {
+				log.Fatalf("ERROR: %s", err)
+			}
+		}
+	} else {
+		stateB, err = client.NewState(stateDirB, baseB)
```
cmarcelo: @matthewrsj 

This "else" was missing. Probably got wrongly removed when I was shuffling code around to the internal package.

Everything would work fine if both URLs were in the same repo (the thing I was mostly testing lately), hence got unnoticed.

cmarcelo:
Ping :-)

tmarcu:
This works for me using both upstream URLs and mixing with internal.

PR #312, opened 2018-05-18
swupd: fix manifest creation

rchiossi:
The compareIncludes() function indicates if two include lists are the
same. On the other hand, the changedIncludes variable indicates if the
list has changed, so it is always the opposite of what is returned by
compareIncludes()nd it must be swapped on assignment.

Also, the comparison using DeepEqual cannot be used to compare the list.
We only care if the include list is still the same, not if the properties
of that include have changes. This problem was never triggered since the\
result of this call was being misused.


matthewrsj:
well, well, well

/me waits for travis

matthewrsj:
Please write a more descriptive commit message title.

PR #313, opened 2018-05-18
mixin: fix lint warning

cmarcelo:


matthewrsj:
This hasn't shown up running `make lint` on my system or on travis...

cmarcelo:
Would be nice if make lint printed the version so we can detect those cases easily ;-)

cmarcelo:
Updated with `go get -u gopkg.in/alecthomas/gometalinter.v2` and the error persists. Taking a quick glance at the code, the warning seems sane.

Unfortunately the `gometalinter.v2 --version` gives me `gometalinter version master built from ? on` (sic), which is not helpful to compare.

tmarcu:
I'm closing since I can't reproduce locally or on travis, and fixed some similar bugs to this one in the last PR. Please re-open if issue persists.

cmarcelo:
I still have the issue :-(. Can you confirm what version of gometalinter.v2 you have? (And how you get it)

Locally I have the following commit as HEAD for gometalinter.v2: "Fix --version and bump to 2.0.5" (46cc1ea...)

matthewrsj:
@cmarcelo I just ran `go get -u github.com/alecthomas/gometalinter` and it still doesn't give me any linter errors (many commits since your checkout though).

matthewrsj:
@cmarcelo I also don't get the linter error you are when I install gometalinter from the commit you referenced.

cmarcelo:
Gometalinter docs says:

```
There are two options for installing gometalinter.

    1. Install a stable version, eg. go get -u gopkg.in/alecthomas/gometalinter.v2. I will generally only tag a new stable version when it has passed the Travis regression tests. The downside is that the binary will be called gometalinter.v2.
    2. Install from HEAD with: go get -u github.com/alecthomas/gometalinter. This has the downside that changes to gometalinter may break.
```

I was trying the first option. Do you get the error with the stable version? Or should I be running the unstable one?

cmarcelo:
https://github.com/alecthomas/gometalinter#installing

cmarcelo:
Note that if you update your gometalinter.v2, it should fail too. (That was my case).

I suggest adding a Makefile target to install/update the metalinter version we want to use, so that everyone can be in the same page (even the CI).

PR #314, opened 2018-05-18
swupd-extract: move clientState to an internal package

cmarcelo:
The client.State struct is used to download and query manifests and
files from a specific swupd repository (and version). It does support
caching of data.

Besides the expected changes to make it is own package

- Only print "downloading" messages when Verbose is set in the struct
- Allow not passing an expected hash when verifying bundles
- Add a GetMoM function to the client.State

This is still an internal package since I think it needs more
iterations before becoming a public package. That said, the upcoming
swupd-inspector tool was implemented with almost no significant change
to the original code.

### DIFF
--- a/internal/client/state.go
+++ b/internal/client/state.go
@@ -1,4 +1,4 @@
-package main
+package client
 
 import (
 	"archive/tar"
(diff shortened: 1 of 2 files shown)


cmarcelo:
@matthewrsj preparation for swupd-inspect. You haven't asked, but its the same case as the archive functions.

icarus-sparry:
I am in 2 minds over the use of internal, if it is going to be useful for Diva then we will need to think about how we break this up into packages some more.

PR #315, opened 2018-05-19
Remove docker image build.

kevincwells:
Up until now, mixer was pulling the "content image" (mixer.tar.xz)
published by DevOps, then doing a 'docker build' command locally on the
client. This functionality has been moved to DevOps as part of release,
and Mixer can now rely entirely on 'docker run' to pull the tagged image
directly from DockerHub.

To facilitate this change, a new "DOCKER_IMAGE_PATH" parameter has been
added to the config, with a default value of "clearlinux/mixer". If in
the future a different registry is needed, the value of this can be
changed (e.g, to "newpaidregistry.com:5000/clearlinux/mixer").

This patch pulls out all of the content image fetching, validation, and
staleness checking code, as well as the Dockerfile creation and 'docker
run' code that is no longer needed.


builder/config.go
```diff
@@ -286,6 +289,7 @@ func (config *MixConfig) legacyParse(filename string) error {
 		{`^LOCAL_BUNDLE_DIR\s*=\s*`, &config.Mixer.LocalBundleDir, false},
 		{`^LOCAL_REPO_DIR\s*=\s*`, &config.Mixer.LocalRepoDir, false},
 		{`^LOCAL_RPM_DIR\s*=\s*`, &config.Mixer.LocalRPMDir, false},
+		{`^DOCKER_IMAGE_PATH\s*=\s*`, &config.Mixer.LocalRPMDir, false},
```
tmarcu: Copy-pasta error?

kevincwells: got lost in my poorly-done rebase effort. Should be `&config.Mixer.DockerImgPath`

tmarcu: *nods*

kevincwells: Finally pushed a fix of this. Sorry for the delay.

kevincwells:
Note: If it's not obvious already, this change only works in conjunction with an in-development change to internal DevOps scripts that do the build and dockerhub push.

PR #317, opened 2018-05-23
mixin: fix local repo bug

### LINKED ISSUE #309: Enable local RPMs in mixin
Mixin does not initialize the local RPM repo when a user has added an rpm to the local-rpms directory. This should be done before adding a package.

matthewrsj:
Fixes #309
Add the local RPMs before checking if they exist via getPackageRepo. If
the .yum-mix.conf does not exist yet initialize this file so the local
repo is configured.

### DIFF
--- a/mixin/package.go
+++ b/mixin/package.go
@@ -110,6 +110,21 @@ func addPackage(pkg string, build bool) (string, error) {
 	b.NumBundleWorkers = runtime.NumCPU()
 	b.NumFullfileWorkers = runtime.NumCPU()
 
+	rpms, err := helpers.ListVisibleFiles(b.Config.Mixer.LocalRPMDir)
+	if err != nil {
+		return "", err
+	}
+
+	if len(rpms) > 0 {
+		// initialize local repo automatically
+		runInitRepo(&cobra.Command{}, []string{})
+
+		err = b.AddRPMList(rpms)
+		if err != nil {
+			return "", err
+		}
+	}
+
 	bundle, err = getPackageRepo(pkg, ver, b.Config.Builder.DNFConf)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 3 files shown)


mixin/repo.go
```diff
@@ -178,6 +178,13 @@ func runInitRepo(cmd *cobra.Command, args []string) {
...
 
+	// save working directory so we can come back,
+	// this function may be called other than by commandline
+	wd, err := os.Getwd()
+	if err != nil {
+		fail(err)
+	}
+
 	err = os.Chdir(mixWS)
 	if err != nil {
```
icarus-sparry: Do we want a
                defer func { _ = os.Chdir(wd)
here, rather than an explicit chdir at the end?

mixin/package.go
```diff
@@ -110,27 +110,32 @@ func addPackage(pkg string, build bool) (string, error) {
...
 	}
 
-	err = appendToFile(filepath.Join(mixWS, "local-bundles", bundle), fmt.Sprintf("%s\n", pkg))
+	bundle, err = getPackageRepo(pkg, ver, b.Config.Builder.DNFConf)
 	if err != nil {
 		return "", err
 	}
 
-	rpms, err := helpers.ListVisibleFiles(b.Config.Mixer.LocalRPMDir)
+	err = b.EditBundles([]string{bundle}, true, true, false)
```
icarus-sparry: How is anyone expected to figure out this api just from examples?

Perhaps instead of the 3 bool flags it could take a struct with 3 bool flags so at least they get names?

Perhaps have a helper routine "EditBundlesWithoutEditor" which passes true as the first flag?

Go the whole https://commandcenter.blogspot.com.au/2014/01/self-referential-functions-and-design.html https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis route?

matthewrsj: I agree that the api for that sucks, but since I just moved the call and didn't change it I'd rather not do the legwork of fixing the API in this PR.

matthewrsj:
@icarus-sparry I addressed your review and added a commit to fix something else.

PR #321, opened 2018-05-24
config-unify part1: Base changes

rchiossi:
This PR is the first part of the config unification patch and includes all the pre-work required to remove the extra config files. With those changes in, the remaining patches can be sent as independent PRs as each one only deals with removing one specific config file (e.g. server.ini). This also is the part that requires the most work rebasing, so once they are in, working on further revisions for the rest should be much easier.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -35,6 +35,7 @@ import (
 	"text/tabwriter"
 	textTemplate "text/template" // "template" conflicts with crypto/x509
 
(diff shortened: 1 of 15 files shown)


rchiossi:
@matthewrsj Can you take a look at the changes to mixin?

matthewrsj:
@rchiossi I tested mixin with your changes and everything seems to be working correctly.

matthewrsj:
That's a good change for mixin btw.

PR #322, opened 2018-05-25
Create SERVER_STATE_DIR with uid/gid of parent

kevincwells:
This patch changes the way `SERVER_STATE_DIR` is created if it does not already exist, in order to facilitate mounting the directory inside the Docker container.

By default, `SERVER_STATE_DIR` in `builder.conf` is set to `PWD + "/update"`, a path that typically does not already exist and is created automaticallly the first time mixer builds bundles. Historically, `mixer build bundles` was run as root; now it is run as root within a Docker container. In either case, this `update` directory ends up owned by root.

For security reasons, mixer will not mount to the container any directory to which the caller does not have read/write permissions. This means that on subsequent calls, mixer refuses to mount the `update` directory it just created, as it is owned by root.

This patch resolves this for most users by creating the `SERVER_STATE_DIR` with the uid/gid of its closest existant ancester. In the default case, this is the uid/gid of the `PWD`. Users with existing mixes will need to manually `chown` the `SERVER_STATE_DIR` to a uid/gid they have read/write permission to.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -32,6 +32,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"syscall"
 	"text/tabwriter"
 	textTemplate "text/template" // "template" conflicts with crypto/x509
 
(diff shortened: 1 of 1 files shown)


kevincwells:
For reference, without this patch, `SERVER_STATE_DIR` typically gets created by `bundles.go` L616 at the time the `bundleDir` is created:

```go
	bundleDir := filepath.Join(b.Config.Builder.ServerStateDir, "image")

	// Bootstrap the directories.
	err = os.MkdirAll(filepath.Join(bundleDir, "0"), 0755)
```

builder/builder.go
```diff
@@ -1427,6 +1428,31 @@ func (b *Builder) NewDNFConfIfNeeded() error {
...
+	gid := int(fi.Sys().(*syscall.Stat_t).Gid)
+	return uid, gid, nil
+}
+
+// getClosestExistingAncestor returns the longest substring of path that is an
+// existing path on the filesystem. If path itself already exists, path is
+// returned unchanged.
+func getClosestExistingAncestor(path string) (string, error) {
+	if _, err := os.Stat(path); os.IsNotExist(err) {
+		path = filepath.Dir(path)
```
rchiossi: Wouldn't it make more sense to have this outside the if statement? The name implies that you are getting an ancestor, so even if path exists, this should return its ancestor, not itself. Also, this would reduce one stat() call in the overall execution, since right now stat() is called twice on ServerStateDir

kevincwells: @rchiossi good point. I wrote the function before I wrote how I was using it, and you're right -- as is there's no good reason for it to be in this order, 'cause it's only ever being called on cases where the full path does not exist.

kevincwells: Heck, I can even just combine my two functions to reduce another stat, and just return the uid/gid of the closest parent. Maaaaybe someday someone will want to use either of these bits of functionality separately, but that need doesn't exist now.

PR #327, opened 2018-05-25
No default bundles

### LINKED ISSUE #300: add '--default-bundles' flag to 'mixer init'
Current State
------------------

Currently, `mixer init` does one of three things:
1. Under normal circumstances, adds the four default bundles (os-core, os-core-update, bootloader, and kernel-native) to your mix.
1. If `--all-upstream` is passed, adds all upstream bundles to your mix. (And/or the niche `--all-local`, for all local bundles, if you already have them at init time.)
1. If the `mixbundles` file already exists, does nothing.

This behavior means people have resorted to creating an empty `mixbundles` file before running `mixer init` just to avoid having the default bundles added. Other users accidentally (or intentionally) had a `mixbundles` file, and expected `mixer init` to add those bundles anyway.

Proposal
------------
Add a new `--no-default-bundles` flag to `mixer init` that defaults to false. If false/omitted, _always_ add those default bundles to the mix bundle list, even if the file already exists. If true/passed, do not modify the list. Behavior of `--all-upstream` would remain unchanged.

kevincwells:
Fixes #300. This has been asked for by numerous people. Even our own BATS tests were using the hack.

Note: Updated the `mixin` call `b.InitMix` to pass `true` for this variable, so it still ends up with only the `os-core` bundle.

---------------

Previously `mixer init` would add four default bundles to your `mixbundles` list, unless it already existed. This resulted in confusing behavior. Some users resorted to creating an empty `mixbundles` list before running 'mixer init' in order to avoid these bundles. Others had a `mixbundles` list already, but expected the bundles to be added anyway. 

This patch changes the behavior to **always** add the four default bundles, even if the `mixbundles` list already exists, unless a new `--no-default-bundles` flag is passed. This does not affect the behavior of `--all-local` or `--all-upstream`.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -194,7 +194,7 @@ mix-bundles/`
 
 // InitMix will initialise a new swupd-client consumable "mix" with the given
 // based Clear Linux version and specified mix version.
-func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allUpstream bool, upstreamURL string, git bool) error {
+func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allUpstream bool, noDefaults bool, upstreamURL string, git bool) error {
 	// Set up local dirs
 	if err := b.initDirs(); err != nil {
 		return err
@@ -253,12 +253,12 @@ func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allU
 	}
 
 	// Initialize the Mix Bundles List
-	if _, err := os.Stat(filepath.Join(b.Config.Builder.VersionPath, b.MixBundlesFile)); os.IsNotExist(err) {
-		// Add default bundles (or all)
-		defaultBundles := []string{"os-core", "os-core-update", "bootloader", "kernel-native"}
-		if err := b.AddBundles(defaultBundles, allLocal, allUpstream, false); err != nil {
-			return err
-		}
+	var bundles []string
+	if !noDefaults {
+		bundles = []string{"os-core", "os-core-update", "bootloader", "kernel-native"}
+	}
+	if err := b.AddBundles(bundles, allLocal, allUpstream, false); err != nil {
+		return err
 	}
 
 	// Get upstream bundles
--- a/mixin/package.go
+++ b/mixin/package.go
@@ -107,6 +107,10 @@ func addPackage(pkg string, build bool) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	err = b.AddBundles([]string{"os-core"}, false, false, false)
+	if err != nil {
+		return "", err
+	}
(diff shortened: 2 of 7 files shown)


kevincwells:
> @rchiossi  I wonder if we should keep the default packages as a list somewhere else rather than inline in the middle of the code.

Yeah, good idea. At least right now it's still only defined in one place, but that place is definitely obscure. Where do you think it oughta live?

rchiossi:
If we have more default values like that in the code, it might be worth it centralizing them on a defaults files. If the number is not relevant, just defining them as globals in the file they are used should be enough to make it a bit more organized.

mixin/package.go
```diff
@@ -103,7 +103,7 @@ func addPackage(pkg string, build bool) (string, error) {
 		return "", err
 	}
 	err = b.InitMix(fmt.Sprintf("%d", ver), fmt.Sprintf("%d", mixVer),
-		false, false, "https://download.clearlinux.org", false)
+		false, false, true, "https://download.clearlinux.org", false)
```
matthewrsj: Can you add "os-core" after this point and remove the `ioutil.WriteFile("mixbundles")` from `mixin/helpers.go:setUpMixDir` that was designed to avoid the defaultbundles additions?

builder/builder.go
```diff
@@ -194,7 +194,7 @@ mix-bundles/`
 
 // InitMix will initialise a new swupd-client consumable "mix" with the given
 // based Clear Linux version and specified mix version.
-func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allUpstream bool, upstreamURL string, git bool) error {
+func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allUpstream bool, noDefaults bool, upstreamURL string, git bool) error {
```
matthewrsj: Eww another bool to the function signature. Maybe this should take a struct with all the bools that are needed? This API is very difficult to use...

/me channels his inner @icarus-sparry

kevincwells: Yeah, it's gross. But it was already gross before this addition. That's a bigger redesign than I want to do in a small feature PR like this.

matthewrsj: fiiiiiiiiine :)

icarus-sparry: The question ought to be "how many of the 16 possible combinations of bool make sense" and unless it is "most" then this ought to be cleaned up. Fair enough not to do it in this PR.

 If you are in an environment where you have say godoc support then it is sort of OK, but if all you see is function(T,T,F,T) then it is hard to reason about what the 3 true and one false is selecting.

mixin/package.go
kevincwells: @matthewrsj  just pushed a new commit that I think is what you're asking for.

PR #330, opened 2018-05-30
Add auto format bump support into mixer

tmarcu:
This PR introduces automatic format-bump support and containerized mixer builds.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -34,6 +34,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"text/tabwriter"
 	textTemplate "text/template" // "template" conflicts with crypto/x509
 
@@ -1403,6 +1404,22 @@ func (b *Builder) NewDNFConfIfNeeded() error {
 	return nil
 }
 
+// getClosestAncestorOwner returns the owner uid/gid of the closest existing
+// ancestor of the file or dir pointed to by path. There is no fully cross-
+// platform concept of "ownership", so this method only works on *nix systems.
+func getClosestAncestorOwner(path string) (int, int, error) {
+	path = filepath.Dir(path)
+	fi, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		return getClosestAncestorOwner(path)
+	} else if err != nil {
+		return 0, 0, err
+	}
+	uid := int(fi.Sys().(*syscall.Stat_t).Uid)
+	gid := int(fi.Sys().(*syscall.Stat_t).Gid)
+	return uid, gid, nil
+}
+
 // BuildBundles will attempt to construct the bundles required by generating a
 // DNF configuration file, then resolving all files for each bundle using dnf
 // resolve and no-op installs. One full chroot is created from this step with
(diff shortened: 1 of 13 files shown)


mixer/cmd/build.go
```diff
@@ -149,13 +150,52 @@ var buildUpstreamFormatCmd = &cobra.Command{
...
 			fail(err)
 		}
+
+		bumpNeeded, err := b.CheckBumpNeeded(true)
+		if err != nil {
+			fail(err)
+		}
+		// Loop through multiple bumps until we reach the latest format
+		for bumpNeeded {
+			cmdToRun = strings.Split("mixer build format-bump new", " ")
```
kevincwells: Instead of copypasting the entire function, why not just wrap the entire original function in the bump check loop?  The premise is that it's already been called and the mix has already been staged by the time it gets here, so the initial condition of the bool can just be true.

I haven't verified this code whatsoever, but something like:

```go
 		b, err := builder.NewFromConfig(configFile)
 		if err != nil {
 			fail(err)
 		}

 		bumpNeeded := true

 		// Loop through multiple bumps until we reach the latest format
 		for bumpNeeded {
			cmdToRun := strings.Split("mixer build format-bump new", " ")
	 		if config.UseNewConfig {
	 			cmdToRun = append(cmdToRun, "--new-config")
	 		}
	 		if err := b.RunCommandInContainer(cmdToRun); err != nil {
	 			fail(err)
	 		}
	 
	 		// Set the upstream version to the previous format's latest version
	 		b.UpstreamVerUint32 -= 10
	 		b.UpstreamVer = strconv.FormatUint(uint64(b.UpstreamVerUint32), 10)
	 		vFile := filepath.Join(b.Config.Builder.VersionPath, b.UpstreamVerFile)
	 		if err := ioutil.WriteFile(vFile, []byte(b.UpstreamVer), 0644); err != nil {
	 			fail(err)
	 		}
	 		cmdToRun = strings.Split("mixer build format-bump old", " ")
	 		if config.UseNewConfig {
	 			cmdToRun = append(cmdToRun, "--new-config")
	 		}
	 		if err := b.RunCommandInContainer(cmdToRun); err != nil {
	 			fail(err)
	 		}
	 		// Set the upstream version back to what the user originally tried to build
	 		if err := b.UnstageMixFromBump(); err != nil {
	 			fail(err)
	 		}
	 		bumpNeeded, err = b.CheckBumpNeeded(true)
			if err != nil {
				fail(err)
			}
 		}
	}
```

builder/builder.go
```diff
@@ -2260,7 +2260,7 @@ New upstream: %d (format: %s)
 	b.UpstreamVerUint32 = nextUpstream
 	b.UpstreamVer = nextUpstreamStr
 
-	if _, err := b.CheckBumpNeeded(); err != nil {
+	if _, err := b.CheckBumpNeeded(false); err != nil {
```
kevincwells: Erm.. Perhaps I'm misunderstanding, but if you call this with `false` here, the user will not be warned that their new state is bad and they need to fix it. Is this a deliberate choice, pushing the warning to the next time they try to do a build?

builder/format.go
```diff
@@ -152,7 +152,7 @@ func (b *Builder) UnstageMixFromBump() error {
 
 // CheckBumpNeeded returns nil if it successfully deduces there is no format
 // bump boundary being crossed.
-func (b *Builder) CheckBumpNeeded() (bool, error) {
+func (b *Builder) CheckBumpNeeded(loop bool) (bool, error) {
```
kevincwells: I don't love this approach, as it a) makes the calling sites more confusing, and b) makes _any_ place that uses this function (e.g., the root `PreRun`) have to think about "looping", when "looping" is only a concept that makes sense for `mixer build upstream-format`.

At minimum I'd want this variable changed to `warn` or `silent` or something, or you could make `CheckBumpNeededSilent` do everything this function does except the warning, and then make `CheckBumpNeeded` just call the silent version and then print a warning before returning.

tmarcu: Sure I can rename it.

mixer/cmd/build.go
tmarcu: This is a better place to put it for sure!

kevincwells:
I think with the most recent change, this is safe to merge to master.

Before this is officially _released_ (i.e., command un-hidden and `--native` default to false), AFB still needs:
- Documentation
  - Good, thorough `--help` messages for the new commands.
  - Updated mixer docs
  - Updated man pages
- Complete, working docker builds integrated into devops flow
- Docker dependency added
  - To build commands dependency list? What about if building natively?
  - To actual `mixer` bundle?

PR #331, opened 2018-05-30
Parallel delta creation

tmarcu:
~This needs to pass make check, opening PR for early review of this new - albeit naive - first pass at re-implementing and parallelizing our delta pack creation.~

In my testing, building delta packs back from 22460 -> 22530 took about 60 minutes with the original code.
With this patch the time is reduced to 5m30s, but there is 1 missing delta and thus 1 added fullfile:
/usr/share/vim/vim81/lang/menu_ja.euc-jp.vim

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -2000,7 +2000,9 @@ func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printRepor
 				fmt.Printf("  Creating delta pack for bundle %q from %d to %d\n", b.Name, b.FromVersion, b.ToVersion)
 				info, err := swupd.CreatePack(b.Name, b.FromVersion, b.ToVersion, outputDir, bundleDir, numWorkers)
 				if err != nil {
-					// set some error
+					fmt.Fprintf(os.Stderr, "ERROR: Pack %q from %d to %d FAILED to be created: %s\n", b.Name, b.FromVersion, b.ToVersion, err)
+					// Do not exit on errors, we have logging for all other failures and deltas are optional
+					continue
 				}
 
 				if len(info.Warnings) > 0 {
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -1950,13 +1950,13 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
 	return nil
 }
 
-func createDeltaPacks(from *swupd.Manifest, to *swupd.Manifest, printReport bool, outputDir, bundleDir string, numWorkers int) error {
+func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printReport bool, outputDir, bundleDir string, numWorkers int) error {
 	timer := &stopWatch{w: os.Stdout}
 	defer timer.WriteSummary(os.Stdout)
 	timer.Start("CREATE DELTA PACKS")
 
-	fmt.Printf("Creating delta packs from %d to %d\n", from.Header.Version, to.Header.Version)
-	bundlesToPack, err := swupd.FindBundlesToPack(from, to)
+	fmt.Printf("Creating delta packs from %d to %d\n", fromMoM.Header.Version, toMoM.Header.Version)
+	bundlesToPack, err := swupd.FindBundlesToPack(fromMoM, toMoM)
 	if err != nil {
 		return err
 	}
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -27,9 +27,11 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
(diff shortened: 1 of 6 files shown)


tmarcu:
Updated test suite appropriately. 

Remaining task is to write the bsdiff failures out to a log file so we can investigate them, but we still should not fail on any delta creation/apply errors.

builder/builder.go
```diff
@@ -1952,44 +1969,74 @@ func createDeltaPacks(from *swupd.Manifest, to *swupd.Manifest, printReport bool
...
 		if !os.IsNotExist(err) {
 			return errors.Wrapf(err, "couldn't access existing pack file %s", packPath)
 		}
-		fmt.Printf("  Creating delta pack for bundle %q from %d to %d\n", bp.Name, bp.FromVersion, bp.ToVersion)
-		info, err := swupd.CreatePack(bp.Name, bp.FromVersion, bp.ToVersion, outputDir, bundleDir, numWorkers)
-		if err != nil {
-			return err
-		}
+	}
+	// Cap the routines for bundles to the number of bundles
```
matthewrsj: Aren't you capping it at NumCPU not number of bundles?

builder/builder.go
```diff
@@ -1952,44 +1969,74 @@ func createDeltaPacks(from *swupd.Manifest, to *swupd.Manifest, printReport bool
...
+	var bundleWorkers int
+	if numWorkers < 1 {
+		numWorkers = 1
+		bundleWorkers = 1
+	} else {
+		bundleWorkers = runtime.NumCPU()
```
matthewrsj: so if numWorkers == 5 you still set bundleWorkers to NumCPU?

builder/builder.go
```diff
@@ -1952,44 +1969,74 @@ func createDeltaPacks(from *swupd.Manifest, to *swupd.Manifest, printReport bool
...
-				fmt.Printf("    WARNING: %s\n", w)
-			}
-			fmt.Println()
-		}
-		if printReport {
-			max := 0
-			for _, e := range info.Entries {
-				if len(e.File.Name) > max {
-					max = len(e.File.Name)
+	var bundleQueue = make(chan *swupd.BundleToPack)
```
matthewrsj: Why not buffer this to the number for bundleWorkers you have?

swupd/hash.go
```diff
@@ -34,16 +35,26 @@ var AllZeroHash = "0000000000000000000000000000000000000000000000000000000000000
...
 var Hashes = []*string{&AllZeroHash}
 var invHash = map[string]Hashval{AllZeroHash: 0}
+var rwMutex sync.RWMutex
 
 // internHash adds only new hashes to the Hashes slice and returns the index at
 // which they are located
 func internHash(hash string) Hashval {
+	rwMutex.RLock()
 	if key, ok := invHash[hash]; ok {
+		rwMutex.RUnlock()
```
matthewrsj: This block can be:
```go
key, ok := invHash[hash]
rwMutex.Runlock()
if ok {
    return key
}
rwMutex.Lock()
...
```
So you don't have the multiple calls to `RUnlock` and you perform the unlock before checking your conditional.

swupd/hash.go
```diff
@@ -34,16 +35,26 @@ var AllZeroHash = "0000000000000000000000000000000000000000000000000000000000000
...
+		return key
+	}
+	rwMutex.RUnlock()
+	rwMutex.Lock()
+	if key, ok := invHash[hash]; ok {
```
matthewrsj: Wait, isn't this the same check as above but using a `Lock` instead of an `RLock`?

swupd/packs.go
```diff
@@ -74,6 +75,44 @@ func (state PackState) String() string {
...
+	if err != nil {
+		return err
+	}
+
+	if fromVersion >= toVersion {
+		return fmt.Errorf("fromManifest version (%d) must be smaller than toManifest version (%d)", fromVersion, toVersion)
+	}
+
+	var c config
+	c, err = getConfig(filepath.Join(outputDir, ".."))
```
matthewrsj: Looks like you can just use `getConfig(filepath.Dir(outputDir))` since we aren't creating `outputDir` with a trailing slash... or better safe than sorry.

builder/builder.go
tmarcu: Yes, ty for pointing it out, I actually want to cap at NumCPU() always unless 0 is passed in specifically, since it's no longer creating deltas as part of the tar pack creation.

swupd/hash.go
tmarcu: Yes

tmarcu: Adding a comment to explain why we're doing this twice 'cause it is actually confusing..

icarus-sparry: And it is essential that we do this check a second time.

We could go for lower performance and just Lock() everywhere, rather than using an RWLock to avoid the second check, but it would be a poor tradeoff unless we think that we are only going to look up keys once.

PR #332, opened 2018-05-31
Do not copy deleted bundles across a minversion

matthewrsj:
If the bundle was not generated anew for the new minversion do not try
to copy a reference to it from the old MoM. This indicates a bundle was
deleted and the old build may be cleaned up.

Sooo... OSVs shouldn't be deleting bundles over a minversion if it isn't associated with a format bump, and the failure that causes in mixer can help catch that. but I think that check should live elsewhere, not in mixer, and mixer should still do the right thing by dropping the bundle reference here. Thoughts @tmarcu?

### DIFF
--- a/swupd/create_manifests_test.go
+++ b/swupd/create_manifests_test.go
@@ -366,6 +366,28 @@ func TestCreateManifestsMinVersion(t *testing.T) {
 	ts.checkNotContains("www/20/Manifest.full", "\t10\t")
 }
 
+func TestCreateManifestsMVDeletes(t *testing.T) {
+	ts := newTestSwupd(t, "minVersionDeletes")
+	defer ts.cleanup()
+
+	ts.Bundles = []string{"test-bundle"}
+	ts.addFile(10, "test-bundle", "/foo", "foo")
+	ts.createManifests(10)
+	ts.checkContains("www/10/Manifest.test-bundle", "10\t/foo\n")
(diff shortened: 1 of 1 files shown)


tmarcu:
This is the right way to fix it so manifests are created correctly, but I don't think we should do this. I'd rather it fail hard because it hides a user error on building mixes -> removing bundles without a format bump. If the bundle isn't getting generated for the minversion, something has gone wrong so instead of **continue** we can stop and print an error right there explaining bundles can't be randomly removed. 
That's my $.02 :) 
Where else can we move the check otherwise? I don't want to drop the reference and continue 'cause then systems have leftover files that shouldn't be there anymore, and that can cause a lot of other issues if things clash on later updates.

matthewrsj:
> we can stop and print an error right there explaining bundles can't be randomly removed.

It isn't as easy as that. This is a really weird situation where we have a bundle that was removed, not associated with a format bump *and* the old content was removed from the builder. It will still create the manifest if the bundle was removed but the old content wasn't cleaned up... At least with this patch we are making valid manifests. Like I said I think the check for this case should be done elsewhere. In Clear Linux it now *is* done elsewhere.

> Where else can we move the check otherwise?

Again, difficult. We don't know until this point what the update actually looks like, so we don't know if this bundle got an update yet or not. We may be able to do some pre-processing at the beginning but I haven't given that much thought yet tbh.

PR #334, opened 2018-05-31
bundles: write bundle info to tracking file

### LINKED ISSUE #333: Adding more bundle metadata to tracking files
It may be nice for swupd users to query specific information about a bundle. For example, they may want to know the status of a bundle, or its description, maintainer, etc. It could be implemented in swupd-client as a new subcommand (maybe `swupd bundle-info BUNDLE`).

To enable this swupd-client feature, mixer could add extra metadata to the bundle tracking files in `/usr/share/clear/bundles`, and later `swupd bundle-info` could start reading this metadata.

matthewrsj:
Fixes #333
Write the bundle information JSON to the bundle tracking file under
/usr/share/clear/bundles/<bundle-name> for future use by a swupd-client
bundle-info command. Write it in a human-readable format for now until
the bundle-info command is usable by writing indents with
json.MarshalIndent.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -520,7 +520,7 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
 		return err
 	}
 
-	return ioutil.WriteFile(filepath.Join(bundleDir, bundle.Name), nil, 0644)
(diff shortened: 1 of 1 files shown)


matthewrsj:
Doing it this way means every bundle gets a size increase. Converting this to only add these files for ALL bundles to os-core-update-index, an optional meta bundle.

PR #336, opened 2018-05-31
os-core-update-index: add bundle metadata files

matthewrsj:
Add bundle information files under /usr/share/clear/allbundles to allow
clients to perform future bundle-info operations. Write with 'pretty'
indented JSON so it is grep-able and human-readable in the mean-time.
Users will have to install the os-core-update-index bundle to get this
functionality.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -520,7 +520,18 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
 		return err
 	}
 
-	return ioutil.WriteFile(filepath.Join(bundleDir, bundle.Name), nil, 0644)
+	err = ioutil.WriteFile(filepath.Join(bundleDir, bundle.Name), nil, 0644)
+	if err != nil {
+		return nil
+	}
+
+	metaPath := filepath.Join(baseDir, "usr/share/clear/allbundles")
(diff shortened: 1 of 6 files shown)


matthewrsj:
The `os-core-update-index` bundle becomes about 387 MB for Clear version 22690 with this change. Of course, once installed users will get delta updates.

matthewrsj:
@tmarcu would also be nice to get this one into the next release if possible, but if you have review comments it can wait.

builder/bundles.go
```diff
@@ -586,6 +597,15 @@ func buildFullChroot(cfg *buildBundlesConfig, b *Builder, set *bundleSet, packag
 	return nil
 }
 
+func writeBundleInfoPretty(bundle *bundle, path string) error {
+	b, err := json.MarshalIndent(*bundle, "", "	")
```
icarus-sparry: Minor suggestion, use \t rather than a literal tab character. After your comment about the size of the file I saw this line in the diff and thought "I wonder if that is a number of spaces, and that makes the size so large?"

matthewrsj: Ah, yes I'll make it a \t

matthewrsj: done.

PR #339, opened 2018-06-06
mixin: add --bundle flag for optional bundle name

matthewrsj:
When adding a package provide an optional --bundle flag so users can set
which bundle their package is added to instead of automatically adding
it to a bundle named after the repo the package came from.

### DIFF
--- a/mixin/package.go
+++ b/mixin/package.go
@@ -55,7 +55,8 @@ var packageCmds = []*cobra.Command{
 }
 
 type packageAddCmdFlags struct {
-	build bool
+	bundle string
+	build  bool
 }
 
 var packageAddFlags packageAddCmdFlags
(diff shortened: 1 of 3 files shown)


matthewrsj:
@bktan8 there ya go.

mixin/package.go
```diff
@@ -134,6 +136,12 @@ func addPackage(pkg string, build bool) (string, error) {
 		return "", err
 	}
 
+	// if the bundleName argument is non-empty use this instead of
```
icarus-sparry: Perhaps getPackageRepo should be renamed getPackageFromRepo, when I saw this code I was thinking we could name the passed in parameter as "bundle" and only call getPackageRepo if we needed a name (i.e. were passed in a empty string) until I checked.

PR #342, opened 2018-06-08
Parallelize version back for deltas

tmarcu:
Run each version back for delta packs in its own thread, and divide the total number of CPUs available and use that in each version thread.
There are cases where only a few files remain to delta, and they take multiple minutes to finish. Each version runs synchronously with many threads, so at the end of a version we have to wait for the long delta to finish before running through another. This changes it so by the end of the run - all the large/slow files are building at the same time, capping it to just one long file run, vs long run * number of versions back.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -2015,19 +2015,53 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
 		cur = m.Header.Previous
 	}
 
-	fmt.Printf("Using %d workers\n", b.NumDeltaWorkers)
 	fmt.Printf("Found %d previous versions\n", len(previousManifests))
 
 	bundleDir := filepath.Join(b.Config.Builder.ServerStateDir, "image")
 	// Create all deltas for all previous versions first based on full manifests
-	for _, fromManifest := range previousManifests {
-		fmt.Println()
-		err = swupd.CreateAllDeltas(outputDir, int(fromManifest.Header.Version), int(toManifest.Header.Version), b.NumDeltaWorkers)
-		if err != nil {
-			return err
-		}
+	var versionQueue = make(chan *swupd.Manifest)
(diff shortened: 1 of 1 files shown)


builder/builder.go
```diff
@@ -2015,19 +2015,53 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
...
-			return err
-		}
+	var versionQueue = make(chan *swupd.Manifest)
+	var wg sync.WaitGroup
+	var deltaErrors []error
+	versionWorkers := 1
+
+	// If we have at least 2x the number of CPUs as versions, give each version
+	// its own thread to build deltas in.
+	if runtime.NumCPU() >= 2*len(previousManifests) {
```
matthewrsj: Are we ignoring what the user sets then?

tmarcu: Yes in this case until we get better stats on other combinations. It will use their number however for the case that they don't have enough CPUs, which is fine 'cause it won't run multiple version thread pools at least (imo).

builder/builder.go
```diff
@@ -2015,19 +2015,53 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
...
+		versionQueue <- previousManifests[i]
 	}
-	// Create any missing delta files and pack all deltas up
+
+	// Send message that no more jobs are being sent
+	close(versionQueue)
+	wg.Wait()
+
+	for i := 0; i < len(deltaErrors); i++ {
+		fmt.Fprintf(os.Stderr, "%s", deltaErrors[i])
```
phmccarty: Should the format string be `"%s\n"` so newlines are added?

phmccarty:
Would be helpful if you add the PR description to the commit as well.

builder/builder.go
tmarcu: YES ty

PR #344, opened 2018-06-11
Fix bug for user who don't have upstreamurl file

tmarcu:
This is the case for when an OSV *is upstream* and does not have a set
upstreamurl. There is nothing to download and figure out the format range
for since they set the bounds manually, so don't crash on this error unless
it is running in a container non-natively.


tmarcu:
I was only able to reproduce the "no protocol found" bug by removing my upstreamurl file, and this fixes that specific scenario so the build can continue.

PR #346, opened 2018-06-12
Set timeout on bsdiff delta creation

tmarcu:
Set timeout to 1 minute for bsdiff. The majority of all delta creations
take significantly less than 1 minute to run, which signifies that bsdiff
is working on a delta that may be very large, or very difficult to diff. In
all the cases where bsdiff took multiple minutes to finish, the delta ended
up not being used because it was larger than the compressed fullfile. This
attempts to skip those cases and improve delta creation time.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -787,9 +787,7 @@ src=%s
 // there are no file collisions in the build.
 func createOsPackagesFile(buildVersionDir string) error {
 	fullChroot := filepath.Join(buildVersionDir, "full")
-	packages, err := helpers.RunCommandOutput(
-		"rpm", "--root="+fullChroot, "-qa", "--queryformat", "%{NAME}\t%{SOURCERPM}\n",
-	)
+	packages, err := helpers.RunCommandOutput("rpm", "--root="+fullChroot, "-qa", "--queryformat", "%{NAME}\t%{SOURCERPM}\n")
 	if err != nil {
 		return err
 	}
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -18,6 +18,7 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"context"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
@@ -282,6 +283,27 @@ func RunCommandSilent(cmdname string, args ...string) error {
 	return err
 }
 
+// RunCommandTimeout runs the given command with timeout + args and does not print command output
+func RunCommandTimeout(timeout int, cmdname string, args ...string) error {
+	ctx := context.Background()
+	// 0 means infinite timeout, ONLY set timeouts when value is > 0
+	if timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(context.Background(), time.Duration(timeout)*time.Second)
+		defer cancel()
+	}
+	cmd := exec.CommandContext(ctx, cmdname, args...)
+	cmd.Stdout = nil
+	cmd.Stderr = nil
+	err := cmd.Run()
+
+	if ctx.Err() == context.DeadlineExceeded {
+		errors.Errorf("Command: %s timed out\n", cmdname)
+	}
+
+	return err
+}
+
 // RunCommandOutput executes the command with arguments and stores its output in
 // memory. If the command succeeds returns that output, if it fails, return err that
 // contains both the out and err streams from the execution.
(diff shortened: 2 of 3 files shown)


builder/bundles.go
```diff
@@ -200,7 +200,8 @@ func resolveFilesForBundle(bundle *bundle, repoPkgs map[string][]string, package
 		for _, p := range pkgs {
 			queryString = append(queryString, p)
 		}
-		outBuf, err := helpers.RunCommandOutput(queryString[0], queryString[1:]...)
+		timeout := 0
+		outBuf, err := helpers.RunCommandOutput(timeout, queryString[0], queryString[1:]...)
```
matthewrsj: Just pass `0` here instead of declaring `timeout`?

helpers/helpers.go
```diff
@@ -278,20 +279,40 @@ func RunCommandInput(in io.Reader, cmdname string, args ...string) error {
...
+	_, err := RunCommandOutput(timeout, cmdname, args...)
 	return err
 }
 
 // RunCommandOutput executes the command with arguments and stores its output in
 // memory. If the command succeeds returns that output, if it fails, return err that
 // contains both the out and err streams from the execution.
-func RunCommandOutput(cmdname string, args ...string) (*bytes.Buffer, error) {
-	cmd := exec.Command(cmdname, args...)
+func RunCommandOutput(timeout int, cmdname string, args ...string) (*bytes.Buffer, error) {
```
matthewrsj: not a huge fan of putting the timeout at the beginning of the args list, but I suppose you need to put it before `cmdname string, args ...string`

builder/bundles.go
```diff
@@ -371,7 +372,8 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 			// status. This exit status is 1, which is the same as any other dnf install
 			// error. Fortunately if this is a different error than we expect, it should
 			// fail in the actual install to the full chroot.
-			outBuf, _ := helpers.RunCommandOutput(queryString[0], queryString[1:]...)
+			timeout := 0
+			outBuf, _ := helpers.RunCommandOutput(timeout, queryString[0], queryString[1:]...)
```
matthewrsj: again, don't see the point of declaring a variable to immediately use and discard

builder/bundles.go
```diff
@@ -787,7 +789,8 @@ src=%s
 // there are no file collisions in the build.
 func createOsPackagesFile(buildVersionDir string) error {
 	fullChroot := filepath.Join(buildVersionDir, "full")
-	packages, err := helpers.RunCommandOutput(
+	timeout := 0
+	packages, err := helpers.RunCommandOutput(timeout,
```
matthewrsj: ...

helpers/helpers.go
tmarcu: Yeah, unfortunately by language design I can't put it at the end.

builder/bundles.go
tmarcu: People (including me) hated on using magic numbers in these kinds of cases. Was just trying to make it more clear, but I'm thinking of how to re-write.

helpers/helpers.go
phmccarty: What about `cmdname string, timeout int, args ...string)` ? That seems better to me.

matthewrsj: That's more difficult to pass 'entireCmd...'

helpers/helpers.go
```diff
@@ -282,6 +283,28 @@ func RunCommandSilent(cmdname string, args ...string) error {
 	return err
 }
 
+// RunCommandTimeout runs the given command with timeout + args and does not print command output
+func RunCommandTimeout(timeout int, cmdname string, args ...string) error {
+	ctx := context.Background()
+	// 0 means infinite timeout, ONLY set timeouts when value is > 0
```
gtkramer: Maybe we want to throw an exception if the timeout is 0? Otherwise, this function call would effectively be something like RunCommandSilent.

helpers/helpers.go
```diff
@@ -282,6 +283,28 @@ func RunCommandSilent(cmdname string, args ...string) error {
...
+	if timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(context.Background(), time.Duration(timeout)*time.Second)
+		defer cancel()
+	}
+	cmd := exec.CommandContext(ctx, cmdname, args...)
+	cmd.Stdout = nil
```
gtkramer: This part here looks like it could be replaced by RunCommandSilent, but alas, the function not use CommandContext.

gtkramer:
Aside from my comments, looks pretty good :+1:

helpers/helpers.go
tmarcu: Yeah :( I don't like kind-of duplicate code but it's hard to merge them all.

helpers/helpers.go
tmarcu: :rocket:

PR #353, opened 2018-06-16
Format Bump general fixes

rchiossi:
This patch address a bunch of different issues I stumbled upon while trying to test the new format bump feature. This patch makes the feature functional again since it is broken on master.

The first two patches are just minor cleanups. The last 5 fix 5 different crashes.

### DIFF
--- a/builder/format.go
+++ b/builder/format.go
@@ -69,10 +69,7 @@ func (b *Builder) GetLastBuildVersion() (string, error) {
 
 	filename := filepath.Join(b.Config.Builder.ServerStateDir, "image/LAST_VER")
 	if lastVer, err = ioutil.ReadFile(filename); os.IsNotExist(err) {
-		// Likely the first build
-		return "", nil
-	} else if err != nil {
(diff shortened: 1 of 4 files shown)


builder/format.go
```diff
@@ -155,10 +147,12 @@ func (b *Builder) UnstageMixFromBump() error {
 func (b *Builder) CheckBumpNeeded(silent bool) (bool, error) {
 	version, err := b.getLastBuildUpstreamVersion()
 	if err != nil {
+		if os.IsNotExist(err) {
```
tmarcu: AH ok, so this is where the nil return moves to since you removed it at the top, correct?

rchiossi: yep. This is the exception case, where this error is OK. I believe it makes more sense than adding checks for lastver == "" everywhere else.

tmarcu: Yeah I agree with this. I'm testing it locally right now myself, but otherwise all the fixups look good to me.

Issue #352, opened 2018-06-15
Add --skip-packs option for 'build update'


phmccarty:
For local testing purposes, I would like to run `sudo mixer build update` without the zero packs being created, since I won't be consuming them. Maybe a `--skip-packs` option could be added?

phmccarty:
Sure, I am okay with having `--skip-fullfiles` as well.

phmccarty:
This was implemented with #355

PR #355, opened 2018-06-19
build: add skip-fullfiles and skip-packs option

rchiossi:
This patch adds the --skip-fullfiles options for the 'build' subcommand
of mixer. This command will prevent fullfiles from being generated
during 'build update'. Likewise, the --skip-packs option prevents zero
packs form being generated.

This PR addresses issue #352

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -84,6 +84,18 @@ type Builder struct {
 	UpstreamVerUint32 uint32
 }
 
+// UpdateParameters contains the configuration parameters for building an update
+type UpdateParameters struct {
+	// Minimum version used to generate delta packs
+	MinVersion int
+	// Format version used in this update
+	Format string
+	// Update latest format version and image version files to current mix
+	Publish bool
+	// Skip signing Manifest.MoM
+	SkipSigning bool
+}
+
 var localPackages = make(map[string]bool)
 var upstreamPackages = make(map[string]bool)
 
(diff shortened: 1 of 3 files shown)


mixin/build.go
```diff
@@ -161,7 +161,7 @@ func buildMix(prepNeeded bool) error {
 		_ = os.Remove(mixFlagFile)
 		return err
 	}
-	err = b.BuildUpdate("", 0, "", false, true, false)
+	err = b.BuildUpdate("", 0, "", false, true, false, false, false)
```
tmarcu: Oh man, this function signature is just ridiculous now ❌ ahah. Can we please figure out a better way to call this, or change every single parameter to a named one so it's not insane to call in a future spot? FTFFF means nothing to even me at this point 😅

matthewrsj: Maybe a struct containing the configuration for this specific command that contains all the bools necessary?

rchiossi: it is! I wasn't sure if I wanted to add to the mess or just clean it up at once. I took the easy path. ;)

Probably a better approach would be to set those properties in the builder object.

tmarcu: I don't blame you, but I like the rest of the code so if you can make it pretty in this commit that'd be awesomeeeeee!

rchiossi: I just noticed that two of those parameters (prefixflag and keepChroots)  are not even used anymore....

tmarcu: 😆 😂

icarus-sparry: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis and the associated video is worth reviewing if you want to do better than just adding a struct, although if I was doing it I would just add a struct.

mixin/build.go
```diff
@@ -161,7 +161,15 @@ func buildMix(prepNeeded bool) error {
...
-	err = b.BuildUpdate("", 0, "", false, true, false)
+	params := builder.UpdateParameters{
+		MinVersion:    0,
+		Format:        "",
+		Publish:       true,
+		SkipSigning:   false,
+		SkipFullfiles: false,
+		SkipPacks:     false,
+	}
+	err = b.BuildUpdate(params)
```
icarus-sparry: maybe
 ``` err := b.BuildUpdate(&builder.UpdateParameters{Publish: true})```
to use a literal struct, as there is just one field that is taking the non-default value in this case.

builder/builder.go
```diff
@@ -84,6 +84,16 @@ type Builder struct {
 	UpstreamVerUint32 uint32
 }
 
+// UpdateParameters contains the configuration parameters for building an update
+type UpdateParameters struct {
```
icarus-sparry: Whilst the names are very good, it wouldn't hurt to add a comment for each field.

mixin/build.go
rchiossi: ah nice! I didn't know that language feature.

builder/builder.go
rchiossi: Sounds good. I'll update

rchiossi:
Updated based on @icarus-sparry feedback.

builder/builder.go
```diff
@@ -1538,15 +1543,25 @@ func (b *Builder) BuildUpdate(prefixflag string, minVersion int, format string,
 	return nil
 }
 
-func (b *Builder) buildUpdateContent(timer *stopWatch, mixVersion uint32, minVersion uint32, format uint32, skipSigning bool) error {
+func (b *Builder) buildUpdateContent(params UpdateParameters, timer *stopWatch) error {
 	var err error
 
+	// TODO: move this to parsing configuration / parameter time.
+	// TODO: should this be uint64?
```
tmarcu: yes.. but we need to update this all over the code to be s/uint32/uint64 or just uint.

rchiossi: this TODO comment just moved along with the code below, it is not my comment.

tmarcu: No prob :) Just reiterating so everyone knows I want to change this in the code at some point.

PR #356, opened 2018-06-21
battest: Add test for format bump

rchiossi:
This patch adds temporary tests for the format bump feature. A permanent
test will require the generation of custom docker images with an updated
build of mixer. Until this tooling is available, this test provides a
minimum integrity check on the feature.

The added test checks the operation of a format bump on a real format
boundary. In addition, it checks the integrity of 4 key values:
- image/LasT_VER should point to the +20 version
- Previous version in +20 MoM should point to +0
- Previous version in +10 MoM should point to +0
- FORMAT property in builder.conf should have the +20 format

### DIFF
--- a/bat/tests/06-test-format-bump/run.bats
+++ b/bat/tests/06-test-format-bump/run.bats
@@ -1,14 +1,48 @@
 #!/usr/bin/env bats
 
 # shared test functions
-#load ../../lib/mixerlib
+load ../../lib/mixerlib
 
-#setup() {
-#  global_setup
-#}
+setup() {
+  global_setup
+}
 
-# TODO: fill out once auto format bump implemented
-#@test "Create initial mix 10" {
-#}
+# TODO: This test should run 'mixer build upstream-format' instead
+# once the tooling required to run custom test docker images for
+# is ready.
+@test "Perform upstream format bump on last format boundary" {
+  upstreamver=$(get-last-format-boundary)
+  mixer-init-stripped-down $upstreamver 10
+
+  #Create initial version in the old format
+  mixer-build-all > $LOGDIR/build_all.log
+
+  #Update to new upstream in the new format
+  mixer-upstream-update $(($upstreamver + 10)) > $LOGDIR/upstream_update.log
+
+  #Build +20
+  mixer-build-format-bump-new > $LOGDIR/build_formatbump_new.log
+
+  #Reset upstream for +10 build
+  echo -n "$upstreamver" > upstreamversion
+
+  #Build +10
+  mixer-build-format-bump-old > $LOGDIR/build_formatbump_old.log
+
+  #restore upstrean
+  mv upstreamversion.bump upstreamversion
+
+  #check if LAST_VER is set to +20
+  test $(< update/image/LAST_VER) -eq 30
+
+  #check if +20 previous version is set to +0
+  awk '$1 == "previous:" { exit $2 != 10}' update/www/30/Manifest.MoM
+
+  #check if +10 previous version is set to +0
+  awk '$1 == "previous:" { exit $2 != 10}' update/www/30/Manifest.MoM
+
+  #check if builder.conf has the +20 format
+  test $(sed -n 's/[ ]*FORMAT[ ="]*\([0-9]\+\)[ "]*/\1/p' builder.conf) -eq 2
+}
 
 # vi: ft=sh ts=8 sw=2 sts=2 et tw=80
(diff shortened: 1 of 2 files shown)


bat/tests/06-test-format-bump/run.bats
```diff
@@ -1,14 +1,48 @@
...
+  echo $upstreamver > upstreamversion
+
+  #Build +10
+  mixer-build-format-bump-old > $LOGDIR/build_formatbump_old.log
+
+  #restore upstrean
+  mv upstreamversion.bump upstreamversion
+
+  #check if LAST_VER is set to +20
+  test $(cat update/image/LAST_VER) -eq 30
```
icarus-sparry: In bash, you can make this slightly more efficient by saying

```test $(< update/image/LAST_VER) -eq 30```

not enough to make it worth changing if this was the only thing to do.

File: bat/tests/06-test-format-bump/run.bats [hunk 2]
```diff
@@ -1,14 +1,48 @@
...
+
+  #check if +20 previous version is set to +0
+  test $(awk -v FS="previous:" 'NF>1{gsub(/\t/, "", $2); print $2}' update/www/30/Manifest.MoM) -eq 10
+
+  #check if +10 previous version is set to +0
+  test $(awk -v FS="previous:" 'NF>1{gsub(/\t/, "", $2); print $2}' update/www/20/Manifest.MoM) -eq 10
```
icarus-sparry: While this works, I would code this as

```awk '$1 == "previous:" { exit $2 == 10}'```

which I find much more obvious. Setting the FS to previous: introduces a corner case if the string previous: occurs anywhere on the line, perhaps as part of a filename.

File: bat/tests/06-test-format-bump/run.bats [hunk 3]
```diff
@@ -1,14 +1,48 @@
...
+
+  #check if builder.conf has the +20 format
+  test $(awk -v FS="FORMAT" 'NF>1{gsub(/=/, "",$2); gsub(/"/,"",$2); gsub(/ /,"",$2); print $2}' builder.conf) -eq 2
```
icarus-sparry: ```test $(sed -n 's/^FORMAT=//p' builder.conf) -eq 2``` would be my suggestion.

bat/tests/06-test-format-bump/run.bats
rchiossi: Updated, thanks.

bat/tests/06-test-format-bump/run.bats
rchiossi: Indeed it looks better. I changed it to 'exit $2 != 10' since I want it to return 0 if they are the same.

bat/tests/06-test-format-bump/run.bats
rchiossi: This doens't work here. It has to work for both config formats:
`FORMAT=2`
`  FORMAT = "2"`
Thats is why I use just FORMAT as FS and use the rest to remove =, " and <space>

icarus-sparry: OK, so ```test $(sed -n s/^FORMAT[ ="]*\([^"]*\)"?/\1/p) -eq 2```

tmarcu: You can also remove all whitespace before parsing the line and use the original suggestion.

bat/tests/06-test-format-bump/run.bats
```diff
@@ -1,14 +1,48 @@
...
-#  global_setup
-#}
+setup() {
+  global_setup
+}
 
-# TODO: fill out once auto format bump implemented
-#@test "Create initial mix 10" {
-#}
+# TODO: This test should run 'mixer build upstream-format' instead
```
tmarcu: 👍

rchiossi:
Updated. I had to redo the last check, but it still uses sed for matching instead of awk.

bat/tests/06-test-format-bump/run.bats
```diff
@@ -1,14 +1,48 @@
...
+  mixer-build-all > $LOGDIR/build_all.log
+
+  #Update to new upstream in the new format
+  mixer-upstream-update $(($upstreamver + 10)) > $LOGDIR/upstream_update.log
+
+  #Build +20
+  mixer-build-format-bump-new > $LOGDIR/build_formatbump_new.log
+
+  #Reset upstream for +10 build
+  echo $upstreamver > upstreamversion
```
tmarcu: Mixer doesn't care anymore but I'd 
`echo $upstreamver > upstreamversion`

PR #357, opened 2018-06-21
Remove source RPMs before creating the repo

tmarcu:
Mixer will not work with source RPMs added and crash looking for files
that will not actually be provided. This patch removes them as it parses
the RPM list so only valid, usable RPMs get added to the repo.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -1838,6 +1838,13 @@ func (b *Builder) AddRPMList(rpms []string) error {
 		if _, err := os.Stat(repoPath); err == nil {
 			continue
 		}
+		// Remove source RPMs because they should not be added to mixes
+		if strings.HasSuffix(rpm, ".src.rpm") {
+			fmt.Printf("Removing %s because source RPMs are not supported in mixes.\n", rpm)
+			if err := os.RemoveAll(filepath.Join(b.Config.Mixer.LocalRPMDir, rpm)); err != nil {
+				return errors.Wrapf(err, "Failed to remove %s, your mix will not generate properly with source RPMs included.", rpm)
+			}
+		}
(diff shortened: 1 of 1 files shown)


builder/builder.go
```diff
@@ -1838,6 +1838,12 @@ func (b *Builder) AddRPMList(rpms []string) error {
 		if _, err := os.Stat(repoPath); err == nil {
 			continue
 		}
+		// Remove source RPMs because they should not be added to mixes
+		if strings.Contains(rpm, ".src.rpm") {
```
matthewrsj: Optional: This could be `strings.HasSuffix` to be a little faster.

builder/builder.go
```diff
@@ -1838,6 +1838,12 @@ func (b *Builder) AddRPMList(rpms []string) error {
...
+			if err := os.RemoveAll(filepath.Join(b.Config.Mixer.LocalRPMDir, rpm)); err != nil {
```
matthewrsj: Optional: This should probably fail if that's a directory instead of a file, so `os.Remove` would probably be better here.

builder/builder.go
```diff
@@ -1838,6 +1838,12 @@ func (b *Builder) AddRPMList(rpms []string) error {
...
+				return errors.Wrapf(err, "Failed to remove %s, your mix will not generate properly with source RPMs included.", rpm)
+			}
```
matthewrsj: No warning when we are removing files they put there?

builder/builder.go
tmarcu: My selective reading thought this wasn't a suffix for some reason, thanks :) changing.

builder/builder.go
tmarcu: I will add one 👍

PR #360, opened 2018-06-27
Do not hit the network for some commands

tmarcu:

### DIFF
--- a/mixer/cmd/root.go
+++ b/mixer/cmd/root.go
@@ -75,8 +75,17 @@ var RootCmd = &cobra.Command{
 			fail(err)
 		}
 
+		networkCheck := true
+		noNetworkCmds := []string{"list", "edit", "validate", "convert", "set", "repo", "add-rpms"}
+		// Don't reach out over network for these commands, it's not needed
+		for _, ignoreCmd := range noNetworkCmds {
+			if cmdContains(cmd, ignoreCmd) {
+				networkCheck = false
+			}
+		}
+
 		// If running natively, check for format missmatch and warn
-		if builder.Native && b.UpstreamURL != "" {
+		if networkCheck && builder.Native && b.UpstreamURL != "" {
 			hostFormat, upstreamFormat, err := b.GetHostAndUpstreamFormats()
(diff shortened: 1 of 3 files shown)


mixer/cmd/root.go
```diff
@@ -75,8 +75,17 @@ var RootCmd = &cobra.Command{
 			fail(err)
 		}
 
+		networkCheck := true
+		noNetworkCmds := []string{"list", "edit", "validate", "convert", "set", "repo", "add-rpms"}
```
rchiossi: Hmm, I don't really like this approach since it ignores commands subcategories. Something like "set" is so generic that it can easily end up in multiple categories.

Other than printing the warnings, this block doesn't do anything else. Wouldn't it be better to just move it to the commands where the warning is relevant?

mixer/cmd/root.go
```diff
@@ -75,8 +75,17 @@ var RootCmd = &cobra.Command{
...
+		for _, ignoreCmd := range noNetworkCmds {
+			if cmdContains(cmd, ignoreCmd) {
+				networkCheck = false
+			}
+		}
+
 		// If running natively, check for format missmatch and warn
-		if builder.Native && b.UpstreamURL != "" {
+		if networkCheck && builder.Native && b.UpstreamURL != "" {
 			hostFormat, upstreamFormat, err := b.GetHostAndUpstreamFormats()
```
rchiossi: Looking at the code below, it also seems weird to me that this should fail() on error instead of just printing the first warning. Afteral, when you get an error here you were "Unable to determine host format".

mixer/cmd/root.go
tmarcu: I'm open to updating this, this was just an initial list I came up before asking anyone for RFC's!

That being said "set" likely is never going to set something over the network, so it's not a bad catch-all.

mixer/cmd/root.go
tmarcu: Hmm, you make a good point. I might have to update this but I want to think about it more :)

PR #368, opened 2018-07-17
Notify user about bundle changes after format bump

tmarcu:
During a format bump, the upstream version for a mix is updated, which may
introduce changes to the upstream bundles. Any changes to included (and
un-edited) bundles will be propagated to the mix, so warn the user to take
changes into consideration.
Local bundle definitions will override upstream ones, so even if upstream
bundles change, they will not change the locally defined versions for a mix.

### DIFF
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -163,6 +163,8 @@ var buildUpstreamFormatCmd = &cobra.Command{
 				fail(err)
 			}
 		}
+		fmt.Println(`NOTE: upstream bundle definitions may have changed,
+please review any upstream bundles included in your mix.`)
 	},
 }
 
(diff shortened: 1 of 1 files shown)


mixer/cmd/build.go
```diff
@@ -163,6 +163,8 @@ var buildUpstreamFormatCmd = &cobra.Command{
 				fail(err)
 			}
 		}
+		fmt.Println(`NOTE: upstream bundle definitions may have changed,
```
matthewrsj: This prints every time, regardless of whether the bundle definitions have change?

tmarcu: yes :O 

the only way to check I think would be to save a copy of the old ones and then diff the folders. Which I can do if that's a better time to print.

tmarcu:
Closing and will update documentation with proper steps to curate bundles as needed.

PR #370, opened 2018-07-19
Use New Config as Default

rchiossi:
This patchset adds a couple changes to the config system to allow the TOML config to be used as default. Among this changes are a versioning system and, auto detection of the config format.

This patch also introduces the mixer.state file. Currently it only holds the format variable, which was the last remaining transient variable in builder.conf, but the idea is to concentrate all the other temporary files (e.g. upstreamversion) under this state file.

Note: I'm adding subsections to mixer.state file and I'll update this commit before it is merged. This will also make the batcheck pass (atm It will fail when trying to validate the format).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,6 +15,7 @@
 package config
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"io/ioutil"
@@ -33,6 +34,9 @@ import (
 // This is an experimental feature.
 var UseNewConfig = false
 
+// CurrentConfigVersion holds the current version of the config file
+const CurrentConfigVersion = "1.0"
+
 // MixConfig represents the config parameters found in the builder config file.
 type MixConfig struct {
 	Builder builderConf
@@ -42,6 +46,7 @@ type MixConfig struct {
 
 	/* hidden properties */
 	filename string
+	version  string
 
 	// Format moved into mixer.state file. This variable is set
 	// if the value is still present in the parsed config to
@@ -116,6 +121,7 @@ func (config *MixConfig) LoadDefaultsForPath(localrpms bool, path string) {
 		config.Mixer.LocalRepoDir = ""
 	}
 
+	config.version = CurrentConfigVersion
 	config.filename = filepath.Join(path, "builder.conf")
 
 	config.hasFormatField = false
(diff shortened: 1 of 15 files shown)


config/config.go
```diff
@@ -31,7 +32,10 @@ import (
 
 // UseNewConfig controls whether to use the new TOML config format.
 // This is an experimental feature.
-var UseNewConfig = false
+var UseNewConfig = true
+
+// CurrentConfigVersion holds the current version of the config file
+const CurrentConfigVersion = "1.0"
```
tmarcu: What does this version actually mean for future btw? Are incremental changes - i.e adding new fields = move to 1.1, and breaking changes would make it 2.0?

rchiossi: I can't really see a scenario where it would Break, unless we change to a new format again.
I was thinking 1.1 for field addition/removal and 2.0 for new sections or section reorg.

rchiossi:
Updated. Now mixer.state has a subsection [Mix] for the Format variable. The only notable change in the code is in the format.go:UpdateFormatVersion() which now always uses MixState for updating the format.

config/config.go
```diff
@@ -113,7 +118,10 @@ func (config *MixConfig) LoadDefaultsForPath(localrpms bool, path string) {
 		config.Mixer.LocalRepoDir = ""
 	}
 
+	config.version = CurrentConfigVersion
 	config.filename = filepath.Join(path, "builder.conf")
+
+	config.formatSet = false
```
tmarcu: maybe formatIsSet? /me is unsure what to suggest here ;(

config/state.go
```diff
@@ -0,0 +1,117 @@
...
+type MixState struct {
+	Mix mixSection
+
+	/* hidden properties */
+	filename string
+	version  string
+}
+
+// CurrentStateVersion is the current revision for the state file structure
+var CurrentStateVersion = "1.0"
```
tmarcu: Does this override or get overwritten by the "1.0" in config.go ? `CurrentConfigVersion` specifically*

config/config.go
tmarcu: Ok, I agree with that but move removal to 2.0 cause that would technically be a "reorg" or breaking change. Thanks for clarifying!

config/state.go
rchiossi: They serve the same purpose but for different files. Config has a version and State has a version. They are not tied, they just happen to be both at version 1.0 right now. I'm not sure how useful it will be for the state file, but since I now see that It would be much better if builder.conf had it from the start, I'm also adding it for the state file.

config/config.go
```diff
@@ -42,6 +46,9 @@ type MixConfig struct {
 
 	/* hidden properties */
 	filename string
+	version  string
+
```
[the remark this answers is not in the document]
matthewrsj: e x t r a         

s p a c e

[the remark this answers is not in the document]
tmarcu: It showed 2 newlines when I reviewed it :O sorry for the confusion.

rchiossi:
No Problem. :)

PR #371, opened 2018-07-23
Remove LD_PRELOAD

tmarcu:
The new dnf release *in Clear* now does this by default, so we do not have to load
the nosync library anymore.

### DIFF
--- a/mixer/main.go
+++ b/mixer/main.go
@@ -15,16 +15,9 @@
 package main
 
 import (
-	"fmt"
-	"os"
-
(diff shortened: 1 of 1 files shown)


phmccarty:
This change to dnf is not even in a Clear Linux build yet. Can we wait to see how well it works first?

icarus-sparry:
I agree with @phmccarty quite apart from the fact that I think removing fsync is a fundamentally bad idea.

tmarcu:
@icarus-sparry pls elaborate

PR #372, opened 2018-07-24
state: Better default values for FORMAT

rchiossi:
Instead of always using '1' as the default FORMAT value, check first if
the value was available in a legacy config. If not, try using the system
format instead. If both fail, fallback to the hardcoded value as it was
previously done.

### DIFF
--- a/config/state.go
+++ b/config/state.go
@@ -18,6 +18,8 @@ import (
 	"bufio"
 	"bytes"
 	"errors"
+	"fmt"
+	"io/ioutil"
 	"os"
 	"regexp"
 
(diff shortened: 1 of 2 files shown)


config/state.go
```diff
@@ -35,19 +37,58 @@ type MixState struct {
...
+	/* Get format from legacy config file */
+	format, err := state.getFormatFromConfig()
+	if err == nil && format != "" {
+		state.Mix.Format = format
+		state.formatSource = "builder.conf"
+		return
+	}
+
+	/* Get format from system */
+	formatBytes, err := ioutil.ReadFile("/usr/share/defaults/swupd/format")
```
matthewrsj: Since this is configurable in swupd (`DEFAULT_FORMAT_PATH`) should it be configurable here? I know, yuck, but if it has potential to change...

rchiossi: Ideally we should have something that keeps both projects in sync regarding this path, but I'm not sure having the same build option is the best way to go.

matthewrsj: OK then at least a const?

rchiossi: Sounds good to me! I'll update the PR.

config/state.go
```diff
@@ -79,6 +123,7 @@ func (state *MixState) Load() error {
 	f, err := os.Open(state.filename)
 	if err != nil {
 		// If state does not exists, create a default state
+		fmt.Println("WARNING: Using FORMAT value from " + state.formatSource)
```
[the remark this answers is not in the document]
tmarcu: I'd say yeah, but technically it's not an error just informative message since it may have found a good source for the format. /me leaves this one to @rchiossi

Issue #374, opened 2018-07-26
swupd: deletes must persist over full minversions


matthewrsj:
When creating manifests for a full minversion, with the minversion set to the current version being built, deletes must persist in the new manifests. The only time old deletes should be cleaned up is when building in a new format.

The following test currently fails.

```diff
diff --git a/swupd/create_manifests_test.go b/swupd/create_manifests_test.go
index b9b26b9..aca4771 100644
--- a/swupd/create_manifests_test.go
+++ b/swupd/create_manifests_test.go
@@ -366,6 +366,28 @@ func TestCreateManifestsMinVersion(t *testing.T) {
        ts.checkNotContains("www/20/Manifest.full", "\t10\t")
 }
 
+func TestCreateManifestsMVDeletes(t *testing.T) {
+       ts := newTestSwupd(t, "minVersionDeletes")
+       defer ts.cleanup()
+
+       ts.Bundles = []string{"test-bundle"}
+       ts.addFile(10, "test-bundle", "/foo", "foo")
+       ts.createManifests(10)
+       ts.checkContains("www/10/Manifest.test-bundle", "10\t/foo\n")
+       ts.checkContains("www/10/Manifest.full", "10\t/foo\n")
+
+       // update and remove /foo
+       ts.createManifests(20)
+       fileDeletedInManifest(t, ts.parseManifest(20, "test-bundle"), 20, "/foo")
+       fileDeletedInManifest(t, ts.parseManifest(20, "full"), 20, "/foo")
+
+       // now create a minversion and make sure the deletes persist
+       ts.MinVersion = 30
+       ts.createManifests(30)
+       fileDeletedInManifest(t, ts.parseManifest(30, "test-bundle"), 30, "/foo")
+       fileDeletedInManifest(t, ts.parseManifest(30, "full"), 30, "/foo")
+}
+
 func TestCreateManifestsPersistDeletes(t *testing.T) {
        ts := newTestSwupd(t, "persistDeletes")
        defer ts.cleanup()
```

matthewrsj:
Fixed by #375

PR #375, opened 2018-07-26
swupd: persist deletes over minversions, but not format bumps

matthewrsj:

See #374


matthewrsj:
Fixes #374

PR #381, opened 2018-07-27
cmd: Add --format flag to mixer init

rchiossi:
The --format flag allows a custom format to be provided to a mix on
initialization


gtkramer:
lgtm, assuming the data structure containing the format is automatically dumped into the toml state file correctly.

PR #384, opened 2018-07-30
config: Add missing default value for docker

rchiossi:
The default value for DOCKER_IMAGE_PATH was only set for the old INI
config. As a result, TOML configs would have an empty path that would
generate an invalid docker command to be issued.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -112,6 +112,7 @@ func (config *MixConfig) LoadDefaultsForPath(localrpms bool, path string) {
 
 	// [Mixer]
 	config.Mixer.LocalBundleDir = filepath.Join(path, "local-bundles")
+	config.Mixer.DockerImgPath = "clearlinux/mixer"
 
 	if localrpms {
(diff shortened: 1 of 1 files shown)


kevincwells:
So initially, this field was not added to the default config, and only a default value was added. This was intentional, because we don't really want people changing that unless they really know what they're doing.

My understanding is that recent config changes make it so that _every_ field is now included in the default config, and hence we ended up with `DOCKER_IMAGE_PATH = ""`.

Is it really not possible to control which fields get written to the default? As is, the default config now also contains `LOCAL_RPM_DIR = ""` and `LOCAL_REPO_DIR = ""`, where previous versions would omit them if they weren't being set (i.e., `--local-rpms`).

kevincwells:
I guess my point is, +1 to this fix for now, 'cause this needs to get fixed ASAP, but longer term I think we'd want the ability to omit fields from the autogenerated config. And at that point, I'd want to take this field back out.

Issue #376, opened 2018-07-26
Add a --no-clean option for build bundles


gtkramer:
Such an option would be useful to allow users to mount a block device with ister in the full image/chroot directory mixer makes during the build bundles operation.  This can be used to allow users to prototype bootable images very quickly if many need prototyped at once because it avoids needing to copy the full image/chroot into a block device and then remove it after the block device is unmounted.

matthewrsj:
But the full chroot sticks around. What are you asking not to clean?

tmarcu:
@matthewrsj when build bundles is called, it wipes update/image/VER no matter what. His use case is specific to some custom testing to already have image/VER/full created, and mount the image to full, so mixer automatically populates the VM filesystem as part of the build. Saves some time copying into the image after building & don't have to wipe /full after to save space from having it in 2 places (workspace + vm).

gtkramer:
Saves time and disk I/O when I already have a guarantee of a clean environment.  It's more of an advanced usage option only for folks who know what they are doing.

PR #390, opened 2018-08-04
Don't wipe output dirs on build bundles

tmarcu:
By default, build bundles removes image/ver and www/ver, so that builds
of the same number start with a clean slate. There are specific use cases
where this is not desirable, and the statedir already has a version folder
populated/mounted which shouldn't be wiped. This allows mixer to write
the full chroot into an actual mounted file system for example that is
attached to update/image/VER/full.

Fixes https://github.com/clearlinux/mixer-tools/issues/376

### DIFF
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -35,6 +35,7 @@ type buildCmdFlags struct {
 	format        string
 	increment     bool
 	minVersion    int
+	clean         bool
 	noSigning     bool
 	noPublish     bool
 	template      string
@@ -73,7 +74,7 @@ var buildCmd = &cobra.Command{
 	Short: "Build various pieces of OS content",
 }
 
-func buildBundles(builder *builder.Builder, signflag bool) error {
+func buildBundles(builder *builder.Builder, signflag, cleanFlag bool) error {
 	// Create the signing and validation key/cert
 	if _, err := os.Stat(builder.Config.Builder.Cert); os.IsNotExist(err) {
 		fmt.Println("Generating certificate for signature validation...")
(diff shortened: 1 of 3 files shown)


tmarcu:
@gtkramer

mixer/cmd/build.go
```diff
@@ -545,6 +546,7 @@ func init() {
 
 	RootCmd.AddCommand(buildCmd)
 
+	buildBundlesCmd.Flags().BoolVar(&buildFlags.noClean, "no-clean", false, "Do not wipe the image/VER directory if it exists")
```
icarus-sparry: Is the double negative (not no-clean) worthwhile? Would it be better to have the flag be called "clean" and default to true? Maybe also be worth saying that the cleanup happens at the start?

tmarcu: mmm Yeah I can change the flag name to be more like "--don't-call-this--normally". 
I like `--ignore-dirty` or `--use-existing`

gtkramer: Probably just want to use --ignore-dirty so users know they are taking on a risk by keeping dirt around :)

matthewrsj: I'd have to agree with @icarus-sparry here. I'd prefer it to use the old state if it exists, maybe printing a notification that it is attempting to use old state for the build. If this isn't what the user wanted they can re-run and pass --clean. I don't think mixer should be wiping previous work by default, especially if that previous work is needed in some use-cases.

The warning from mixer that the clean is happening is essentially useless, since it warns right before it actually wipes the old state. The user has no way to abort that process. Consider that someone puts work into that directory for mixer to use, mixer runs and wipes it without them knowing they had to pass --ignore-dirty to keep their work.

gtkramer:
I tested this with some automation and it works great!  Mixer can now write directly to a raw image that's mounted as a loopback device.

+1

tmarcu:
Updated to use --clean instead and by default not wipe anything :)

mixin/build.go
```diff
@@ -67,7 +67,7 @@ func buildBundles(b *builder.Builder) error {
 		template = helpers.CreateCertTemplate()
 	}
 
-	return errors.Wrap(b.BuildBundles(template, privkey, false), "Error building bundles")
+	return errors.Wrap(b.BuildBundles(template, privkey, false, false), "Error building bundles")
```
matthewrsj: do we want to clean here by default, since this will be hidden from the user?

tmarcu: Really good question, and I think the answer is no - because iirc mixin will always increment the mix version number. If it doesn't, then this needs to be changed to true, or have mixin always bump the mixver number at the beginning to always start on a new ver.

matthewrsj: yep, ok.

tmarcu: Ok good point, mixin only incrememnts mixver if the previous build was valid, so if it crashes it will not be valid and using bad data possibly. Changing to true.

PR #392, opened 2018-08-07
Use cobra for exact args limits

tmarcu:

### DIFF
--- a/mixer/cmd/repos.go
+++ b/mixer/cmd/repos.go
@@ -15,7 +15,6 @@
 package cmd
 
 import (
-	"errors"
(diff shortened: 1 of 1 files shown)


rchiossi:
I think there where similar cases in the other commands as well. Did you check the rest of the code?

tmarcu:
@rchiossi no :upside_down_face:

tmarcu:
Ok after checking the code, I haven't figured out a way to use multiple constraints in the Args: section, so the remaining ones can't be changed like in this PR because they already defined different constraints, i.e cobra.ValidArgsOnly.

PR #393, opened 2018-08-08
swupd: export IndexBundle for external use

matthewrsj:
Other packages want to use the IndexBundle const to skip processing this
manifest for some validation cases.


matthewrsj:
Info not in commit message: need this exported for diva manifestcheck

PR #394, opened 2018-08-09
Add excludepkgs option to repo command

tmarcu:

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -1840,6 +1840,33 @@ func (b *Builder) SetURLRepo(name, url string) error {
 	return DNFConf.SaveTo(b.Config.Builder.DNFConf)
 }
 
+// SetExcludesRepo sets the ecludes for the repo <name> to [pkgs...]
+func (b *Builder) SetExcludesRepo(reponame, pkgs string) error {
+	if err := b.NewDNFConfIfNeeded(); err != nil {
+		return err
+	}
+
+	DNFConf, err := ini.Load(b.Config.Builder.DNFConf)
+	if err != nil {
+		return err
+	}
+
+	s, err := DNFConf.GetSection(reponame)
+	if err != nil {
+		return err
+	}
+
+	k, err := s.GetKey("excludepkgs")
+	if err != nil {
+		if k, err = DNFConf.Section(reponame).NewKey("excludepkgs", pkgs); err != nil {
+			return err
+		}
+	}
+
+	k.SetValue(pkgs)
+	return DNFConf.SaveTo(b.Config.Builder.DNFConf)
+}
+
 // RemoveRepo removes a configured repo <name> if it exists in the DNF configuration.
 // This will fail if a DNF conf has not yet been generated.
 func (b *Builder) RemoveRepo(name string) error {
--- a/mixer/cmd/repos.go
+++ b/mixer/cmd/repos.go
@@ -16,6 +16,7 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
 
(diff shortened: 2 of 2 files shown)


gtkramer:
@tmarcu  and I pair-programmed this.

+1

mixer/cmd/repos.go
```diff
@@ -66,12 +67,21 @@ var setURLRepoCmd = &cobra.Command{
 	Run:   runSetURLRepo,
 }
 
+var setExcludesRepoCmd = &cobra.Command{
+	Use:   "exclude <repo> <pkg> [<pkg>...]",
+	Short: "Exclude packages from a specified repo",
+	Long:  `Exclude packages from a specified repo. These packages will be ignored during build bundles. Globbing is supported.`,
```
matthewrsj: If you really are adding this functionality to mixer as opposed to letting users modify the yum conf themselves you should probably put the use case in the `Long` text here.

builder/builder.go
```diff
@@ -1840,6 +1840,33 @@ func (b *Builder) SetURLRepo(name, url string) error {
 	return DNFConf.SaveTo(b.Config.Builder.DNFConf)
 }
 
+// SetExcludesRepo sets the ecludes for the repo <name> to [pkgs...]
+func (b *Builder) SetExcludesRepo(name, pkgs string) error {
```
matthewrsj: Maybe rename `name` to `repoName` to be more explicit about what it is naming. I had to look at the caller to figure out what that was.

gtkramer:
I think this is a feature that would help users guarantee correctness of bundles built by mixer in lieu of some shortcomings with dnf where dnf produces an unexpected dependency solution.  This mainly happens when a user is intending to fully substitute a package from Clear's repo with one from their own.  If anything is wrong with the substitution, a feature like this would expose the issue early on.

Although, I see how this starts to turn mixer into a little bit of a fancy interface for dnf confs.  For non-human users working in a bash environment, editing config files tends to be a pain and somewhat brittle.  Having mixer offer commands to interface with dnf confs like `mixer repo add` are definitely nice and welcomed features for automation.

tmarcu:
Mixer has historically changed the configs it uses and their formats, I think the most user friendly solution is to have them edited by mixer itself, so when they inevitably change again, it's completely supported and transparent to the user since they will still be calling the same "add/edit" commands.

matthewrsj:
> Mixer has historically changed the configs it uses and their formats

It changes the in-house configs. This is a DNF config and we as mixer developers have no control over DNF config format.

> they will still be calling the same "add/edit" commands

Well I see your point, but those were only added because we needed top-level support for remote repos.

tmarcu:
That's true, but it's still less work IMO to just use mixer to say exclude <list> than edit (hidden) confs in scripts and elsewhere.

matthewrsj:
**maybe** the conf shouldn't be hidden, but ok...

mixer/cmd/repos.go
matthewrsj: Please still add the use-case here. Something like "Allows the user to explicitly select which packages to use when building bundles by excluding the unwanted ones"

tmarcu:
It's only hidden when generated by default if the user doesn't specify their own conf, it can be named and placed however the user wants.

matthewrsj:
:disappointed:

mixer/cmd/repos.go
matthewrsj: you didn't add the use case

matthewrsj: You didn't add the use case

Issue #378, opened 2018-07-27
Mixer still checks online when --offline is passed to build bundles


tmarcu:
Don't check online when --offline is passed...

kevincwells:
Unless it's changed since I introduced it, `--offline` only prevents mixer from pulling upstream bundle definitions. It's a short-circuit early return in the fetch function. What other network calls is mixer making that you're trying to prevent?

tmarcu:
#398 Merged and released,

PR #398, opened 2018-08-10
Fix global --offline usage

tmarcu:
When --offline is passed, any command executed should not reach over the
network. This introduces a tricky scenario when it is passed to a build
command with --native=false (default), because the command in the
container will be run with --offline, but the native mixer binary will
reach over the network to pull/update docker containers before executing
in them. This patch also makes that process offline only, attempting to
use a cached docker image so that it can run offline fully, but exits if
none exist.

### DIFF
--- a/builder/docker.go
+++ b/builder/docker.go
@@ -160,24 +160,58 @@ func (b *Builder) getDockerMounts() ([]string, error) {
 	return reduceDockerMounts(mounts), nil
 }
 
+func (b *Builder) prepareContainer() (string, error) {
+	format, err := b.getUpstreamFormat(b.UpstreamVer)
+	if err != nil {
+		return "", err
+	}
+	imageName, err := b.getDockerImageName(format)
+	if err != nil {
+		return "", errors.Wrap(err, "Unable to get docker image name for format "+format)
+	}
+	fmt.Println("Updating docker image")
+	if err = helpers.RunCommand("docker", "pull", imageName); err != nil {
+		fmt.Printf("WARNING: Unable to pull docker image for format %s. Trying with cached image\n", format)
+	}
+
+	return imageName, nil
+
+}
+
+func (b *Builder) prepareContainerOffline() (string, error) {
+	imageName, err := b.getDockerImageName(b.State.Mix.Format)
+	if err != nil {
+		var hostFormat []byte
+		if hostFormat, err = ioutil.ReadFile("/usr/share/defaults/swupd/format"); err != nil {
+			return "", err
+		}
+		imageName, err = b.getDockerImageName(string(hostFormat))
+		if err != nil {
+			return "", errors.Wrapf(err, "Unable to get docker image name for format %s", string(hostFormat))
+		}
+	}
+
+	// We know the right image name and format now so only need to run this once
+	if err := helpers.RunCommandSilent("docker", "image", "inspect", imageName); err != nil {
+		return "", errors.Errorf("Failed to find usable docker image, cannot run offline: %s", err)
+	}
+
+	return imageName, nil
+}
+
 // RunCommandInContainer will pull the content necessary to build a docker
 // image capable of running the desired command, build that image, and then
 // run the command in that image.
 func (b *Builder) RunCommandInContainer(cmd []string) error {
-	format, err := b.getUpstreamFormat(b.UpstreamVer)
+	var imageName string
+	var err error
+	if Offline {
+		imageName, err = b.prepareContainerOffline()
+	} else {
+		imageName, err = b.prepareContainer()
+	}
 	if err != nil {
 		return err
-	}
-
-	imageName, err := b.getDockerImageName(format)
-	if err != nil {
-		return errors.Wrap(err, "Unable to get docker image name for format "+format)
-	}
-
-	fmt.Println("Updating docker image")
-
-	if err = helpers.RunCommand("docker", "pull", imageName); err != nil {
-		fmt.Printf("WARNING: Unable to pull docker image for format %s. Trying with cached image\n", format)
 	}
 
 	fmt.Printf("Running command in container: %q\n", strings.Join(cmd, " "))
--- a/mixer/cmd/root.go
+++ b/mixer/cmd/root.go
@@ -110,6 +110,14 @@ var RootCmd = &cobra.Command{
 		// If so: inform, stage, and exit.
 		// If not: run command in container and cancel pre-run
 		if !cmdContains(cmd, "format-bump") && !cmdContains(cmd, "upstream-format") && cmdContains(cmd, "build") {
+			// --offline=true AND --native=false, try to see if container exists
+			if builder.Offline && !builder.Native {
+				fmt.Println("Warning: Unable to determine upstream format in --offline mode, build may fail if building across format boundaries.")
+				if err := b.RunCommandInContainer(reconstructCommand(cmd, args)); err != nil {
+					fail(err)
+				}
+				return nil
+			}
 			if bumpNeeded, err := b.CheckBumpNeeded(false); err != nil {
 				return err
 			} else if bumpNeeded {


builder/docker.go
```diff
@@ -163,21 +164,35 @@ func (b *Builder) getDockerMounts() ([]string, error) {
...
+	var err error
+	if !offline {
+		format, err = b.getUpstreamFormat(b.UpstreamVer)
+		if err != nil {
+			return err
+		}
+	} else {
+		var hostFormat []byte
+		hostFormat, err = ioutil.ReadFile("/usr/share/defaults/swupd/format")
+		if err != nil && !os.IsNotExist(err) {
```
phmccarty: If I'm reading this correctly, this block is entered when any error is raised other than the file not existing. If the file doesn't exist, then no content is read, so using the `hostFormat` value makes little sense. So, why specify `!os.IsNotExist(err)` here?

builder/docker.go
```diff
@@ -163,21 +164,35 @@ func (b *Builder) getDockerMounts() ([]string, error) {
 // RunCommandInContainer will pull the content necessary to build a docker
 // image capable of running the desired command, build that image, and then
 // run the command in that image.
-func (b *Builder) RunCommandInContainer(cmd []string) error {
-	format, err := b.getUpstreamFormat(b.UpstreamVer)
-	if err != nil {
-		return err
+func (b *Builder) RunCommandInContainer(cmd []string, offline bool) error {
```
kevincwells: Why introduce a new `offline` variable? `docker.go` is within the `builder` package, so it can already access the package-wide `Offline` variable that gets set when `--offline` is passed. (See [`getUpstreamBundles`](https://github.com/clearlinux/mixer-tools/blob/master/builder/builder.go#L414) for how it's used elsewhere.) Do you actually have cases where you want to pass a value of `offline` here that is different from what `builder.Offline` will hold?

builder/docker.go
kevincwells: Elsewhere in `docker.go` we have [`GetHostAndUpstreamFormats`](https://github.com/clearlinux/mixer-tools/blob/master/builder/docker.go#L34).

a) It'd be nice if these two methods used the same code snippet to parse the format file.
b) In that method, `!os.IsNotExist(err)` makes sense, because the file not existing isn't an error condition, but a warning -- the code that calls it checks for the returned empty string and warns that the format could not be determined. Here, yes, it seems like you want to error out if it cannot be determined. In that case, though, drop the `&& os.IsNotExist(err)` altogether. Otherwise you're _only_ catching File Not Exist.

mixer/cmd/root.go
```diff
@@ -105,6 +108,18 @@ var RootCmd = &cobra.Command{
 		// If so: inform, stage, and exit.
 		// If not: run command in container and cancel pre-run
 		if !cmdContains(cmd, "format-bump") && !cmdContains(cmd, "upstream-format") && cmdContains(cmd, "build") {
+			// --offline=true AND --native=false, try to see if container exists
+			if !networkCheck && !builder.Native {
```
kevincwells: Why not just look at `builder.Offline` here, rather than `networkCheck`? You say this is based on `--offline=true`, but you're routing that through `networkCheck`, which could be set for other reasons. Feels like a bad coupling to me.

mixer/cmd/root.go
```diff
@@ -105,6 +108,18 @@ var RootCmd = &cobra.Command{
...
+				fmt.Println("Warning: Unable to determine upstream format in --offline mode, build may fail if format boundary is crossed.")
```
kevincwells: Nitpick: Would prefer phrasing like ".. may fail if building across format boundaries". As it it sounds like the build would fail only if it is a build intended to cross a boundary.

mixer/cmd/root.go
```diff
@@ -105,6 +108,18 @@ var RootCmd = &cobra.Command{
...
+				format := b.State.Mix.Format
+				_, err := b.GetDockerImageName(format)
```
kevincwells: Why are you checking this here? It gets checked in `RunCommandInContainer`, which you call directly below.

mixer/cmd/root.go
```diff
@@ -33,6 +33,13 @@ import (
 
 var configFile string
 
+// Adding function here instead of including all of helpers package
+func runCommand(cmdname string, args ...string) error {
```
kevincwells: The benefit of the wrapper is that you know all the places that use it are calling the underlying command in a uniform way. But if this is now a possibly-out-of-sync copy of that wrapper, its utility is lost.

Just use `exec.Command(...).Run()` directly below rather than copying this pretty thin wrapper.

builder/docker.go
```diff
@@ -47,7 +47,8 @@ func (b *Builder) GetHostAndUpstreamFormats() (string, string, error) {
 	return string(hostFormat), upstreamFormat, nil
 }
 
-func (b *Builder) getDockerImageName(format string) (string, error) {
+// GetDockerImageName gets the existing image name
+func (b *Builder) GetDockerImageName(format string) (string, error) {
```
kevincwells: If you're not calling this in `root.go` anymore, it doesn't need to be exported.  Though I guess it's not _hurting_ anything...

mixer/cmd/root.go
```diff
@@ -105,6 +112,17 @@ var RootCmd = &cobra.Command{
 		// If so: inform, stage, and exit.
 		// If not: run command in container and cancel pre-run
 		if !cmdContains(cmd, "format-bump") && !cmdContains(cmd, "upstream-format") && cmdContains(cmd, "build") {
+			// --offline=true AND --native=false, try to see if container exists
+			if builder.Offline && !builder.Native {
+				if err := runCommand("docker", "image", "inspect", "clearlinux/mixer:"+b.State.Mix.Format); err != nil {
```
kevincwells: You might want to consider putting this check inside `RunCommandInContainer` itself.

As is, it's within a code block that only runs for non-format-bump commands. But what happens if someone tries to run a format bump command with `--offline`? Technically if they already have the right containers cached, they should be able to... Is there a check somewhere that stops them?

Interesting thought: technically speaking, `mixer build format-bump` (the _downstream format bump_ command) could have a `--native` version. As long as your host tooling is the right version for the format you're building, we could call in to the build libs directly in the right order, rather than calling out to `RunCommandInContainer`...

tmarcu:
https://github.com/clearlinux/mixer-tools/issues/378

builder/docker.go
```diff
@@ -47,6 +47,7 @@ func (b *Builder) GetHostAndUpstreamFormats() (string, string, error) {
 	return string(hostFormat), upstreamFormat, nil
 }
 
+// GetDockerImageName gets the existing image name
```
rchiossi: Did you plan on exporting this function and gave up?

builder/docker.go
```diff
@@ -164,20 +165,37 @@ func (b *Builder) getDockerMounts() ([]string, error) {
...
+		format, err = b.getUpstreamFormat(b.UpstreamVer)
+		if err != nil {
+			return err
+		}
+	} else {
+		if err = helpers.RunCommandSilent("docker", "image", "inspect", "clearlinux/mixer:"+b.State.Mix.Format); err != nil {
+			return errors.Errorf("Failed to find usable docker image, cannot run offline: %s", err)
+		}
+		var hostFormat []byte
+		hostFormat, err = ioutil.ReadFile("/usr/share/defaults/swupd/format")
```
rchiossi: Why do you use the host format instead of `b.State.Mix.Format`? When inside the container, the format that that mixer version knows how to build is `b.State.Mix.Format`, not `/usr/share/defaults/swupd/format`

builder/docker.go
```diff
@@ -164,20 +165,37 @@ func (b *Builder) getDockerMounts() ([]string, error) {
...
+		if err != nil {
+			return err
+		}
+		format = string(hostFormat)
+
 	}
 
-	imageName, err := b.getDockerImageName(format)
+	imageName, err = b.getDockerImageName(format)
```
rchiossi: Given the comment above, in this point getDockerImageName() will fail if you changed the format from `b.State.Mix.Format` to `/usr/share/defaults/swupd/format`. If you really need to enforce the use of the host format, it may be better to set it before the call to docker image inspect, so you can fail earlier.

builder/docker.go
tmarcu: It was exported in a V2 :P

builder/docker.go
tmarcu: hmmm you may be right, unless the user changed their b.State.Mix.Format to a value that is out of range with what image they currently have on their system, but that's the same as their host system being out of range too. I'll update this ^

rchiossi:
I meant the else of that block.

builder/docker.go
```diff
@@ -160,26 +160,60 @@ func (b *Builder) getDockerMounts() ([]string, error) {
...
 		fmt.Printf("WARNING: Unable to pull docker image for format %s. Trying with cached image\n", format)
 	}
 
+	return imageName, nil
+
+}
+
+func (b *Builder) prepareContainerOffline() (string, error) {
+	var format string
+	if err := helpers.RunCommandSilent("docker", "image", "inspect", "clearlinux/mixer:"+b.State.Mix.Format); err != nil {
```
rchiossi: Shouldn't you use `getDockerImageName()` here instead of `"clearlinux/mixer:" + b.State.Mix.Format`? That string is configurable, so the user might have changed it. If so, the cached image won't have that name.

tmarcu: That's a good point, but the user should never change the cached image name otherwise they'd not be able to do auto format bumps anymore.

tmarcu:
@rchiossi can you check the latest revision of docker.go, and approve it explicitly if it's goood?

PR #401, opened 2018-08-14
Using ldflags for mixer version string

alexjch:
This change moves the setting of mixer version string from source code
to a variable defined during compilation. This approach is preferable
because:

1- builder.go does not need to change just to change version.
2- simplifies logic in Makefile, no need to retrieve value from source.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -44,8 +44,8 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Version of Mixer. Also used by the Makefile for releases.
-const Version = "5.1.0"
+// Version of Mixer. This is provided by ldflags in Makefile during compilation
+var Version = ""
 
 // Native controls whether mixer runs the command on the native machine or in a
 // container.
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ include Makefile.bats
 
 .NOTPARALLEL:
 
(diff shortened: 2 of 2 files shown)


builder/builder.go
```diff
@@ -44,8 +44,8 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Version of Mixer. Also used by the Makefile for releases.
-const Version = "5.1.0"
+// Version of Mixer. This is provided by ldflags in Makefile during compilation
+var Version = ""
```
tmarcu: This is used in https://github.com/clearlinux/mixer-tools/blob/master/mixer/cmd/root.go#L56

so mixer --version will print "" if this is left empty.

rchiossi: This won't be empty. The value is set on compile time by `-ldflags="-X ${GO_PACKAGE_PREFIX}/builder.Version=${VERSION}"`
Doccumentation regarding -X option is pretty bad, but this seems right.

alexjch: [Line #40](https://github.com/clearlinux/mixer-tools/pull/401/files/288c5be3ce7215f1b0e5ffa8ff524fefe419f862#diff-b67911656ef5d18c4ae36cb6741b7965R40) in Makefile provides the value for Version using ldflags. I agree documentation is bad.

PR #402, opened 2018-08-16
[RFC Don't merge yet] Fix concurrent map read/writes

tmarcu:

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -230,16 +230,18 @@ func resolveFiles(numWorkers int, set bundleSet, bundleRepoPkgs map[string]map[s
 
 	fileWorker := func() {
 		for bundle := range bundleCh {
+			var e error
 			fmt.Printf("processing %s\n", bundle.Name)
 			// Resolve files for this bundle, passing it the map of repos to packages
-			err = resolveFilesForBundle(bundle, bundleRepoPkgs[bundle.Name], packagerCmd)
-			if err != nil {
+			e = resolveFilesForBundle(bundle, bundleRepoPkgs[bundle.Name], packagerCmd)
+			if e != nil {
 				// break on the first error we get
 				// causes wg.Done to be called and the worker to exit
 				// the break is important so we don't overflow the errorCh
-				errorCh <- err
+				errorCh <- e
 				break
 			}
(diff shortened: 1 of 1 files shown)


builder/bundles.go
```diff
@@ -240,6 +243,8 @@ func resolveFiles(numWorkers int, set bundleSet, bundleRepoPkgs map[string]map[s
 				errorCh <- err
 				break
 			}
+			pkgMutex.Unlock()
```
matthewrsj: move this to before the error checking.

matthewrsj: Actually, this doesn't need a mutex, it just needs a different name for the `err` variable so all goroutines aren't using the same one.

builder/bundles.go
```diff
@@ -383,6 +389,7 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 					bundle.AllPackages[pkg] = true
 				}
 			}
+			pkgMutex2.Unlock()
```
matthewrsj: Don't even need these.

builder/bundles.go
```diff
@@ -230,16 +230,18 @@ func resolveFiles(numWorkers int, set bundleSet, bundleRepoPkgs map[string]map[s
...
+			e = resolveFilesForBundle(bundle, bundleRepoPkgs[bundle.Name], packagerCmd)
+			if e != nil {
 				// break on the first error we get
 				// causes wg.Done to be called and the worker to exit
 				// the break is important so we don't overflow the errorCh
-				errorCh <- err
+				errorCh <- e
 				break
 			}
+
```
matthewrsj: extra whitespace?

tmarcu: yeah I formatted with go fmt just now and repushed :) thx

builder/bundles.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+var pkgMutex sync.RWMutex
```
matthewrsj: Can you move this declaration closer to the function that uses it?

tmarcu: yes

Issue #348, opened 2018-06-13
Race condition in 'mixer build bundles'


cmarcelo:
Just creating a mix with init, then running repeatedly mixer build bundles was sufficient to trigger. Sometimes I get it in the first run.

See the "fatal error: concurrent map writes" snippet below.

```
Resolving packages using 4 workers
processing os-core
processing bootloader
processing os-core-update
processing kernel-native
... done with os-core
... done with os-core-update
fatal error: ... done with bootloader
concurrent map writes
```

and the offending goroutine:

```
goroutine 37 [running]:
runtime.throw(0x868c9a, 0x15)
	/usr/lib/golang/src/runtime/panic.go:616 +0x81 fp=0xc4202dac58 sp=0xc4202dac38 pc=0x42b691
runtime.mapassign_faststr(0x7f14c0, 0xc420169740, 0xc420538fab, 0xd, 0x0)
	/usr/lib/golang/src/runtime/hashmap_fast.go:779 +0x3ce fp=0xc4202dacc8 sp=0xc4202dac58 pc=0x40d82e
github.com/clearlinux/mixer-tools/builder.resolvePackages.func1()
	/builddir/build/BUILD/mixer-tools-4.6.5/.gopath/src/github.com/clearlinux/mixer-tools/builder/bundles.go:378 +0x41b fp=0xc4202dafe0 sp=0xc4202dacc8 pc=0x750ddb
runtime.goexit()
	/usr/lib/golang/src/runtime/asm_amd64.s:2361 +0x1 fp=0xc4202dafe8 sp=0xc4202dafe0 pc=0x457f01
created by github.com/clearlinux/mixer-tools/builder.resolvePackages
	/builddir/build/BUILD/mixer-tools-4.6.5/.gopath/src/github.com/clearlinux/mixer-tools/builder/bundles.go:392 +0x2de
```

Full trace:

```
$ mixer build bundles
Building bundles...
=> BUILD BUNDLES
Wiping away previous version 10...
Creating bundles for version 10 based on Clear Linux 23000
Preparing new /home/c/src/mix2/update/image/10
  and dnf config: /home/c/src/mix2/.yum-mix.conf
Packager command-line: dnf --config=/home/c/src/mix2/.yum-mix.conf -y --releasever=23000
Resolving packages using 4 workers
processing os-core
processing bootloader
processing os-core-update
processing kernel-native
... done with os-core
... done with os-core-update
fatal error: ... done with bootloader
concurrent map writes

[...]

PR #403, opened 2018-08-16
bundles: fix data race in building bundles

matthewrsj:
Two issues, first is resolved by renaming the error variable to a
goroutine isn't sharing err from outer scope with all other goroutines.

The second issue is resolved by locking concurrent writes on the same
map and using a sync.Map to safely read a map via the Range function.


matthewrsj:
Fixes #348

matthewrsj:
@tmarcu `s/rp/repoPkgMap/`

Issue #251, opened 2018-04-03
mixer build update merges paths incorrectly


mpurzynski:
When trying to update my mix from upstream version 21530 to 21630 I'm getting the following error (after having updated chroots, bumped a local version, etc).

ERROR: couldn't build update: failed to create update metadata: lstat /home/clr/mix/update/image/390/full/usr/share/systemd/language-fallback-map/usr/lib/systemd/system/multi-user.target.wants/prebuilt-trust-store.service: not a directory

Looks like two paths that have little to do with each other got merged?

A full log of my actions follows

root@clr-141f7bb770c54e568adf625071d8151f /home/clr/mix # mixer versions update --new-swupd
Current mix:      380
Current upstream: 21630 (format: 25)

Updated mix:      390
Updated upstream: 21630 (format: 25)

Wrote mixversion.
Wrote upstreamversion.
root@clr-141f7bb770c54e568adf625071d8151f /home/clr/mix # sudo mixer build chroots --new-chroots
Building bundles...
=> BUILD BUNDLES
Creating bundles for version 390 based on Clear Linux 21630
Preparing new /home/clr/mix/update/image/390
  and dnf config: /home/clr/mix/.yum-mix.conf
Packager command-line: dnf --config=/home/clr/mix/.yum-mix.conf -y --releasever=21630
Resolving packages using 2 workers
processing kernel-native
processing os-core
... done with kernel-native
processing os-core-update
... done with os-core-update
processing syslog-ng
... done with os-core
processing bootloader
... done with bootloader
processing bro
... done with bro
... done with syslog-ng
Resolving files using 2 workers
processing os-core-update
processing syslog-ng
Bundle os-core-update	2654 files
processing bootloader
Bundle bootloader	836 files
processing bro
Bundle bro	4856 files
processing kernel-native
Bundle syslog-ng	1446 files
processing os-core
Bundle kernel-native	3657 files
Bundle os-core	1974 files
Cleaning DNF cache before full install
Installing all bundles to full chroot
[1/6] bro
[2/6] kernel-native
[3/6] os-core
... building special os-core content
[4/6] os-core-update
... Adding swupd default values to os-core-update bundle
[5/6] syslog-ng
[6/6] bootloader

TIMINGS
  BUILD BUNDLES 25.193s
TOTAL: 25.193s
root@clr-141f7bb770c54e568adf625071d8151f /home/clr/mix # sudo mixer build update --new-swupd
=> CREATE MANIFESTS
[...]

PR #408, opened 2018-08-24
bundles: Enforce UTF-8 when running dnf

rchiossi:
In some situations dnf may try to combine files from different
encodings. If the command is not running is a UTF-8 locale, this
combination will be missing a new line, causing and error.

An example of this occurrence is the command `dnf repoquery --quiet -l
systemd ca-certs-static ...` which combines a file list from systemd and
ca-certs-static

This patch enforces UTF-8 locale on every dnf execution.

### DIFF
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -308,11 +308,19 @@ func RunCommandTimeout(timeout int, cmdname string, args ...string) error {
 // memory. If the command succeeds returns that output, if it fails, return err that
 // contains both the out and err streams from the execution.
 func RunCommandOutput(cmdname string, args ...string) (*bytes.Buffer, error) {
+	return RunCommandOutputEnv(cmdname, args, []string{})
+}
+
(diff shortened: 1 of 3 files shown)


rchiossi:
Fix issue #251

tmarcu:
It looks fine, but I'm going to try to reproduce the issue and make sure this fixes it before merging!

helpers/helpers.go
```diff
@@ -308,11 +308,19 @@ func RunCommandTimeout(timeout int, cmdname string, args ...string) error {
 // memory. If the command succeeds returns that output, if it fails, return err that
 // contains both the out and err streams from the execution.
 func RunCommandOutput(cmdname string, args ...string) (*bytes.Buffer, error) {
+	return RunCommandOutputEnv(cmdname, args, []string{})
```
matthewrsj: No `RunCommandSilentEnv` as well?

Issue #386, opened 2018-08-02
mixer init --offline returns an error


phmccarty:
I want to use mixer with local bundles only, so I tried running `mixer init --offline`, but it returned an error:

```
$ mixer init --offline
Creating new builder.conf configuration file...
Initializing mix version 10 from upstream version 24180
ERROR: Cannot find bundle "os-core" in local or upstream bundles
```

phmccarty:
This is starting from a clean workspace, by the way.

phmccarty:
@matthewrsj just suggested that I try `mixer init --offline --no-default-bundles`, and that worked. Maybe `--no-default-bundles` should be an implied setting in this case?

tmarcu:
I agree it should be implied on that setting, but in all other cases it doesn't make sense to try to init a workspace offline since it relies on bundles being available :P will update this ty!

PR #409, opened 2018-08-24
builder: No default bundles for offline init

rchiossi:
When running 'mixer init' with the --offline flag, there is no upstream
to fetch the default bundles from, so the mix should be initialized
without the default bundles.


rchiossi:
This PR addresses issue #386

PR #411, opened 2018-08-27
Introducing the Iterative to-Manifest

otaviobp:
When creating manifests for a new build, also create to-Manifests
for each bundle that has changes in this version.

The goal on publishing to-Manifests is to reduce the download size
on updates, according to what was discussed on
https://github.com/clearlinux/swupd-client/issues/498


otaviobp:
@tmarcu @matthewrsj: First version of the IManifest. Can you review that for me?

otaviobp:
v2:
 - Fix ContentSize header in IManifest

swupd/create_manifests.go
```diff
@@ -257,14 +257,51 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
...
-			appendTo.Files = append(appendTo.Files, f)
+		if f.Name == IndexBundle {
+			// this is generated new each time
+			continue
+		}
+		if f.Type == TypeManifest &&
+			f.findFileNameInSlice(appendTo.Files) != nil {
+			continue
+		}
+		if f.Type == TypeIManifest &&
```
matthewrsj: Since you are doing the same operation whether it is a Manifest or an IManifest, why not just define a method on f that reports if it is any manifest type, `f.IsManifest()` or something?

swupd/create_manifests.go
```diff
@@ -257,14 +257,51 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
...
+			f.findIManifestInSlice(appendTo.Files) != nil {
```
matthewrsj: If you do what I suggest above you don't have to use two different implementations to find the manifest in appendTo

swupd/create_manifests.go
```diff
@@ -257,14 +257,51 @@ func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
...
+
+// writeIterativeManifests writes Iterative Manifests for all bundles in
+// newManifests and add them to the MoM.
+// Iterative Manifests are manifests with all file changes added in versions
+// greater than fromVersion.
+func (MoM *Manifest) writeIterativeManifests(newManifests []*Manifest, out string, fromVersion uint32) error {
+	var err error
+	// write manifests then add them to the MoM
+	for _, bMan := range newManifests {
+		//Don't write I manifest for full and new bundles
```
matthewrsj: Please make your comment spacing consistent. Leave a space after the `//` like you do everywhere else.

matthewrsj: We might as well create an iterative manifest since the client will try to download it anyways. It not being there is just an extra network request the client has to make before it can get to work. If they are identical can we just link the two?

swupd/create_manifests.go
matthewrsj: I see you have to do some special work to determine an iterative manifest name. Two implementations may be the best way to handle that.

swupd/manifest.go
```diff
@@ -319,6 +319,32 @@ func (m *Manifest) WriteManifest(w io.Writer) error {
...
+// updated in versions greater than fromVersion
+func (m *Manifest) filterFilesVersion(fromVersion uint32) *Manifest {
+	fm := &Manifest{
+		Header:     m.Header,
+		Name:       m.Name,
+		BundleInfo: m.BundleInfo,
+	}
+	fm.Header.ContentSize = 0
+
+	for _, f := range m.Files {
```
matthewrsj: This function could be a lot simpler if you just sort the files by version then search for the first instance of the version and return files[firstInstance:]

swupd/manifest.go
```diff
@@ -319,6 +319,32 @@ func (m *Manifest) WriteManifest(w io.Writer) error {
...
+		if f.Version > fromVersion {
+			fm.appendFile(f)
+		}
+	}
+
+	for _, f := range m.DeletedFiles {
+		if f.Version > fromVersion {
+			fm.DeletedFiles = append(fm.DeletedFiles, f)
```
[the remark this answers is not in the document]
otaviobp: Append file appends the file to Files list and updates the ContentSize. As this is a standard behavior, I think it makes sense to have a function to do that. Btw, I'd update FileCount in this function and have all usages of append(Manifest.Files, ...) replaced by this call.
There's nothing to update in the Manifest for DeletedFiles, so I'm appending it to the list

swupd/create_manifests.go
otaviobp: We aren't creating a manifest for regular Manifests, so I don't think we need to create a link for IManifests. We're adding the IManifest to the MoM, so the client can download it from the older version folder like we do for regular Manifests.

swupd/manifest.go
otaviobp: This function is O(n) and sorting is O(n*log(n)) or worse. It's preferable to sort the list after filtering so we're going to sort a smaller list.

The ideal would be to use something similar to python filters. Is there something like that in GoLang?

    fm = [f if f.Version > fromVersion]

or

    fm = filter(lambda x: x.Version > fromVersion, f)

swupd/create_manifests.go
otaviobp: ok. thanks

otaviobp:
Fixed all suggestions that I didn't comment on. 
Master branch is failing on make lint, so this branch will fail too. Waiting for #418 to be merged.
I had some difficulties on running make lint locally, so I'm not sure if there's something else wrong. Working on that

swupd/manifest.go
tmarcu: You can use lambdas and simulate filters in Go, but it's really not efficient vs a regular for loop because it does a ton of extra work & copying:
https://github.com/robpike/filter
tl;dr rob pike also created Go and said don't use it. It will likely be much more doable once Go supports generics and this can be implemented more efficiently.

otaviobp: thanks. I prefer to let this function the way it is. If you prefer sorting + selecting for readability I can change that.

tmarcu: No I don't prefer sorting ahead of time, this looks ok. I think obsessively sorting is something we need to look at again in both mixer & swupd, 'cause I think there are some places where we do it too much "just in case."

matthewrsj:
Yes please let me review. I'll try to do it tonight or tomorrow morning.

swupd/create_manifests.go
matthewrsj: I think I'd prefer adding the argument like you mention just to simplify the logic in this function.

swupd/manifest.go
matthewrsj: There are probably cases where we do it too much just in case, but I think instead of removing operations on sorted slices we should consider 1) inserting elements in the correct location to generate a sorted list or 2) convert to using a set instead of a slice for this sort of thing. We should *never* have implemented these as slices.

swupd/manifest.go
```diff
@@ -319,6 +319,32 @@ func (m *Manifest) WriteManifest(w io.Writer) error {
...
+	}
+	fm.Header.ContentSize = 0
+
+	for _, f := range m.Files {
+		if f.Version > fromVersion {
+			fm.AppendFile(f)
+		}
+	}
+
+	for _, f := range m.DeletedFiles {
```
jwakre: Is this second for loop redundant? It looks like m.DeletedFiles is a subset of m.Files, so the deleted files are iterated over twice.

otaviobp: oh, I didn't noticed that. Good catch.

swupd/manifest.go
otaviobp: 1) I have a strong feeling that keeping the list sorted is worse than sorting later.
2) Assuming a Set is a hash table we're going to be faster on searches if we use that, but adding and sorting is going to be the same as using a list. I don't know how this structure is used in other areas of mixer to know if it's worth changing that.

otaviobp: I think I understood your point on (1) now. The point isn't just about replacing a sort by insertion on a sorted list. it is about not trying to sort it more than once. It makes sense in this case. But maybe (2) is even better if there are searches.

matthewrsj: I'm not sure about the usefulness of (2) in regards to searches. IIRC there are only a couple of times we need to search for something without iterating over the entire set anyways.

For many operations a sorted list is very important, such as when we `linkPeersAndChange`. We walk like that a couple times in the code so we should (at some point, not now) evaluate if it is better to sort once after list construction or if we should always insert in the correct place.

Part of that evaluation has to check if we create the slice, operate on the sorted slice, then insert again and operate again. If this is the case it makes sense to do correctly sorted inserts, at least for the second and further iterations.

matthewrsj:
Marking as blocked, needs a format bump in Clear Linux* OS.

swupd/manifest.go
tmarcu: If you need to sort often (i.e after every few inserts or deletes), then using a balanced tree to insert/remove quickly and keep things sorted as elements are added/removed is way better.
If we do many inserts/deletes and only require 1 sort at the end, then we'll get better performance just using a regular slice to add/remove unsorted and then call sort once when we're done.

otaviobp:
New version:
 - Adding tests
 - New commit: Don't create packs for Iterative Manifests
 - Use last version of that bundle instead of the previous version of the build to create a Iterative Manifest, i.e., if you have a bundle test1 that was added in 10, and updated in 50, the iterative Manifest will be called test1.I.10 and not test1.I.40 (Header Previous will also be 10)

otaviobp:
New version:
 - Create tarballs for Iterative Manifests. This was a bug I haven't realized.
 - Rebased on master

otaviobp:
Changes in this version:
 - New commit: Don't create Iterative Manfests on minversions
 - Adding delta manifests commits to this pull request so we can have one branch with both changes for tests before merging.

matthewrsj:
Needs a rebase to resolve conflicts.

swupd/create_manifests.go
```diff
@@ -429,15 +470,25 @@ func CreateManifests(version uint32, minVersion uint32, format uint, statedir st
...
+			return nil, err
+		}
+		newManifests = append(newManifests, iManifests...)
+	}
+
+	// copy over unchanged manifests
+	addUnchangedManifests(&newMoM, oldMoM, groups)
+
 	var osIdx *Manifest
 	if osIdx, err = writeIndexManifest(&c, &ui, allManifests); err != nil {
```
jwakre: @otaviobp Should the index manifest have an iterative manifest? Currently, a delta manifest for the index manifest is created, but an iterative manifest is not.

jwakre: This isn't going to break anything and could be added later as a future enhancement.

otaviobp:
rebased and fixed issue @jwakre reported.

otaviobp:
I tested this branch with current swupd and it's working fine. No breaks. There's no blocker from my side to have this merged.

@matthewrsj commented about setting a format barrier for this code. @jwakre Can you work on that?

jwakre:
@otaviobp Yes, I can do that.

Also, I used the validate-swupd.sh script to test this branch with swupd's delta/iterative manifest branch (master_new_manifests). This passed, but I have not yet verified whether swupd successfully consumed the delta/iterative manifests.

* Note: validate-swupd.sh script was run with the desktop bundle instead of --all-upstream.

otaviobp:
> Note: validate-swupd.sh script was run with the desktop bundle instead of --all-upstream.
￼￼ 1  

Make sure you are using those two to install too, because the standard script is only installing a couple of bundles

jwakre:
> Make sure you are using those two to install too, because the standard script is only installing a couple of bundles

Yes, actually I forgot to add the desktop bundle to swupd when I tested earlier... Although, now I am running swupd manually on the update content that I generated earlier with mixer and I am including the bundle. I might have found an iterative manifest swupd bug, but I need to look a little deeper to confirm that.

builder/update.go
```diff
@@ -189,6 +189,9 @@ func (b *Builder) buildUpdateContent(params UpdateParameters, timer *stopWatch)
 		timer.Start("CREATE ZERO PACKS")
 		bundleDir := filepath.Join(b.Config.Builder.ServerStateDir, "image")
 		for _, bundle := range mom.Files {
+			if bundle.Type != swupd.TypeManifest {
+				continue
+			}
 			// TODO: Evaluate if it's worth using goroutines.
```
matthewrsj: this is NOT in the scope of this PR, but this really *is* worth using goroutines for. Need to evaluate if `swupd.CreatePack` is threadsafe, which I believe it is.

swupd/create_manifests.go
```diff
@@ -270,24 +271,64 @@ func processBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error
...
-			continue
-		}
-
 		if f.Name == IndexBundle {
 			// this is generated new each time
 			continue
 		}
 
+		bundleName := f.Name
+		if f.Type == TypeIManifest {
```
matthewrsj: This can all be rewritten as
```go
if f.findFileNameInSlice(appendTo.Files) != nil {
    continue
}
bundleName := f.Name
if f.Type == TypeIManifest {
    bundleName = strings.Split(bundleName, ".I.")[0]
}
```

swupd/helpers_test.go
```diff
@@ -270,13 +295,30 @@ func checkFileInManifest(t *testing.T, m *Manifest, version uint32, name string)
...
 	t.Errorf("couldn't find file %s in manifest %s version %d", name, m.Name, m.Header.Version)
 }
 
+func fileInManifestHash(t *testing.T, m *Manifest, version uint32, name string, hash string) *File {
+	t.Helper()
+	for _, f := range m.Files {
+		if f.Name == name {
+			if f.Version != version {
+				t.Fatalf("in manifest %s version %d: file %s has version %d but expected %d", m.Name, m.Header.Version, f.Name, f.Version, version)
+			} else if f.Hash.String() != hash {
```
matthewrsj: logic seems weird here, why does the hash check depend on the version check preceding it? YMMV and this is just a style issue but it seems like this would read better as two separate conditionals instead of an `if/else if`

swupd/packs.go
```diff
@@ -113,6 +113,26 @@ func CreateAllDeltas(outputDir string, fromVersion, toVersion, numWorkers int) e
...
+func writeDeltaManifest(tw *tar.Writer, outputDir string, dManifest *Manifest, toVersion uint32) error {
+	if dManifest == nil || len(dManifest.Files) == 0 {
+		return nil
+	}
+
+	dManifest.Header.FileCount = uint32(len(dManifest.Files))
+	dManifestPath := filepath.Join(outputDir, fmt.Sprint(toVersion),
+		fmt.Sprintf("Manifest.%s.D.%d", dManifest.Name, dManifest.Header.Version))
+
+	if err := dManifest.WriteManifestFile(dManifestPath); err != nil {
```
matthewrsj: hmm, we should have made manifests an interface when we started having different types of them... not necessary for this PR, but it could be updated later.

swupd/format_switch_test.go
```diff
@@ -1,21 +1,27 @@
...
 
 import (
+	"strings"
 	"testing"
 )
 
-// Format 25 should not support minversions or iterative manifests. Support for
-// these features was added in format 26.
+// Format 25 should not support minversions, iterative manifests, or delta manifests.
+// Support for these features was added in format 26.
```
matthewrsj: This test is becoming a pretty big test and is testing several different format-switch features. Can we split it out like was done for the contentsize test below?

jwakre: Ya, I can split these out

swupd/format_switch_test.go
```diff
@@ -85,14 +69,79 @@ func TestManifestFormats25to26(t *testing.T) {
...
+	ts.createManifests(10)
+
+	ts.addFile(20, "test-bundle", "/foo", "new content")
+	ts.createManifests(20)
+	checkManifestContains(t, ts.Dir, "20", "MoM", "MANIFEST\t25")
+
+	// Iterative manifests should not have entries in the MoM or be generated
+	checkManifestNotContains(t, ts.Dir, "20", "MoM", "I...\t")
+	ts.checkNotExists("www/20/Manifest.test-bundle.I.10")
+	ts.checkNotExists("www/20/os-core.I.10")
```
matthewrsj: Needs to be `"www/20/Manifest.os-core.I.10"`, you are missing the "Manifest." part of the name.

jwakre: Thanks, I pushed the fix.

jwakre:
I used the validate-swupd.sh script (with small bundle subset) to generate content with this branch.

* Iterative/delta manifests are not generated in format 25
* Iterative/delta manifests are generated in format 26
* Swupd in Clear 26410 successfully updates with generated content.

swupd/helpers_test.go
tmarcu: Yes I'd like a comment on this block if possible.

swupd/packs.go
```diff
@@ -212,6 +233,19 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
...
+	if fromManifest != nil {
+		dManifest = &Manifest{
+			Header: fromManifest.Header,
+			Name:   fromManifest.Name,
+		}
+		dManifest.Header.ContentSize = 0
+		// Removing Includes from Header because the include list isn't necessary
+		// by the client to find which file is going to be used to apply deltas
+		// As we won't need a format bump to add that later if we decide to use
+		// it we can leave Includes empty to save some bytes when downloading
```
tmarcu: This is a hard sentence to read but the content is good.

swupd/create_manifests.go
matthewrsj: I just noticed this didn't get simplified as I requested. Can that be done before merge please?

swupd/packs_test.go
```diff
@@ -643,22 +742,85 @@ func TestPackRenames(t *testing.T) {
...
+
+	// Check Delta Manifest 10->40
+	{
+		m := checkDeltaManifest(ts, 10, 40, "os-core", 1)
+		fileInManifestHash(t, m, 10, "/file1", hashIn10)
+	}
+}
+
+// TestPackNoDeltas will test cases where there are no delta files are
+// created in the update
```
tmarcu: s/are//

swupd/create_manifests.go
jwakre: @matthewrsj There is an problem with this suggested change:

This change can add stale iterative manifests that are no longer valid. For instance, if the FROM MoM has an entry for bundle1.I.10 and this bundle was updated, then the TO MoM would have entries for bundle1.I.10 and bundle1.I.20. In this case, bundle1.I.10 would no longer be valid and should be omitted from the TO MoM.

Would it be sufficient to add a comment here or do I need to look at refactoring this implementation?

otaviobp: I created a generic findBundleInManifest to look for bundle names in both IManifest and Manifests. This function won't work for other files, so I'm returning nil in that case. Not sure if it's really better than use the version before that, but I'd let you decide. If you have any other implementation ideas, we can rewrite that too.

swupd/files.go
```diff
@@ -271,11 +271,18 @@ func (f *File) findFileNameInSlice(fs []*File) *File {
...
+		return nil
+	}
+
+	name_split := func(str string) string {
+		return strings.SplitAfter(str, ".I.")[0]
+	}
+
+	name := name_split(f.Name)
+	for _, file := range manifest.Files {
+		if name == name_split(file.Name) {
```
tmarcu: I think the original is much more performant just checking the prefix and not doing a `SplitAfter()` on _every_ bundle.

swupd/files.go
```diff
@@ -271,11 +271,18 @@ func (f *File) findFileNameInSlice(fs []*File) *File {
...
-	// get bundle name without IManifest version
-	prefix := strings.SplitAfter(f.Name, ".I.")[0]
-	for _, file := range fs {
-		if strings.HasPrefix(file.Name, prefix) {
+func (f *File) findBundleInManifest(manifest *Manifest) *File {
+	if f.Type != TypeManifest && f.Type != TypeIManifest {
+		return nil
+	}
+
+	name_split := func(str string) string {
```
tmarcu: This is not idiomatic function naming in Go, lint is probably crying about it. I don't think you need it since you will only call once, otherwise update to `nameSplit()`.

swupd/create_manifests.go
matthewrsj: You're right, totally thought those `f.funcName1` and `f.funcName2` were the same functions...

PR #414, opened 2018-08-30
Introducing the Delta-Manifests

otaviobp:
Currently, when updating from version X to Y, in order to apply a delta, swupd
needs to download the Manifest from version X to check which file the
from-hash in delta file is pointing to. Downloading that Manifest some times
is even larger than the delta pack, so reducing the download size would make
a significant improvement in update time.

The Delta-Manifest is a Manifest included in delta packs with all information
from Manifest X in order to apply deltas, i.e., all files that are going to be
used to apply deltas.


tmarcu:
Need to run make lint on this & re-push.

matthewrsj:
Note: not blocked by format bump in Clear Linux* OS, since swupd will not break on these extra files existing.

otaviobp:
New version:
 - add tests
 - Fix Header of the Delta Manifest (Version and Previous were incorrect)

swupd/packs.go
```diff
@@ -232,10 +242,27 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
...
 		}
 
+		dManifest.Files = append(dManifest.Files, d.from)
 		info.DeltaCount++
 		hasDelta[d.to.Hash] = d
 	}
 
+	if fromManifest != nil {
+		dManifest.Header.FileCount = uint32(len(dManifest.Files))
+		if dManifest.Header.FileCount > 0 {
```
matthewrsj: I think it would be easier to read and you save an indentation level if you turn this into a guard instead of a nested if, so
```go
if dManifest.Header.FileCount <= 0 {
    continue
}
...
```
but ymmv

swupd/packs.go
```diff
@@ -232,10 +242,27 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
...
+			dManifestPath := filepath.Join(outputDir, fmt.Sprint(toVersion), "delta",
+				fmt.Sprintf("Manifest.%s.D.%d", dManifest.Name, fromVersion))
+
+			if err = dManifest.WriteManifestFile(dManifestPath); err != nil {
+				return nil, err
+			}
+
+			if _, err = copyFromPath(tw, dManifestPath, "delta/"); err != nil {
```
matthewrsj: Not entirely sure I love throwing this in `delta/` and I'm not sure `swupd` will handle this gracefully enough. Since `swupd` iterates through everything in `<statedir>/delta` I believe it will print warnings that it `couldn't use delta <delta manifest>` for every delta manifest it gets. Can we put the delta manifest in the top level of the delta pack instead?
```
delta/
    <delta files>
staged/
    <full files>
Manifest.os-core.D.24710
```

swupd/packs_test.go
```diff
@@ -637,22 +687,46 @@ func TestPackRenames(t *testing.T) {
...
 	mustHaveDeltaCount(t, info, 1)
 
 	// Note that the delta refers to the version of the file, which is still 30.
 	checkFileInPack(t, ts.path("www/40/pack-os-core-from-10.tar"), fmt.Sprintf("delta/10-30-%s-%s", hashIn10, hashIn30))
+
+	// Check Delta Manifest 10->40
+	{
+		m := checkDeltaManifest(ts, 10, 40, "os-core", 1)
+		fileInManifestHash(t, m, 10, "/file1", hashIn10)
+	}
```
matthewrsj: These are good tests, can you please add a few negative cases as well? We want to make sure the delta manifest contains only those files that were used as from deltas and no other files.

swupd/packs.go
```diff
@@ -212,6 +213,15 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
 		fullChrootDir = filepath.Join(chrootDir, fmt.Sprint(toVersion), "full")
 	}
 
+	if fromManifest != nil {
+		dManifest = &Manifest{
+			Header: toManifest.Header,
```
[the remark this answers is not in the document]
otaviobp: Hmmm. I'm not sure what's the best approach now.
If we are creating a delta pack from 30 to 40 we are going to create a Manifest.bundle.D.30. What should be the version of the file? 30? And what about the Previous?

In this commit the header will be like this:
version:    40
previous:   30

If I use from:
version:    30
previous:   20 (Or some other version < 30)

Btw, Includes aren't useful for delta manifests, right? I'm going to take them out.

swupd/packs.go
otaviobp: Good point. But if we do that, we are going to have Delta manifests listed in /var/lib/swupd. Is this a problem?

And I realized by your comment that we have another issue with this PR. We may have, for example, Manifest.os-core.D.10 in version 20 and 30. If we download them both (I don't see a reason to do that, but this may happen in some combinations of swupd updates), one file will replace the other because they are extracted in the same directory. What about renaming them to Manifest.<bundle>.D.<from_version>.<to_version>.

I'll check if swupd isn't breaking with this PR and if we need a format bump. I've done that with IManifests, but not for delta manifests.

matthewrsj: So currently individual manifests are placed under `/var/lib/swupd/<version>/Manifest.<bundle>`, so you could potentially do the same thing here. The packs would then be constructed by creating a new directory `<version>` containing the delta manifest. This way it will be sent to the correct location on the client when extracted.

I'd prefer that approach to placing them at the top-level because it keeps the structure of the cache nicer and places the delta manifests alongside the full manifests.

On the client side these delta manifests should probably be cleaned up on updates since they are no longer valid after being used (if the client updated).

otaviobp: I was wondering a good way to do that on swupd, but it makes a lot of sense to create a <version> folder on the pack. Good idea.

So packs content (for your example) should be like this:

```
delta/
    <delta files>
staged/
    <full files>
25720/
     Manifest.os-core.D.24710
```

swupd/packs_test.go
otaviobp: True. I'll check the number of files in the manifest.

swupd/packs.go
matthewrsj: Correct

otaviobp:
New version with following updates:
 - checking the number of files and deleted files in Delta manifest tests to make sure no unneeded file was added
 - Don't save Delta Manifest in delta/ folder. Save it in the root directory of the version update (for example, update/www/10/) and in a directory with the version name in the tarball, like we discussed in the comments.
 - Took includes out of the Delta Manifests and add a test to check that
 - Use Header version and previous from FromManifest and not from ToManifest as suggested.
 - Broke up long lines with prints

swupd/packs.go
otaviobp: This isn't inside a loop so I can't use continue. I don't see how I can reduce the number of nested ifs. Any suggestion?

rchiossi: Do you actually reuse this object anywhere else other than here? If not, you don't really need to set `dManifest.Header.FileCount` in the line above when `len(dManifest.Files)` is 0 and you can merge both ifs into a single one.

Also note that this function is kinda big. It wouldn't be bad to put this into a secondary function so you are able to use a `return` statement.

swupd/packs.go
```diff
@@ -212,6 +233,15 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
 		fullChrootDir = filepath.Join(chrootDir, fmt.Sprint(toVersion), "full")
 	}
 
+	if fromManifest != nil {
+		dManifest = &Manifest{
+			Header: fromManifest.Header,
+			Name:   fromManifest.Name,
+		}
+		dManifest.Header.ContentSize = 0
+		dManifest.Header.Includes = nil
```
matthewrsj: Please add that comment I asked for describing why we are setting this to `nil` and point out the fact that if we ever need to add it back in it does *not* require a format bump for the client.

swupd/packs.go
```diff
@@ -113,6 +113,26 @@ func CreateAllDeltas(outputDir string, fromVersion, toVersion, numWorkers int) e
...
+
+	dManifest.Header.FileCount = uint32(len(dManifest.Files))
+	dManifestPath := filepath.Join(outputDir, fmt.Sprint(toVersion),
+		fmt.Sprintf("Manifest.%s.D.%d", dManifest.Name, dManifest.Header.Version))
+
+	if err := dManifest.WriteManifestFile(dManifestPath); err != nil {
+		return err
+	}
+
+	if _, err := copyFromPath(tw, dManifestPath, fmt.Sprintf("%d/", toVersion)); err != nil {
```
matthewrsj: Do you really need the trailing slash here?

swupd/packs.go
```diff
@@ -221,7 +251,7 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
 			continue
 		}
 		var fallback bool
-		fallback, err = copyFromDelta(tw, d)
+		fallback, err = copyFromPath(tw, d.Path, "delta/")
```
matthewrsj: Why add the trailing slash?

swupd/packs.go
```diff
@@ -314,13 +349,13 @@ func copyFromDelta(tw *tar.Writer, delta *Delta) (fallback bool, err error) {
...
 	if !fi.Mode().IsRegular() {
-		return true, fmt.Errorf("delta %s is not a regular file", delta.Path)
+		return true, fmt.Errorf("file %s is not a regular file", path)
 	}
 	hdr, err := getHeaderFromFileInfo(fi)
 	if err != nil {
 		return true, err
 	}
-	hdr.Name = "delta/" + fi.Name()
+	hdr.Name = pathPrefix + fi.Name()
```
matthewrsj: `filepath.Join` here and you don't have to append `/` everywhere.

otaviobp: nice catch

otaviobp:
New version:
  - Adding comment to NULL Include list
  - Use filepath.Join to contact paths

otaviobp:
Closing this PR as patches were included now on #411

PR #415, opened 2018-08-30
lint: Fix lint errors

rchiossi:
These errors are reported by the current version of gometalinter. Aside
from the errors in mixer code, it is also reporting errors in the
standard golang libraries, so golang/src path was excluded from the lint
errors in the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,7 @@ lint: gopath
 	--enable=unused \
 	--enable=vetshadow \
 	--enable=errcheck \
+	--exclude=golang/src \
 	./...
 
 clean:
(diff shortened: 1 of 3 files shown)


tmarcu:
@rchiossi I merged the one that was above this from auke first, which has the 2 linter changes. Would you mind changing this to just include the exclude change only? 👍

rchiossi:
Sure, I'll do it once I get back.

tmarcu:
Once I do gometalinter.v2 --install, it no longer tries to look into the golang src path

rchiossi:
Ah nice. Then I'll close this PR.

PR #416, opened 2018-08-30
Ignore possible %ghost files in RPM files.

ahkok:
Due to the way that we use repoquery, we are returned files that
may be %ghost files in RPMs. These files may be missing and this
is legal and compliant.

While Clear Linux OS does not use %ghost files in any of the RPMs
created, there is nothing wrong with permitting other vendor RPM
files that may list %ghost files, and mixer should just ignore
these files.

This change does just that: If repoquery returns files that are
not existing in the fullimage, prints a warning and continues with
out dropping the manifest for that bundle.

### DIFF
--- a/swupd/bundleinfo.go
+++ b/swupd/bundleinfo.go
@@ -94,6 +94,10 @@ func (m *Manifest) addFilesFromBundleInfo(c config, version uint32) error {
 	for fpath := range m.BundleInfo.Files {
 		fullPath := filepath.Join(chrootDir, fpath)
 		fi, err := os.Lstat(fullPath)
+		if os.IsNotExist(err) {
+			fmt.Fprintf(os.Stderr, "Warning: Missing file, assuming %%ghost: %s\n", fpath)
+			continue
+		}
 		if err != nil {
 			return err
 		}


ahkok:
#418 contains fixes for the travis build problems.

swupd/bundleinfo.go
```diff
@@ -94,7 +94,12 @@ func (m *Manifest) addFilesFromBundleInfo(c config, version uint32) error {
 	for fpath := range m.BundleInfo.Files {
 		fullPath := filepath.Join(chrootDir, fpath)
 		fi, err := os.Lstat(fullPath)
+		if os.IsNotExist(err) {
+			fmt.Fprintf(os.Stderr, "Warning: Missing file, assuming %%ghost: %s\n", fpath)
+			continue
+		}
 		if err != nil {
+			fmt.Println(err, err.Error())
```
rchiossi: This is not the proper way of passing the error to the user. This message error is not showing currently because of a bug on initBundles(), which is the caller of addFilesFromBundleInfo(). 
Patch #417 addresses this issue.

rchiossi: Also, is there any other possible error tho Lstat other than IsNotExist? I wonder if this block will ever be executed. (I did a quick search but I couldn't find what are the possible PathErrors that can be returned by Lstat)

swupd/bundleinfo.go
```diff
@@ -94,7 +94,12 @@ func (m *Manifest) addFilesFromBundleInfo(c config, version uint32) error {
 	for fpath := range m.BundleInfo.Files {
 		fullPath := filepath.Join(chrootDir, fpath)
 		fi, err := os.Lstat(fullPath)
+		if os.IsNotExist(err) {
+			fmt.Fprintf(os.Stderr, "Warning: Missing file, assuming %%ghost: %s\n", fpath)
```
rchiossi: Is this safe for a quick hack? The correct solution should filter out the %ghost files before they are added to the bundle's file list.
If this doesn't break anything, I'm ok with it, but it would be nice to also have a TODO comment explaining what the proper fix should be.

swupd/bundleinfo.go
ahkok: We should really pass an error here, I'm open to improvements.

And yes, os.Lstat can return lots of errors, like ENOPERM, even ENOMEM (stat allocates a statbuf internally).

swupd/bundleinfo.go
ahkok: This is reasonably safe. The only unsafe path would be if people manually muck around the fullimage, but that is already out of scope - nothing protects against that.

We already catch all dnf errors anyway.

swupd/bundleinfo.go
rchiossi: Ah, indeed. Still, the Print should not be here or it will be printed twice with the fix to initBundles

swupd/bundleinfo.go
rchiossi: Alright, then lets do it!

ahkok:
Revised, taken out the second print.

PR #417, opened 2018-08-30
Manifest error

rchiossi:
This patch fix the error handling issue in manifest creation and change it so the number of worker threads matches the one defined by --bundle-workers option. See commit message for detailed description on the issue.

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -53,7 +53,6 @@ func getOldManifest(path string) (*Manifest, error) {
 }
 
 func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
-	var err error
 	var wg sync.WaitGroup
 	workers := len(ui.bundles)
 	wg.Add(workers)
@@ -64,6 +63,7 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 	fmt.Println("Generating initial manifests...")
 	bundleWorker := func() {
 		defer wg.Done()
+		var err error
 		for bundleName := range bundleChan {
 			bundle := &Manifest{
 				Header: ManifestHeader{
@@ -101,7 +101,7 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 				err = bundle.GetBundleInfo(c.stateDir, biPath)
 				if err != nil {
 					errorChan <- err
-					break
+					return
 				}
 
 				if useBundleInfo {
@@ -110,14 +110,14 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 			}
 			if err != nil {
 				errorChan <- err
-				break
+				return
 			}
 
 			// detect type changes
 			// fail out here if a type change is detected since this is not yet supported in client
 			if bundle.hasUnsupportedTypeChanges() {
 				errorChan <- errors.New("type changes not yet supported")
-				break
+				return
 			}
 
 			// remove banned debuginfo if configured to do so
@@ -136,24 +136,29 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 		go bundleWorker()
 	}
 
+	var err error
 	for _, bn := range ui.bundles {
 		select {
 		case bundleChan <- bn:
 		case err = <-errorChan:
-			// break as soon as we see a failure
-			break
+			// Closing the channel here prevents workers from start processing
+			// a new Bundle while mixer is exiting.
+			close(bundleChan)
+
+			// If there is an error, there is no need to wait for the workers
+			// to finish since mixer will exit with that error, so we skip
+			// wg.Wait() here and return right away. Exiting mixer will clean
+			// any running worker.
+			return nil, err
 		}
 	}
+
 	close(bundleChan)
 	wg.Wait()
 
 	if err == nil && len(errorChan) > 0 {
 		err = <-errorChan
-	}
-
-	chanLen := len(errorChan)
-	for i := 0; i < chanLen; i++ {
-		<-errorChan
+		return nil, err
 	}
 
 	// Now handle the full manifest last, we know the full chroot is populated
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sync"
 	"time"
 )
@@ -52,12 +53,18 @@ func getOldManifest(path string) (*Manifest, error) {
 	return ParseManifestFile(path)
 }
 
-func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
+func initBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error) {
 	var wg sync.WaitGroup
-	workers := len(ui.bundles)
-	wg.Add(workers)
+	if numWorkers <= 0 {
+		numWorkers = runtime.NumCPU()
+	}
+	if numWorkers > len(ui.bundles) {
+		numWorkers = len(ui.bundles)
+	}
+	wg.Add(numWorkers)
+
 	bundleChan := make(chan string)
-	errorChan := make(chan error, workers)
+	errorChan := make(chan error, numWorkers)
 	mux := &sync.Mutex{}
 	tmpManifests := []*Manifest{}
 	fmt.Println("Generating initial manifests...")
@@ -132,7 +139,7 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 		}
 	}
 
-	for i := 0; i < workers; i++ {
+	for i := 0; i < numWorkers; i++ {
 		go bundleWorker()
 	}
 
@@ -182,11 +189,11 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 	return tmpManifests, err
 }
 
-func processBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
+func processBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error) {
 	var newFull *Manifest
 	var err error
 	// initialize bundles with with all files and their info
-	tmpManifests, err := initBundles(ui, c)
+	tmpManifests, err := initBundles(ui, c, numWorkers)
 	if err != nil {
 		return nil, err
 	}
@@ -327,7 +334,7 @@ func aggregateManifests(newManifests []*Manifest, newMoM *Manifest, version uint
 }
 
 // CreateManifests creates update manifests for changed and added bundles for <version>
-func CreateManifests(version uint32, minVersion uint32, format uint, statedir string) (*MoM, error) {
+func CreateManifests(version uint32, minVersion uint32, format uint, statedir string, numWorkers int) (*MoM, error) {
 	var err error
 	var c config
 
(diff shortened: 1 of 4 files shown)


rchiossi:
Why what?

tmarcu:
Why can we not use one goroutine per bundle here? That change is completely irrelevant to the issue you're trying to fix in the first place here & not needed.

swupd/create_manifests.go
```diff
@@ -64,6 +63,7 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
 	fmt.Println("Generating initial manifests...")
 	bundleWorker := func() {
 		defer wg.Done()
+		var err error
```
tmarcu: This is good +1

swupd/create_manifests.go
```diff
@@ -110,14 +110,14 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
...
-				break
+				return
 			}
 
 			// detect type changes
 			// fail out here if a type change is detected since this is not yet supported in client
 			if bundle.hasUnsupportedTypeChanges() {
 				errorChan <- errors.New("type changes not yet supported")
-				break
+				return
```
tmarcu: Kind-of +1 to these. It doesn't matter if we just break out of processing this or not IMO, cause if we write an error it should exit in the main routine anyway. BUT, this is fine here, maybe add a comment explicitly saying that the goroutine itself is no longer usable, which is not the case we want in other code paths which should continue on error.

swupd/create_manifests.go
```diff
@@ -136,14 +136,22 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
...
 	}
 
-	for _, bn := range ui.bundles {
-		select {
-		case bundleChan <- bn:
-		case err = <-errorChan:
-			// break as soon as we see a failure
-			break
+	// Generator
+	var err error
```
tmarcu: I don't think this is the right way to do this, I think a simpler method could be used - **as long as the intended result is to quit execution of mixer if something fails here:**
```
var err error
// Exit earliy if we get an error
go func() {
	for e := range errorChan {
		fmt.Println(e)
		os.Exit(1)
	}
}()

for i := 0; i < workers; i++ {
	bundleChan <- i
}
close(bundleChan)
wg.Wait()
close(errorChan) // So we don't deadlock in error goroutine if we didn't get any errors
```
Even in the goroutine you start for adding bundles to bundleChan, it will never hit the error case because there are always goroutines available to pick up a bundle.

swupd/create_manifests.go
rchiossi: You don't break out of processing, you just move to the next bundle. But once an error occurs, mixer should stop.

swupd/create_manifests.go
rchiossi: It does work, the routines stop getting bundles once one of them sends an error. The problem before was that break was a no-op since it was breaking the select instead of the for loop. But this code works as intended.

rchiossi:
The reason I added the variable number of workers is because the code was designed as such, so I believe the original intention of the author was to have something like this. Otherwise, it makes no sense to loop over bundleChan if every worker will only ever get a single bundle. If having one worker per bundle is what we want, than the patch should be completely different. I've played with it as well, I have a draft ready of what it should look like.

swupd/create_manifests.go
tmarcu: Ugh, you're right https://tip.golang.org/ref/spec#Break_statements
It never actually breaks out of the for in the original code. Return works since you're spawning off a routine, but perhaps we could do it without a waitgroup or extra goroutine if we follow this code ^ and do 
```
BUNDLE_LOOP:
	for _, bn := range ui.bundles {
		select {
		case bundleChan <- bn:
		case err = <- errorChan:
			break BUNDLE_LOOP
		}
	}
```

swupd/create_manifests.go
```diff
@@ -136,14 +136,20 @@ func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
...
+		if done {
+			break
+		}
+
 		select {
 		case bundleChan <- bn:
 		case err = <-errorChan:
-			// break as soon as we see a failure
-			break
+			done = true
```
tmarcu: 🙌 This looks much cleaner now and actually does what it's saying

swupd/create_manifests.go
```diff
@@ -52,18 +52,23 @@ func getOldManifest(path string) (*Manifest, error) {
 	return ParseManifestFile(path)
 }
 
-func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
-	var err error
+func initBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error) {
 	var wg sync.WaitGroup
-	workers := len(ui.bundles)
-	wg.Add(workers)
+	if numWorkers > len(ui.bundles) {
```
matthewrsj: what if `numWorkers == 0`?

swupd/create_manifests.go
```diff
@@ -52,18 +52,23 @@ func getOldManifest(path string) (*Manifest, error) {
...
+		numWorkers = len(ui.bundles)
+	}
+	wg.Add(numWorkers)
+
 	bundleChan := make(chan string)
-	errorChan := make(chan error, workers)
+	errorChan := make(chan error, numWorkers)
 	mux := &sync.Mutex{}
 	tmpManifests := []*Manifest{}
+
```
matthewrsj: just adding newlines?

swupd/create_manifests.go
```diff
@@ -52,18 +52,23 @@ func getOldManifest(path string) (*Manifest, error) {
...
 	fmt.Println("Generating initial manifests...")
+
 	bundleWorker := func() {
 		defer wg.Done()
+		var err error
```
matthewrsj: +1

swupd/create_manifests.go
rchiossi: it is never 0 since that condition is already handled at program start. If it is 0 mixer will crash in multiple points, it won't even make it this far.

swupd/create_manifests.go
rchiossi: It looked a bit messy, I added new lines for clarity

swupd/create_manifests.go
```diff
@@ -322,7 +332,7 @@ func aggregateManifests(newManifests []*Manifest, newMoM *Manifest, version uint
 }
 
 // CreateManifests creates update manifests for changed and added bundles for <version>
-func CreateManifests(version uint32, minVersion uint32, format uint, statedir string) (*MoM, error) {
+func CreateManifests(version uint32, minVersion uint32, format uint, statedir string, numWorkers int) (*MoM, error) {
```
matthewrsj: Ok, you mentioned that numWorkers will never be 0, yet it could very easily be set to 0 by the caller of this external function. This exists as a library intended to be usable outside of the mixer framework. You must ensure that numWorkers is not set to 0 in this codepath.

swupd/create_manifests.go
tmarcu: So this is covered by setWorkers(), which checks if builder.<type>NumWorkers is < 1, and then sets it to numCPU() if it is. I'm not sure if we have to check that again in here, since this function is likely not going to get called by any other thing individually.
I would have like numWorkers to be able to be > len(ui.Bundles), but since we actually do IO here we probably can't do that.

matthewrsj: OK, but see my additional review comment on `CreateManifests`. This is not the only way to get this point.

rchiossi: E.g.: resolveFiles() and resolvePackages() inside buildBundles() also rely on workers > 0, but they are executed way earlier in the build process.

matthewrsj: that's fine, those are all internal functions not exposed as part of the API.

swupd/create_manifests.go
rchiossi: Fair enough. Al thought the same can be said about BuildBundles() which is exported in the build package.
Ill add the check and reupload.

matthewrsj: I wouldn't use the `build` subpackage as a good example of a package that follows good API practices. I think it is more intended as an internal package than an external one 😕 

The swupd subpackage on the other hand should try to follow good practices for a useful API. I'm sure it **never** violates that principle. 😜

swupd/create_manifests.go
rchiossi: Not directly, but they are exposed by the builder package and there is no check in the execution path.

swupd/create_manifests.go
rchiossi: haha I agree with you. I'm just explaining my reasoning behind the missing check. Ideally we would have checks for all those cases.

swupd/create_manifests.go
matthewrsj: I have issues with the builder package. I think it should be an internal package instead of exposed at all, since it is only intended to be used via the mixer cmdline. 🤷‍♂️ But yes, it definitely violates the principles of a good external API (but probably shouldn't be one).

swupd/create_manifests.go
```diff
@@ -52,12 +52,15 @@ func getOldManifest(path string) (*Manifest, error) {
 	return ParseManifestFile(path)
 }
 
-func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
+func initBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error) {
 	var wg sync.WaitGroup
-	workers := len(ui.bundles)
-	wg.Add(workers)
+	if numWorkers > len(ui.bundles) || numWorkers == 0 {
```
matthewrsj: `numWorkers` is a signed `int` so you need to check if it is `<= 0` not just `== 0`.

matthewrsj: although if it is `<= 0` you might want to set to `1` instead of `len(ui.bundles)`

rchiossi: hmm, maybe I should just change the type instead. It makes no sense having numWorkers < 0

matthewrsj: sure, that would make sense, but you'll have to change it in the entire call path from the mixer cmdline as well if you do that.

rchiossi: No problem. It looks better overall.

matthewrsj: ok +1

rchiossi: hmm, nevermind... doing so would require me to manually convert the result of runtime.NumCPU() to uint in multiple points of the code. It will be less clunky to just handle <= 0 here.

swupd/create_manifests.go
```diff
@@ -52,12 +53,18 @@ func getOldManifest(path string) (*Manifest, error) {
 	return ParseManifestFile(path)
 }
 
-func initBundles(ui UpdateInfo, c config) ([]*Manifest, error) {
+func initBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error) {
 	var wg sync.WaitGroup
-	workers := len(ui.bundles)
-	wg.Add(workers)
+	if numWorkers <= 0 {
+		numWorkers = runtime.NumCPU()
```
rchiossi: I set the default to runtime.NumCPU() to be consistent with the default value used in other parts of the code.

tmarcu:
@rchiossi could you add a simple test for this that is expected to fail so we can make sure the code is exiting like it's supposed to now?

rchiossi:
So far I've been reverting #416 and creating a mix with the default packages + the chrome RPM. I'll need some time to come up with a test for it that can be merged and does't take 5 minutes to run.

rchiossi:
Also, I usually have to run it several times to test, since the order the manifests are created matter. If the problematic bundle is the last one, it skips the error path and doesn't test the changes, so I need to run until it falls in the middle.

Issue #113, opened 2018-01-26
/usr/lib/os-release of a mix should have information about the Mix itself


cmarcelo:
The /usr/lib/os-release is a file with various lines KEY=VALUE that has some well-known keys (like NAME, VERSION_ID). When creating a mix, only the VERSION_ID is updated, all other information remains like in Clear Linux, so we get a confusing output.

There are some useful things to do here:

- Fix other fields to make sure it is a mix
- Create a section in builder.conf that would allow the user to provide its own fields (e.g. NAME could be something more specific than "A mix of Clear Linux")
- Create a couple of new fields to reference the upstream version and upstream URL from where this mix was made

cmarcelo:
We are in a good position to implement this now. One idea I had was allow the user provide a os-release file and process it with the template syntax. If file isn't available, we just use a default template that has a generic NAME "Mix of clearlinux" and fills in the version id.

Also: would be nice to add a couple of fields with "UPSTREAM_VERSION=" and "UPSTREAM_URL=".

cmarcelo:
For debugging libarchive, some starting points:

<thiago> if anyone wnats to debug this, put a breakpoint in archive_entry.c line 1301 with condition r != 0
<thiago> the debug fuse doesn't work for me (container)

matthewrsj:
@cmarcelo wrong issue?

cmarcelo:
Yes! Sorry for the noise.

reaganlo:
@tmarcu @phmccarty 
Based on the Issue desc and 'man os-release', these are the changes that I recommend. Please let me know.

For a clear linux mix, the os-release file should be updated as follows:
- NAME=_"\<builder.conf ==>swupd ==> NAME>"_
E.g. **NAME="Clear Linux Mix"**
This value will be read from a new builder.conf [swupd] field called NAME and can be edited by the user. 
- ID=_\<based on above NAME field>_
E.g. **ID=clear-linux-mix**
- VERSION=_"\<mix-ver> (\<upstream-ver>)"_
E.g. **VERSION="10 (24750)"**
- PRETTY_NAME=_"\<NAME> \<mix-ver> (\<upstream-ver)"_
E.g. **PRETTY_NAME="Clear Linux Mix 10 (24750)"**
- VERSION_ID=_\<mix-ver>_
E.g. **VERSION_ID=10**
- BUILD_ID=_\<upstream-ver>_
E.g. **BUILD_ID=24750**
[...]

PR #420, opened 2018-09-01
#113 bundles:Update os-release file with mix info

reaganlo:
#113: For a clear linux mix, the os-release file will be updated as follows:
- NAME=_"Clear Linux Mix"_
  Eventually this value should be read from user input when #424 is implemented.
- ID=_\<based on above NAME field>_
E.g. **ID=clear-linux-mix**
- VERSION=_"\<mix-ver> (\<upstream-ver>)"_
E.g. **VERSION="10 (24750)"**
- PRETTY_NAME=_"\<NAME> \<mix-ver> (\<upstream-ver)"_
E.g. **PRETTY_NAME="Clear Linux Mix 10 (24750)"**
- VERSION_ID=_\<mix-ver>_
E.g. **VERSION_ID=10**
- BUILD_ID=_\<upstream-ver>_
E.g. **BUILD_ID=24750**
- MIX_UPSTREAM_URL=_"\<upstream-url>"_
This will be a new field in os-release.
E.g.  **MIX_UPSTREAM_URL="https://download.clearlinux.org"**

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -379,7 +379,7 @@ func initRPMDB(chrootDir string) error {
 	)
 }
 
-func buildOsCore(packagerCmd []string, chrootDir, version string) error {
+func buildOsCore(packagerCmd []string, chrootDir string, mixinfo mixInfo) error {
 	err := initRPMDB(chrootDir)
 	if err != nil {
 		return err
@@ -389,11 +389,11 @@ func buildOsCore(packagerCmd []string, chrootDir, version string) error {
 		return err
 	}
 
-	if err := createClearDir(chrootDir, version); err != nil {
+	if err := createClearDir(chrootDir, mixinfo.VersionID); err != nil {
 		return err
 	}
 
-	if err := fixOSRelease(filepath.Join(chrootDir, "usr/lib/os-release"), version); err != nil {
+	if err := fixOSRelease(filepath.Join(chrootDir, "usr/lib/os-release"), mixinfo); err != nil {
 		return errors.Wrap(err, "couldn't fix os-release file")
 	}
 
@@ -500,7 +500,7 @@ func rmDNFStatePaths(fullDir string) {
 	}
 }
 
-func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVersionDir, version string) error {
+func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVersionDir string, mixinfo mixInfo) error {
 	fmt.Println("Cleaning DNF cache before full install")
 	if err := clearDNFCache(packagerCmd); err != nil {
 		return err
@@ -515,7 +515,7 @@ func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVers
 		// special handling for os-core
 		if bundle.Name == "os-core" {
 			fmt.Println("... building special os-core content")
-			if err := buildOsCore(packagerCmd, fullDir, version); err != nil {
+			if err := buildOsCore(packagerCmd, fullDir, mixinfo); err != nil {
 				return err
 			}
 		}
@@ -552,6 +552,16 @@ func writeBundleInfo(bundle *bundle, path string) error {
 	}
 
 	return ioutil.WriteFile(path, b, 0644)
+}
+
+type mixInfo struct {
+	OSName      string
+	VersionID   string
+	BuildID     string
+	UpstreamURL string
+	ID          string
+	OSVersion   string
+	PrettyName  string
 }
 
 func (b *Builder) buildBundles(set bundleSet) error {
@@ -617,20 +627,20 @@ src=%s
 	}
 
 	// Mixer is used to create both Clear Linux or a mix of it.
-	var version string
+	var versionID string
 	if b.MixVer != "" {
 		fmt.Printf("Creating bundles for version %s based on Clear Linux %s\n", b.MixVer, b.UpstreamVer)
-		version = b.MixVer
+		versionID = b.MixVer
 	} else {
 		fmt.Printf("Creating bundles for version %s\n", b.UpstreamVer)
-		version = b.UpstreamVer
+		versionID = b.UpstreamVer
 		// TODO: This validation should happen when reading the configuration.
-		if version == "" {
+		if versionID == "" {
 			return errors.Errorf("no Mixver or Clearver set, unable to proceed")
 		}
 	}
 
-	buildVersionDir := filepath.Join(bundleDir, version)
+	buildVersionDir := filepath.Join(bundleDir, versionID)
 	fmt.Printf("Preparing new %s\n", buildVersionDir)
 	fmt.Printf("  and dnf config: %s\n", b.Config.Builder.DNFConf)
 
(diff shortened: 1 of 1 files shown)


builder/bundles.go
```diff
@@ -32,6 +32,7 @@ type buildBundlesConfig struct {
 	UpdateBundle string
 	ContentURL   string
 	VersionURL   string
+	OSName       string
```
rchiossi: It doesn't seems like you ever use this variable in the code.

rchiossi:
Oh, and notice that Travis will try to run `make lint`, `make check` and `make batcheck` on your pah and it will block your PR if those tests fail, so it is a good idea to test them locally before resubmitting. (also, it is much faster than waiting for Travis to test it for you! ;) )

builder/bundles.go
phmccarty: It's used in some new code in readBuildBundlesConfig()...

builder/bundles.go
```diff
@@ -925,15 +952,77 @@ func fixOSRelease(filename, version string) error {
...
 		if strings.HasPrefix(text, "VERSION_ID=") {
-			text = "VERSION_ID=" + version
+			text = "VERSION_ID=" + mixInfo[VersionID]
+			mixInfoFlags[VersionID] = true
+		}
+		if strings.HasPrefix(text, "BUILD_ID=") {
+			text = "BUILD_ID=" + mixInfo[BuildID]
+			mixInfoFlags[BuildID] = true
+		}
+		if strings.HasPrefix(text, "UPSTREAM_URL=") {
```
phmccarty: I don't see `UPSTREAM_URL` defined in the os-release documentation, so I would recommend following the guidelines in os-release(5) on naming distro-specific keys:

> Operating system vendors may extend the file format and introduce new fields. It is highly recommended to prefix new fields with an OS specific name in order to avoid name clashes. Applications reading this file must ignore unknown fields. Example: "DEBIAN_BTS="debbugs://bugs.debian.org/""

So, perhaps use `CL_UPSTREAM_URL` instead?

builder/bundles.go
rchiossi: It is read, but never used. I sent a PR that completely removes this part of the code since it is not needed anymore.

phmccarty: Right, I misunderstood your comment, sorry about that.

builder/bundles.go
```diff
@@ -617,6 +621,20 @@ func writeBundleInfo(bundle *bundle, path string) error {
 	return ioutil.WriteFile(path, b, 0644)
 }
 
+// Count of  mix info fields written to os-release
+const MixInfoCount = 7
```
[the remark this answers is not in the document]
reaganlo: Removed MixInfoCount as it is not required anymore.

builder/bundles.go
```diff
@@ -925,15 +952,77 @@ func fixOSRelease(filename, version string) error {
...
+	if mixInfoFlags[OSVersion] == false {
+		fmt.Fprintln(&newBuf, "VERSION=" + "\"" + mixInfo[OSVersion] + "\"")
+		mixInfoFlags[OSVersion] = true
+	}
+
+	if mixInfoFlags[PrettyName] == false {
+		fmt.Fprintln(&newBuf, "PRETTY_NAME=" + "\"" + mixInfo[PrettyName] + "\"")
+		mixInfoFlags[PrettyName] = true
+	}
+
```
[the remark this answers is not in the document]
reaganlo: Good point. Implemented.

builder/bundles.go
```diff
@@ -764,8 +782,17 @@ src=%s
...
+	mixInfo[VersionID] = versionID
+	mixInfo[BuildID] = b.UpstreamVer
+	mixInfo[UpstreamURL] = b.UpstreamURL
+	mixInfo[ID] = strings.Replace(strings.ToLower(mixInfo[OSName]), " ", "_", -1)
+	mixInfo[OSVersion] = mixInfo[VersionID] + " (" + mixInfo[BuildID] + ")"
+	mixInfo[PrettyName] = mixInfo[OSName] + " " + mixInfo[OSVersion]
+
 	// install all bundles in the set (including os-core) to the full chroot
-	err = buildFullChroot(cfg, b, &set, packagerCmd, buildVersionDir, version)
+	err = buildFullChroot(cfg, b, &set, packagerCmd, buildVersionDir, mixInfo[:])
```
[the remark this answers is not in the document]
reaganlo: Good point. Implemented.

builder/bundles.go
```diff
@@ -925,15 +952,77 @@ func fixOSRelease(filename, version string) error {
...
+			text = "ID=" + mixInfo[ID]
+			mixInfoFlags[ID] = true
+		}
+		if strings.HasPrefix(text, "VERSION=") {
+			text = "VERSION=" + "\"" + mixInfo[OSVersion] + "\""
+			mixInfoFlags[OSVersion] = true
+		}
+		if strings.HasPrefix(text, "PRETTY_NAME=") {
+			text = "PRETTY_NAME=" + "\"" + mixInfo[PrettyName] + "\""
+			mixInfoFlags[PrettyName] = true
```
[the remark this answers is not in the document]
reaganlo: Implemented a little differently as part of review comment changes.

builder/bundles.go
```diff
@@ -78,7 +78,6 @@ func readBuildBundlesConfig(path string) (*buildBundlesConfig, error) {
 	if err != nil {
 		return nil, err
 	}
-
```
rchiossi: Left over?

builder/bundles.go
```diff
@@ -917,21 +935,32 @@ func fixOSRelease(filename, version string) error {
...
-			text = "VERSION_ID=" + version
+		// Skipping the fields that are already written to the buffer
+		if strings.HasPrefix(text, "NAME=") ||
+			strings.HasPrefix(text, "VERSION_ID=") ||
+			strings.HasPrefix(text, "BUILD_ID=") ||
+			strings.HasPrefix(text, "UPSTREAM_URL=") ||
+			strings.HasPrefix(text, "ID=") ||
+			strings.HasPrefix(text, "VERSION=") ||
+			strings.HasPrefix(text, "PRETTY_NAME=") {
+			continue
```
rchiossi: you don't need the continue since you moved the line bellow into an else block.

builder/bundles.go
```diff
@@ -764,8 +773,17 @@ src=%s
 		}
 	}
 
+	var mixinfo mixInfo
+	mixinfo.OSName = "Clear Linux Mix" // TODO: Read user defined value when #424 is implemented.
```
tmarcu: :+1: :fireworks:

tmarcu:
```
root@clr-57a2f76cbfb2499797db96cef2624041 ~ # cat /usr/lib/os-release 
NAME="Clear Linux Mix"
VERSION_ID=10
BUILD_ID=24870
UPSTREAM_URL="https://download.clearlinux.org"
ID=clear_linux_mix
VERSION="10 (24870)"
PRETTY_NAME="Clear Linux Mix 10 (24870)"
ANSI_COLOR="1;35"
HOME_URL="https://clearlinux.org"
SUPPORT_URL="https://clearlinux.org"
BUG_REPORT_URL="mailto:dev@lists.clearlinux.org"
PRIVACY_POLICY_URL="http://www.intel.com/privacy"
```

Things are looking like they have the right info! I'm wondering if we want to keep the **24870** in the line` PRETTY_NAME="Clear Linux Mix 10 (24870)"` though, but not a blocker in any case.

builder/bundles.go
phmccarty: I think my point still stands. IMO, we should follow the guidelines in os-release(5) and use a distro-specific prefix to avoid future naming clashes; any generic name, like `UPSTREAM_URL`, has the potential of being adopted as standard in the future.

The arguments against using the `CL_` prefix are valid, so you can just choose a different prefix. Loosening the definition of "distro" to "tool that creates the distro", perhaps use a `MIX_` prefix?

PR #421, opened 2018-09-03
bundles: Remove redundant parsing of config file.

rchiossi:
All the values inside buildBundlesConfig are already parsed and
available in the code through builder.Config, so there is no need to
parse them again from the config file.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -16,71 +16,8 @@ import (
 	"time"
 
 	"github.com/clearlinux/mixer-tools/helpers"
-	"github.com/go-ini/ini"
 	"github.com/pkg/errors"
 )
-
-// TODO: Move this to the more general configuration handling.
-type buildBundlesConfig struct {
-	// [Server] section.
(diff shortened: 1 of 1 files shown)


rchiossi:
I saw some funny things going on in bundles.go while reviewing #420 so I decided to fix them. Hopefully it won't conflict too much with that PR. ;)

builder/bundles.go
```diff
@@ -662,15 +592,15 @@ func (b *Builder) buildBundles(set bundleSet) error {
 emptydir=%s/empty
 imagebase=%s/image/
 outputdir=%s/www/
-`, b.Config.Builder.ServerStateDir, b.Config.Builder.ServerStateDir, b.Config.Builder.ServerStateDir)
-	if cfg.HasServerSection {
```
tmarcu: is this ok now because there will always be a section with a default value?

rchiossi: Yes, the server section is always present now so this is always true.

PR #425, opened 2018-09-08
Do not use container for build image

tmarcu:
The build image command cannot run in a container because there are no
loopback devices to mount in a container environment. Force running with
native if build image is called.


tmarcu:
I don’t disagree with the state it’s in, but that’s a different issue to be addressed. Right now mixer users can’t run build image without knowing to pass —native.

Issue #58, opened 2017-12-23
Proposal: Automate Ister config file


kevincwells:
# Proposal
Change mixer to automate the acquisition/editing of the `ister` config file (currently `release-image-config.json`).

# Notes
- At a minimum, automate the acquisition of this file so the user doesn't have to `curl` it. Ship a local copy? Curl it ourselves in `mixer init`?
- Automate the setting of the "version" and "format" variables using the values from `builder.conf`?
  - "Version" is in the `config.json` file, while "format" is passed to `ister` as a commandline flag.
- Introduce a "Image Bundles List" in `builder.conf` (or elsewhere), then generate the `config.json` file's "Bundles" list automatically?

cmarcelo:
> At a minimum, automate the acquisition of this file so the user doesn't have to curl it. Ship a local copy? Curl it ourselves in mixer init?

No download. We can embed a default in our code for purposes of "mix init". The format of this configuration for ister should be stable by now.

> Introduce a "Image Bundles List" in builder.conf (or elsewhere), then generate the config.json file's "Bundles" list automatically?

I thought about this before too, but I've changed my mind and don't think we should do this.

The partitioning, mounting table are also important bits for the user to tweak, e.g. "the default 10G images are too big for me" or "I need an extra partition for data by default". The config used by ister has exactly this level of information. Note: I'm envisioning I would add these configuration files for the images I want to keep into version control for the mix I'm creating.

> Automate the setting of the "version" and "format" variables using the values from builder.conf?
>    "Version" is in the config.json file, while "format" is passed to ister as a commandline flag.

I feel neither Version or Format should be in this configuration file for image.

kevincwells:
@cmarcelo 
> No download. We can embed a default in our code for purposes of "mix init". The format of this configuration for ister should be stable by now.

Agreed - I don't think we should be curling it from the network. Easiest right now would be to just ship a local copy with mixer's install, and copy it in to the mix with the init (just like how `builder.conf` gets created right now). It could also be generated programmatically, especially if we did want to have ister config parsing/editing logic in mixer.

> I thought about this before too, but I've changed my mind and don't think we should do this.

Yeah, I've changed my mind on this topic too, I think. Especially after talking to @mdhorn about how people would likely want to use this in practice, I don't think mixer can auto-generate ister config in any meaningfully useful way.

> I feel neither Version or Format should be in this configuration file for image.

[...]

PR #427, opened 2018-09-14
bundles: Build image with default ister template file (#58)

reaganlo:
When building an image without providing a template, if the default ister template
"release-image-config.json" is not present in the mix directory,
mixer will do the following:
 - Copy "/usr/share/defaults/ister/release-image-config.json" to the mix directory.
 - Update "release-image-config.json" file with the bundle list from "mixbundles" file.
 - Inform the user about the above two steps.
 - Continue the existing build process.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -1729,6 +1730,9 @@ func (b *Builder) buildUpdateContent(params UpdateParameters, timer *stopWatch) 
 	return nil
 }
 
+const imageTemplate = "release-image-config.json"
+const isterConfigDir = "/usr/share/defaults/ister"
+
 // BuildImage will now proceed to build the full image with the previously
 // validated configuration.
 func (b *Builder) BuildImage(format string, template string) error {
@@ -1737,9 +1741,23 @@ func (b *Builder) BuildImage(format string, template string) error {
 		format = b.State.Mix.Format
 	}
 
-	// If the user did not pass in a template, default to release-image-config.json
+	// If the user did not pass in a template, use the default template
 	if template == "" {
-		template = "release-image-config.json"
+		template = imageTemplate
+
+		// If the default image template is not present in the mix workspace,
+		// copy it from the default ister config directory and update the bundle list based on mix bundles
+		templateFile := filepath.Join(b.Config.Builder.VersionPath, template)
+		if _, err := os.Stat(templateFile); os.IsNotExist(err) {
+			fmt.Printf("Warning: Image template %s not found\n", templateFile)
+			configFile := filepath.Join(isterConfigDir, template)
+			fmt.Printf("Copying image template from %s\n", configFile)
+			if err = b.copyImageTemplate(configFile, templateFile); err != nil {
+				return err
+			}
+		} else if err != nil {
+			return err
+		}
 	}
 
 	// swupd (client) called by itser will need a temporary directory to act as its stage dir.
@@ -1758,6 +1776,48 @@ func (b *Builder) BuildImage(format string, template string) error {
 	imagecmd.Stderr = os.Stderr
 
 	return imagecmd.Run()
+}
+
+// copyImageTemplate will copy the image template from the default ister config directory
+// and update the image bundle list based on mix bundles.
+// If there is an error updating the image bundle list, the default bundle list will be used.
+func (b *Builder) copyImageTemplate(configFile, templateFile string) error {
+	// Read ister template
+	configValues, err := ioutil.ReadFile(configFile)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Updating image bundle list based on %s\n", b.MixBundlesFile)
+	mixBundles, err := b.getMixBundlesListAsSet() // returns empty set with no error if mix bundles file is not present
+	if err == nil && len(mixBundles) > 0 {        // check if set is not empty
+		var bundles []string
+		for _, bundle := range mixBundles {
+			bundles = append(bundles, bundle.Name)
+		}
+		if err = updateImageBundles(&configValues, bundles); err != nil {
+			return errors.Wrap(err, "Failed to copy image template. Invalid JSON format")
+		}
+	} else {
+		fmt.Printf("Warning: Failed to read %s. Using default bundle list instead\n", b.MixBundlesFile)
+	}
+
+	return ioutil.WriteFile(templateFile, configValues, 0644)
+}
+
+// updateImageBundles will update the image bundle list based on the mix bundles.
+func updateImageBundles(configValues *[]byte, bundles []string) error {
+	var data map[string]interface{}
+	err := json.Unmarshal(*configValues, &data)
+	if err != nil {
+		return err
+	}
+
+	data["Bundles"] = bundles
+
+	*configValues, err = json.MarshalIndent(data, "", " ")
+
+	return err
 }
 
 type dnfRepoConf struct {


builder/builder.go
```diff
@@ -1737,13 +1738,37 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+
 	// If the user did not pass in a template, default to release-image-config.json
 	if template == "" {
 		template = "release-image-config.json"
+
+		// If  the default image template is not present in the mix workspace,
+		// copy it from the default ister config directory.
+		templateFile := filepath.Join(wd, template)
+		if _, err := os.Stat(templateFile); os.IsNotExist(err) {
+			var isterConfigDir = "/usr/share/defaults/ister"
```
matthewrsj: This is a hard-coded variable. List it as a const at global scope. That way when (if) it changes for Clear Linux it is easier to find and changing it in one place changes it for the entire codebase. Really this is a configuration thing but let's not add config options right now.

builder/builder.go
```diff
@@ -1737,13 +1738,37 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+			fmt.Printf("Default image template %s not found in mix workspace\n", template)
+			fmt.Printf("Copying default image template from %s\n", isterConfigDir)
+			if err := helpers.CopyFile(templateFile, filepath.Join(isterConfigDir, template)); err != nil {
+				return err
+			}
+
+			// Update the default image template based on the mix bundles.
+			bundles, err := helpers.ReadFileAndSplit(filepath.Join(wd, b.MixBundlesFile))
+			if err != nil || len(bundles) <= 1 {
+				// In case of error or empty file, use default image template as is.
```
matthewrsj: The empty `if` block is really weird
```go
if err == nil && len(bundles) > 1 {
    fmt.Printf("Updating image template based on %s\n", b.MixBundlesFile)
    if err := updateImageTemplate(templateFile, bundles[:len(bundles)-1]); err != nil {
        return err
    }
}
```
I'm guessing you guard against 1 because of os-core? Add a comment about why you are using 1 here.

builder/builder.go
```diff
@@ -1737,13 +1738,37 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+			} else {
+				fmt.Printf("Updating image template based on %s\n", b.MixBundlesFile)
+				if err := updateImageTemplate(templateFile, bundles[:len(bundles)-1]); err != nil {
```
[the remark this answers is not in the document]
matthewrsj: default os-core? needs a comment for sure.

builder/builder.go
```diff
@@ -1760,6 +1785,29 @@ func (b *Builder) BuildImage(format string, template string) error {
 	return imagecmd.Run()
 }
 
+// Updates bundle list in the default image template file
+func updateImageTemplate(filePath string, bundles []string) error {
+	f, err := os.Open(filePath)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
```
matthewrsj: I think this will fail `make lint`

builder/builder.go
```diff
@@ -1760,6 +1785,29 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+
+	byteValue, _ := ioutil.ReadAll(f)
```
matthewrsj: stop throwing away the errors in critical code paths.

builder/builder.go
```diff
@@ -1760,6 +1785,29 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+	var data map[string]interface{}
+	if err := json.Unmarshal([]byte(byteValue), &data); err != nil {
+		return err
+	}
+
+	data["Bundles"] = bundles
+	b, err := json.MarshalIndent(data, " ", " ")
```
matthewrsj: Why do you supply a `prefix`? That looks weird in the result:
```go
package main

import (
	"encoding/json"
	"fmt"
	"log"
)

func main() {
	data := map[string]int{
		"a": 1,
		"b": 2,
	}

	json, err := json.MarshalIndent(data, "<prefix>", "<indent>")
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println(string(json))
}
```
produces
```
{
<prefix><indent>"a": 1,
<prefix><indent>"b": 2
<prefix>}
```

reaganlo:
@matthewrsj @rchiossi FYI. Not ready for review yet. I'm working on the suggested changes. 'make lint' on my system shows errors for code that has not been modified. So pushed the code just to verify if it passes Travis CI.

builder/builder.go
```diff
@@ -1737,13 +1741,61 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+				return err2
+			}
+
+			// Read mix bundles
+			bundlesFile := filepath.Join(wd, b.MixBundlesFile)
+			bundles, err2 := helpers.ReadFileAndSplit(bundlesFile) // Note: Returns  trailing empty element if file is empty.
+			// Update bundle list  only if no error and mix bundles not empty
+			if err2 == nil && len(bundles) > 1 {
+				fmt.Printf("Updating image template based on %s\n", bundlesFile)
+				data["Bundles"] = bundles[:len(bundles)-1] // Avoid trailing empty element
```
rchiossi: This is a very dangerous way of getting rid of the trailing empty element. If for any reason the mixbundle file does not contain the trailing 0x0a, you will actually remove a valid bundle.  This also doesn't handle empty lines in the middle of the list. I think it is safer to iterate `bundles` and add one element at a time, skipping if the element is empty.

reaganlo: @rchiossi Thanks for the comments. But I had mentioned its not ready for review yet. I was in the middle of refactoring the code.

builder/builder.go
```diff
@@ -1737,13 +1741,61 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+
+	// If the user did not pass in a template, use the default
 	if template == "" {
-		template = "release-image-config.json"
+		template = imageTemplate
+
+		// If  the default image template is not present in the mix workspace,
+		// copy it from the default ister config directory and update the bundle list based on mix bundles
+		templateFile := filepath.Join(wd, template)
+		if _, err1 := os.Stat(templateFile); os.IsNotExist(err1) {
```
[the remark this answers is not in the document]
reaganlo: @rchiossi I know. But make lint was throwing a warning saying it was shadowing a previous error variable.

builder/builder.go
```diff
@@ -1737,13 +1741,61 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+			fmt.Printf("Image template %s not found\n", templateFile)
+			var err2 error
```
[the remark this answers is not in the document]
reaganlo: @rchiossi same as my above response

builder/builder.go
rchiossi: ah sorry, I saw it passed travis checks I assumed it was ready for review

builder/builder.go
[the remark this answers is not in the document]
rchiossi: It is probably because you were redifining with `:=` instead of assigning with `=`

builder/builder.go
reaganlo: Done.

builder/builder.go
reaganlo: This was to skip the empty trailing element returned by ReadFileAndSplit(). Modified code to read list using "for loop" instead.

builder/builder.go
```diff
@@ -1737,13 +1738,37 @@ func (b *Builder) BuildImage(format string, template string) error {
 		format = b.State.Mix.Format
 	}
 
+	// swupd (client) called by itser will need a temporary directory to act as its stage dir.
+	wd, _ := os.Getwd()
```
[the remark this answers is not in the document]
reaganlo: Done. This was displaced existing code.

builder/builder.go
reaganlo: Fixed.

builder/builder.go
```diff
@@ -1737,13 +1738,37 @@ func (b *Builder) BuildImage(format string, template string) error {
...
 		template = "release-image-config.json"
+
+		// If  the default image template is not present in the mix workspace,
+		// copy it from the default ister config directory.
+		templateFile := filepath.Join(wd, template)
+		if _, err := os.Stat(templateFile); os.IsNotExist(err) {
+			var isterConfigDir = "/usr/share/defaults/ister"
+			fmt.Printf("Default image template %s not found in mix workspace\n", template)
+			fmt.Printf("Copying default image template from %s\n", isterConfigDir)
+			if err := helpers.CopyFile(templateFile, filepath.Join(isterConfigDir, template)); err != nil {
```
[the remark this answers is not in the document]
reaganlo: Done.

builder/builder.go
```diff
@@ -1737,13 +1738,37 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+	wd, _ := os.Getwd()
+
 	// If the user did not pass in a template, default to release-image-config.json
 	if template == "" {
 		template = "release-image-config.json"
+
+		// If  the default image template is not present in the mix workspace,
+		// copy it from the default ister config directory.
+		templateFile := filepath.Join(wd, template)
+		if _, err := os.Stat(templateFile); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
reaganlo: Done. 
Stat() on template file. ReadFile() on isterConfig file. Then pass the value read by reference.

builder/builder.go
reaganlo: Fixed.

builder/builder.go
reaganlo: Fixed. Empty prefix.

builder/builder.go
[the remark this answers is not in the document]
reaganlo: Fixed.

builder/builder.go
```diff
@@ -1737,13 +1741,61 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+			b, err2 := json.MarshalIndent(data, "", " ")
+			if err2 != nil {
+				return err2
+			}
+			err2 = ioutil.WriteFile(templateFile, b, 0644)
+			if err2 != nil {
+				return err2
+			}
+		} else {
+			if err1 != nil {
```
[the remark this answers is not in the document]
reaganlo: Fixed.

builder/builder.go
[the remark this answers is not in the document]
reaganlo: Fixed.

reaganlo:
@rchiossi Ready for review.

builder/builder.go
```diff
@@ -1760,6 +1798,38 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+		return err
+	}
+
+	var bundles []string
+	for _, bundle := range mixBundles {
+		bundle = strings.TrimSpace(bundle)
+		if bundle != "" {
+			bundles = append(bundles, bundle)
+		}
+	}
```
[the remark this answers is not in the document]
reaganlo: I considered using this method. But it gets additional bundle info and not just the name. So to avoid unnecessary additional data I decided not to use it.

builder/builder.go
```diff
@@ -1737,13 +1741,47 @@ func (b *Builder) BuildImage(format string, template string) error {
 		format = b.State.Mix.Format
 	}
 
-	// If the user did not pass in a template, default to release-image-config.json
+	// swupd (client) called by itser will need a temporary directory to act as its stage dir.
```
matthewrsj: `s/itser/ister/`

builder/builder.go
```diff
@@ -1737,13 +1741,47 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+	wd, err := os.Getwd()
```
matthewrsj: I don't like this approach. Don't be looking in the current working directory, look in the mixer working directory as defined in the conf. Users can call mixer from anywhere and define a different mixer working directory. Case in point is the Clear Linux devops process.

builder/builder.go
```diff
@@ -1737,13 +1741,47 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+	// If the user did not pass in a template, use the default
 	if template == "" {
-		template = "release-image-config.json"
+		template = imageTemplate
+
+		// If  the default image template is not present in the mix workspace,
+		// copy it from the default ister config directory and update the bundle list based on mix bundles
+		templateFile := filepath.Join(wd, template)
+		if _, err = os.Stat(templateFile); os.IsNotExist(err) {
+			fmt.Printf("Warning: Image template %s not found\n", templateFile)
```
matthewrsj: Please factor this into another function with a useful name like `copyTemplateFile(src, dest)` instead of all the code in this third indentation level.

builder/builder.go
```diff
@@ -1737,13 +1741,47 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+			fmt.Printf("Copying image template from %s\n", configFile)
+
+			// Read ister template
+			var configValues []byte
+			configValues, err = ioutil.ReadFile(configFile)
+			if err != nil {
+				return err
+			}
+
+			mixBundlesFile := filepath.Join(wd, b.MixBundlesFile)
```
matthewrsj: This will absolutely fail if running mixer from outside the mixer working directory.

builder/builder.go
```diff
@@ -1737,13 +1741,47 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+			fmt.Printf("Updating image bundle list based on %s\n", mixBundlesFile)
+			if err = UpdateImageBundles(&configValues, mixBundlesFile); err != nil {
+				fmt.Println("Warning: Could not update image bundle list. Using default bundle list instead.")
```
matthewrsj: You aren't being consistent about your printfs. Either put a period at the end of the sentence or don't, please do not mix styles.

builder/builder.go
```diff
@@ -1737,13 +1741,47 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+			if err != nil {
+				return err
+			}
+		} else if err != nil {
+			return err
+		}
 	}
 
-	// swupd (client) called by itser will need a temporary directory to act as its stage dir.
-	wd, _ := os.Getwd()
```
matthewrsj: Oh yikes, I see the current directory was being used all along.

Regardless, since you are modifying this code let's fix it now.

builder/builder.go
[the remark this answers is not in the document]
matthewrsj: I'd rather used the well-defined way to get the bundle list. If you need something more stripped-down define it as a function on on `builder` like `getMixBundlesListAsSet` is.

I really don't want to add manual parsing of a file in random parts of the codebase. It is very unmaintainable if we end up changing the file format in the future.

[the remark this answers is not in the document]
reaganlo: Done. Used existing b.getMixBundlesListAsSet()

builder/builder.go
reaganlo: Done.

builder/builder.go
reaganlo: Done. Using b.Config.Builder.VersionPath to obtain mixer working directory

builder/builder.go
reaganlo: Fixed. This printf had multiple sentences. Thats why used a period. But anyhow removed the period at the end of the sentence.

builder/builder.go
reaganlo: As discussed, not fixing this existing bug in this PR. Created another Issue to track this.

builder/builder.go
```diff
@@ -1760,6 +1778,55 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+			bundles = append(bundles, bundle.Name)
+		}
+		if err = updateImageBundles(&configValues, bundles); err != nil {
+			return errors.Wrap(err, "Failed to copy image template. Invalid JSON format")
+		}
+	} else {
+		fmt.Printf("Warning: Failed to read %s. Using default bundle list instead\n", b.MixBundlesFile)
+	}
+
+	if err = ioutil.WriteFile(templateFile, configValues, 0644); err != nil {
```
rchiossi: `return ioutil.WriteFile(templateFile, configValues, 0644)`

builder/builder.go
```diff
@@ -1760,6 +1778,55 @@ func (b *Builder) BuildImage(format string, template string) error {
...
+	var data map[string]interface{}
+	err := json.Unmarshal(*configValues, &data)
+	if err != nil {
+		return err
+	}
+
+	data["Bundles"] = bundles
+
+	*configValues, err = json.MarshalIndent(data, "", " ")
+	if err != nil {
```
rchiossi: `return err`

builder/builder.go
reaganlo: done.

builder/builder.go
reaganlo: done

tmarcu:
@reaganlo bump?

matthewrsj:
needs a rebase xD that builder cleanup is making a mess.

reaganlo:
Rebasing was getting complicated. So created a new PR #492 with the same code changes and added bat tests. Closing this one.

Issue #428, opened 2018-09-18
mixer init --upstream-url error


qzheng527:
I got below errors when setting the upstream-url for mixer init.

$mixer init --upstream-url=http://linux-ftp.sh.intel.com/pub/mirrors/clearlinux/
ERROR: Failed to retrieve latest published upstream version. Missing proxy configuration? : got status "404 Not Found" when downloading: http://linux-ftp.sh.intel.com/latest

Looks like the mixer tool ignored the strings after the first "/" for upstream-url.

rchiossi:
Fixed by #430

PR #430, opened 2018-09-18
builder: Fix URL build for base URL with subpath

rchiossi:
When building the upstream URL, if the `b.UpstreamURL` was set do a url
with a subpath included, the subpath was being discarded by
`URL.ResolveReference()`. In order to preserve the subpath in the base,
the provided subpath must be manually added to the base URL object.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -27,6 +27,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


rchiossi:
This patch fixes issue #428

otaviobp:
+1

PR #431, opened 2018-09-18
Revert "helpers: Reduce download timeout"

rchiossi:
This reverts commit 9cf854d8d6e45dcb087c0bf4f61bfc633c16e43b.

The reduction to the timeout has caused failures for users on slow
networks, so this must be reverted to the default values.

The reduction to the timeout was added to improve user experience when
connection failed due to a proxy error. It would take mixer too long to
fail and the error message didn't indicate what the problem could be.

The error is now properly indicated, so even if it takes a while for the
error to occur the first time, the user should be able to identify the
problem on the spot.

### DIFF
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -365,10 +365,7 @@ func ListVisibleFiles(dirname string) ([]string, error) {
 }
 
 func getDownloadFileReader(url string) (*io.ReadCloser, error) {
-	httpClient := http.Client{
-		Timeout: time.Duration(10 * time.Second),
-	}
-	resp, err := httpClient.Get(url)
+	resp, err := http.Get(url)
 	if err != nil {
 		return nil, err
 	}


cmarcelo:
What about the suggestion in https://github.com/clearlinux/mixer-tools/commit/9cf854d8d6e45dcb087c0bf4f61bfc633c16e43b#commitcomment-30484286 to set the dialer timeout?

rchiossi:
> What about the suggestion in [9cf854d#commitcomment-30484286](https://github.com/clearlinux/mixer-tools/commit/9cf854d8d6e45dcb087c0bf4f61bfc633c16e43b#commitcomment-30484286) to set the dialer timeout?

Honestly, I don't think the trade-off is not worth it. I'd rather have users who forgot to configure their proxy wait for 20 seconds than have users with correct configuration be unable to create a mix because we set the timeout to a low value. If this continues to be an issue even with the new error messages we can revisit this.

matthewrsj:
> I'd rather have users who forgot to configure their proxy wait for 20 seconds than have users with correct configuration be unable to create a mix because we set the timeout to a low value.

+1

cmarcelo:
> If this continues to be an issue even with the new error messages we can revisit this.

Fair enough.

ahkok:
A dialer timeout (initial handshake) is not a bad idea, but you'd really have to implement a few retries if you do, due to the way that DNS can take forever :)

Issue #429, opened 2018-09-18
swupd-server to mixer (yocto integration)


aaronovz1:
@matthewrsj / @kevincwells creating a ticket here to follow on from conversations in https://github.com/clearlinux/swupd-server/issues/111.

Some slight progress but running into some issues and segfaults. These are the steps I've taken so far:

1. Start the docker image: `sudo docker run -v /media/data/yp2.5/build/intel/tmp/deploy/swupd/intel-64/intel-image:/home/clr/mix -it clearlinux/clr-sdk`
2. `sudo mixer init --native`
3. Modify `mixversion` file to have value of 1 instead of 10. Otherwise I get `ERROR: Couldn't build update: failed to create update metadata: no full manifest found` because my OS version starts at 1 not the default 10. I can probably fix this by using `--mix-version=1` in mixer init?
4. `sudo mixer build update --new-config=false --native` this appears to try to do something but then errors with a segfault

```
clr@13a48903e139 ~/mix $ sudo mixer build update --new-config=false --native
=> CREATE MANIFESTS
Generating initial manifests...
  os-core
  connectivity
  security
  dss
  kernelmodules
  debugging
  security-tools
Reading bundle includes...
Performing manifest file subtraction...
Detecting manifest changes...
Writing manifest files...

TIMINGS
  CREATE MANIFESTS 0s
TOTAL: 0s
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x68 pc=0x7265f9]

goroutine 1 [running]:
github.com/clearlinux/mixer-tools/swupd.(*Manifest).subtractManifestFromManifest(0xc4204e2380, 0x0)
	/builddir/build/BUILD/mixer-tools-5.1.3/.gopath/src/github.com/clearlinux/mixer-tools/swupd/manifest.go:592 +0x59
github.com/clearlinux/mixer-tools/swupd.(*Manifest).subtractManifests(0xc4204e2380, 0x0)
	/builddir/build/BUILD/mixer-tools-5.1.3/.gopath/src/github.com/clearlinux/mixer-tools/swupd/manifest.go:631 +0x92
github.com/clearlinux/mixer-tools/swupd.writeIndexManifest(0xc42014b1c8, 0xc42014b178, 0xc420133a78, 0x1, 0x1, 0xc4203ea240, 0xd, 0xc4203c1fa0)
[...]

PR #433, opened 2018-09-19
cmd: Respect --offline flag for network check

rchiossi:
There are a couple mixer commands that don't require network and thus
skip any network check. This behavior is also intended when the
--offline flag is used, but the flag is never checked in that scenario.

This patch unifies builder.Offline (which is set by --offline) and
networkCheck variable to make sure the network check is also skipped if
--offline flag is used.


rchiossi:
This PR addresses one of the issues identified in #429

otaviobp:
lgtm

PR #438, opened 2018-09-27
swupd: Don't add deleted bundles to MoM

rchiossi:
When a bundle gets deleted, its entry still exists in the previous MoM
but it shouldn't exist in the new MoM. Since the deleted bundle will not
exist in the bundle list anymore, it must be also verified to determined
if an missing entry should be propagated to the new MoM.

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -267,14 +267,22 @@ func processBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error
 	return newManifests, nil
 }
 
-func addUnchangedManifests(appendTo *Manifest, appendFrom *Manifest) {
+func addUnchangedManifests(appendTo *Manifest, appendFrom *Manifest, bundles []string) {
 	for _, f := range appendFrom.Files {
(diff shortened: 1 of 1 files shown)


swupd/create_manifests.go
```diff
@@ -267,14 +267,22 @@ func processBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error
...
+		if f.findFileNameInSlice(appendTo.Files) != nil {
+			continue
+		}
+
+		if f.Name == IndexBundle {
+			// this is generated new each time
+			continue
+		}
+
+		for _, bundle := range bundles {
```
matthewrsj: This for loop should just be 
```go
if f.findFileNameInSlice(bundle) != nil {
    appendTo.Files = append(appendTo.Files, f)
}
```
right now that function isn't optimized but it could be at a later point and this code would get the benefit.

matthewrsj: oh wait, that function only works for `*File` slices, not `string` slices. Never mind...

tmarcu:
Approved once we see the testing reflect the right output.

PR #446, opened 2018-10-02
builder: Use same commenting rules for package file and bundles

cmarcelo:
Make package file parsing ignore '#' and anything after that until the
end of the line.  Also trim spaces.

This makes the comment syntax the same as bundle files.


cmarcelo:
I'm not sure if we have other consumers of this file that will get disrupted.

@bryteise @tmarcu @rchiossi

cmarcelo:
Thanks for the suggestion, @tmarcu. Let me know if you want any other particular scenario added to the tests list.

Since swupd-extract and swupd-inspector are currently in mixer-tools repository, feel free to CC me on policy changes on the repo usage :-)

cmarcelo:
Pushed a new version with the tests added.

PR #449, opened 2018-10-02
AFB: Add afb.sh script to outline format bump steps

matthewrsj:
This script is intended as a reference and architecture document for a
new format bump flow. This format bump flow is much simpler than the
original/current flow which requires jumping back and forth between the
+20 and +10 builds to create the updates.

### DIFF
--- a/afb.sh
+++ b/afb.sh
@@ -1,6 +1,28 @@
 #!/bin/bash
 
 ###############################################################################
+# FORMAT BUMPS
+#
+# A format defines a range of OS versions that have compatible update metadata
+# and content. An update client can update a system from the oldest version in
+# the format to the latest version in the format without worrying about
+# compatibility issues in the update content for the version it is updating to.
+#
+# A format bump occurs when the update metadata or content is changed in such a
+# way that will cause client updates to break. In this case the format number
+# must be incremented so clients will not attempt to update to the new versions
+# in the new format without crossing the format boundary. Update clients update
+# only to the latest build in their format. Once that update is complete the
+# update client may then update forward again because the last version in the
+# current format has identical content to the first version in the new format,
+# including the new update client needed to understand the new format.
+#
+# Clients must be able to update to the latest version in their format using
+# the update metadata it understands. Once it is on that version it should also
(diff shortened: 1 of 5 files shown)


matthewrsj:
much of this wording duplicated in this PR: https://github.com/clearlinux/clr-man-pages/pull/4

matthewrsj:
I've updated this to give more information about format bumps. As discussed we should merge this so we can track history of this file.

afb.sh
```diff
@@ -1,5 +1,27 @@
...
+# A format bump occurs when the update metadata or content is changed in such a
+# way that will cause client updates to break. In this case the format number
+# must be incremented so clients will not attempt to update to the new versions
+# in the new format without crossing the format boundary. Update clients update
+# only to the latest build in their format. Once that update is complete the
+# update client may then update forward again because the last version in the
+# current format has identical content to the first version in the new format,
+# including the new update client needed to understand the new format.
+#
+# So that clients can update to latest in a format using their format's
```
tmarcu: This is a very hard sentence to read (for me), could you re-phrase it? Maybe it got mashed together during a revision.

matthewrsj:
@tmarcu clarified comment and added a bat test with this procedure.

matthewrsj:
> Shouldn't the bat test be under "06-format-bump-test" as well. I know that right now it is testing downstream format bump, but I believe it would be a bit more organized to have everything format bump related in the same test directory.

Good point, will do.

> I'd also squash the fist 3 commits.

I actually want a really complete history of revisions on this document so I opted not to squash. If @tmarcu also wants a squash though I'll bow to the majority.

tmarcu:
No squash pls.

matthewrsj:
> > Shouldn't the bat test be under "06-format-bump-test" as well. I know that right now it is testing downstream format bump, but I believe it would be a bit more organized to have everything format bump related in the same test directory.
>
> Good point, will do.

Never mind, I'm running into a lot of issues with the shared directory and having to clean up and re-init mixer to do what I want. Looking through the other tests there isn't a clean way to solve this (with the current mixerlib.bash) so I'd rather just leave it the way it is.

PR #450, opened 2018-10-03
swupd: add minversion header to MoM

matthewrsj:
An upcoming improvement to swupd-client requires the minversion header
to exist in the MoM in order to determine the strategy with which to
update files in the update.

When the minVersion argument is provided to createManifests populate a
header in the MoM with this version. If no flag is provided and the
previous MoM had a minversion set populate the new MoM with this
version.

Write the minversion to the manifest header when the minversion != 0.
Since we only set minversion for the MoM the minversion will only be
written to the MoM and not to bundle manifests.


matthewrsj:
@jwakre

swupd/manifest_test.go
```diff
@@ -210,10 +210,10 @@ func TestCheckInvalidManifestHeaders(t *testing.T) {
...
 		header ManifestHeader
 	}{
-		{"format not set", ManifestHeader{0, 100, 90, 553, time.Unix(1000, 0), 100000, nil}},
-		{"version zero", ManifestHeader{10, 0, 90, 553, time.Unix(1000, 0), 100000, nil}},
-		{"no files", ManifestHeader{10, 100, 90, 0, time.Unix(1000, 0), 100000, nil}},
-		{"no timestamp", ManifestHeader{10, 100, 90, 553, zeroTime, 100000, nil}},
+		{"format not set", ManifestHeader{0, 100, 90, 0, 553, time.Unix(1000, 0), 100000, nil}},
+		{"version zero", ManifestHeader{10, 0, 90, 0, 553, time.Unix(1000, 0), 100000, nil}},
+		{"no files", ManifestHeader{10, 100, 90, 0, 0, time.Unix(1000, 0), 100000, nil}},
+		{"no timestamp", ManifestHeader{10, 100, 0, 90, 553, zeroTime, 100000, nil}},
```
phmccarty: These changes should appear in the commit after this one, because you don't change the `ManifestHeader` struct until then.

matthewrsj: I ordered the commits to the tests came first... Instead of dealing with this I just reordered the commits so the tests come after the implementation. `¯\_(ツ)_/¯`

phmccarty: I'm okay with having test commits come first, but (ideally) mixer-tools should build successfully on every commit. That's what my comment was aimed at.

matthewrsj: Yeah I agree, I'm just saying the way I resolved your comment is to just reorder the commits instead of picking that bit of history into another commit.

Also, mixer *does* build even with the change that way. Go will not build the test files unless it runs the tests.

phmccarty: Oh, right, sorry about not noticing that.

PR #456, opened 2018-10-09
swupd: do not update delete versions over minversions

### LINKED ISSUE #455: Minversion should not modify versions for deleted files
In Clear Linux OS, there was a (full) minversion bump for build 25440. After I updated to this version, I saw my content in /usr/local/bin wiped, because a while back /usr/local/bin was introduced to the distro by accident, and mixer bumped the version for /usr/local/bin to the minversion to all relevant manifests.

To avoid this issue of repeatedly deleting content that should have been deleted once (when considering `swupd update` only), I think mixer should avoid modifying versions of deleted files, even in the event of a full or partial minversion bump.

matthewrsj:
Fixes #455
Minversions should not update deleted files to the minversion as this
may cause swupd to attempt to delete content twice. This exacerbates
issues where a directory (/usr/local/bin) was mistakenly deleted in a
version because it was mistakenly added earlier. Updating the delete to
the minversion then causes the directory to be deleted off the client
system again.

### DIFF
--- a/swupd/create_manifests_test.go
+++ b/swupd/create_manifests_test.go
@@ -400,8 +400,8 @@ func TestCreateManifestsMVDeletes(t *testing.T) {
 	// now create a minversion and make sure the deletes persist
 	ts.MinVersion = 30
 	ts.createManifests(30)
-	fileDeletedInManifest(t, ts.parseManifest(30, "test-bundle"), 30, "/foo")
-	fileDeletedInManifest(t, ts.parseManifest(30, "full"), 30, "/foo")
+	fileDeletedInManifest(t, ts.parseManifest(30, "test-bundle"), 20, "/foo")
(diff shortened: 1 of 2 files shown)


matthewrsj:
I don't like that approach. Imagine the following:
* client on version 400
* file deleted in 500
* minversion in 600
* client tries to update to 700, file is marked as ghosted because of the minversion, doesn't actually get deleted.

phmccarty:
@matthewrsj Yep, you're right. I approve this approach then.

cmarcelo:
(I also thought we got rid of ghosted files :-P)

LGTM.

cmarcelo:
Note: we lose the property of minversion being the minimum version being mentioned in the files of the Manifest, but I guess for deleted files '00000...' this won't be a problem.

PR #459, opened 2018-10-10
swupd: increase test coverage

matthewrsj:
Improves test coverage for following files:
* archive.go up to >80%
* manifest_format.go to 100%
* packs.go up to >85%
* fullfiles.go up to >80%
* fullchroot.go add explicit test

### DIFF
--- a/swupd/archive_test.go
+++ b/swupd/archive_test.go
@@ -1,0 +1,115 @@
+package swupd
+
+import (
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestNewCompressedTarReaderUncompressedRegularFile(t *testing.T) {
+	testCases := []string{
+		"/etc/protocols",
+		"/usr/share/defaults/etc/protocols",
+		"/usr/share/doc/systemd/LICENSE.GPL2",
+	}
+	for _, tc := range testCases {
+		var f *os.File
+		var err error
+		if f, err = os.Open(tc); err != nil {
+			// not checking non-existent files
+			// not testing the os.Open call
+			continue
+		}
+		defer func() {
+			_ = f.Close()
+		}()
+		_, err = NewCompressedTarReader(f)
+		if err != nil {
+			t.Errorf("NewCompressedTarReader returned error %v for uncompressed regular file %s",
+				err, tc)
+		}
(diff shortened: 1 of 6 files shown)


rchiossi:
What are the other 20%?

matthewrsj:
All the error returns from the uncompressors.

matthewrsj:
I need to look into https://godoc.org/github.com/golang/mock/gomock more, but it is not easy to test those without mocking the external uncompressors out.

swupd/archive_test.go
```diff
@@ -0,0 +1,115 @@
...
+	testCases := []string{
+		"/etc/protocols",
+		"/usr/share/defaults/etc/protocols",
+		"/usr/share/doc/systemd/LICENSE.GPL2",
+	}
+	for _, tc := range testCases {
+		var f *os.File
+		var err error
+		if f, err = os.Open(tc); err != nil {
+			// not checking non-existent files
```
tmarcu: :+1:

swupd/archive_test.go
```diff
@@ -0,0 +1,115 @@
...
+			// not testing the os.Open call
+			continue
+		}
+		defer func() {
+			_ = f.Close()
+		}()
+		_, err = NewCompressedTarReader(f)
```
tmarcu: I realize now this is a not good name to really explain that this is actually opening and uncompressing a file from <whatever> compression type. :+1: though

swupd/packs_test.go
```diff
@@ -686,3 +692,188 @@ func checkFileInPack(t *testing.T, packname, name string) {
...
+		if ps != tc.exp {
+			t.Errorf("PackState.String() returned \"%s\"; expected \"%s\"", ps, tc.exp)
+		}
+	}
+
+	p := PackedFullfile
+	p++
+	ps := p.String()
+	if ps != "invalid" {
+		t.Errorf("invalid PackState.String() returned \"%s\"; expected \"invalid\"", ps)
```
tmarcu: did you mean to say "valid PackState.String() returned" ?

tmarcu: :question:

matthewrsj: err no, but I'll update it to be a little easier to understand.

PR #460, opened 2018-10-10
'mixer config' improvements

rchiossi:
This PR allows `mixer config` subcommand to also manipulate `mixer.state` file. It also adds a the `mixer config get ` subcommand to read values in the format `Section.Property`.

### DIFF
--- a/config/base.go
+++ b/config/base.go
@@ -1,0 +1,81 @@
+// Copyright © 2018 Intel Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// BaseConfig is a generic interface for TOML configs in mixer
+type BaseConfig interface {
+	Save() error
+}
+
+// SetProperty parses a property in the format "Section.Property", finds and sets it within the
+// config structure and saves the config file.
+func SetProperty(bc BaseConfig, propertyStr string, value string) error {
+	tokens := strings.Split(propertyStr, ".")
+	property, sections := tokens[len(tokens)-1], tokens[:len(tokens)-1]
+
+	sectionV := reflect.ValueOf(bc).Elem()
+	for i := 0; i < len(sections); i++ {
+		sectionV = sectionV.FieldByName(sections[i])
+
+		if !sectionV.IsValid() {
+			return fmt.Errorf("Unknown config sectionV: '%s'", tokens[i])
+		}
+	}
+
+	sectionT := reflect.TypeOf(sectionV.Interface())
+	for i := 0; i < sectionV.NumField(); i++ {
+		tag, ok := sectionT.Field(i).Tag.Lookup("toml")
+
+		if ok && tag == property {
+			sectionV.Field(i).SetString(value)
+			return bc.Save()
+		}
+	}
+
+	return fmt.Errorf("Property not found in config file: '%s'", property)
+}
+
+// GetProperty parses a property in the format Section.Property, finds the property and returns its
+// current value
+func GetProperty(bc BaseConfig, propertyStr string) (string, error) {
+	tokens := strings.Split(propertyStr, ".")
+	property, sections := tokens[len(tokens)-1], tokens[:len(tokens)-1]
+
+	sectionV := reflect.ValueOf(bc).Elem()
+	for i := 0; i < len(sections); i++ {
+		sectionV = sectionV.FieldByName(sections[i])
+
+		if !sectionV.IsValid() {
+			return "", fmt.Errorf("Unknown config sectionV: '%s'", tokens[i])
+		}
+	}
+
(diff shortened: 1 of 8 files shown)


rchiossi:
Hmm, it is a bit more complicated than it looks like, but let me give it a try.

config/base.go
```diff
@@ -0,0 +1,81 @@
...
+package config
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// BaseConfig is a generic interface for TOML configs in mixer
+type BaseConfig interface {
```
tmarcu: In the effort of writing more usable code, I'd change `BaseConfig` to just `Config`. I am not a fan of **Base** `Base Classes` :) i.e,
```
class Dog {}
class Labrador : public Dog {} 
```

makes a lot more sense than 
```
class BaseDog{} 
class Labrador : public BaseDog{}
```

What is a BaseDog? :smiley:

config/config.go
```diff
@@ -185,6 +185,33 @@ func (config *MixConfig) SetProperty(propertyStr string, value string) error {
 	return errors.Errorf("Property not found in config file: '%s'", property)
 }
 
+// GetProperty parse a property in the format Section.Property, finds the property and returns its
+// current value
+func (config *MixConfig) GetProperty(propertyStr string) (string, error) {
```
tmarcu: This duplicates the code in the later PR it seems?

config/base.go
rchiossi: I agree, but the reason why I don't wanna call it "Config" is to avoid naming conflict. It happened when we had a global variable named "config" in `cmd/` and I also regret naming the file that handles builder conf `config/config.go`. on top of that we also have `swupd/config.go` that declares a structure named "Config".
So yeah, I don't wanna add yet another thing to the salad.

config/config.go
rchiossi: It is moved to BaseConfig in that later commit. I din't rewrite the first two commits since I feel they still have a logical progression and that could save me some time. I got nothing against re-working the order though.

config/base.go
tmarcu: `Config` and `config` are not naming conflicts in Go. If it helps to rename the files please feel free to, they don't all have to be named config.go, however since they're namespaced by package it won't cause any conflicts since they will be prefaced by `swupd.<etc> `and `config.<etc>` whenver they're used elsewhere.

rchiossi: And that is the reason why the config inside `swupd` hasn't moved into the `config` package, which would make much more sense. Also the reason why `swapd` package doesn't use `config` package, since if conflicts with variables named `config`.

But ok, lets rename stuff:
- base.go -> config.go
- config.go -> mixconfig.go
- state.go -> mixstate.go
- BaseConfig -> Config

Sounds good?

tmarcu: Sounds great :smile:

config/base.go
```diff
@@ -0,0 +1,81 @@
...
+	Save() error
```
matthewrsj: Any reason not to add `SetProperty` and `GetProperty` to this interface?

matthewrsj: Also `Load`? (instead of `LoadConfig`, which would be a stutter)

rchiossi: Save is the only item in the interface because it is the only shared method needed for the feature. Moving into a more generic definition, `Load` would be a nice inclusion.

As for `SetProperty` and `GetProperty`, they have a single implementation for all sub-classes, unlike `Load`and `Save` that are subclass specific. Ideally they would be methods of the `BaseConfig` class and the subclasses would just inherit them (instead of overloading them), but since there is no such a thing in go, we keep them out of the interface.

matthewrsj: example: https://github.com/matthewrsj/encap

rchiossi: So, If I get this right, your suggestion is to have `ConfigInterface` which define the methods, an empty `Config` structure that allows me to implement `SetProperty` and `GetProperty` and the other classes that will include `Config` as a property. Is it right?

matthewrsj: Right but I wouldn't call it `ConfigInterface`, probably just `Config`
```go
type Config interface {
    Save() error
    Load(string) error
    SetProperty(string, string) error
    GetProperty(string) (string, error)
}

type Base struct {
    // whatever
}

func (b *Base) Save() error {
}

func (b *Base) Load(s string) error {
}

func (b *Base) SetProperty(k, v string) error {
}

func (b *Base) GetProperty(k string) (string, error) {
}

type MixConfig struct {
    Base
}

// we need a different for GetProperty for MixConfig for whatever reason
func (m *MixConfig) GetProperty(k string) (string, error) {
}
```
(I might have gotten some of the function sigs wrong)
This allows you to call `m.SetProperty` since it is defined on the embedded base struct. Basically just a different syntax for class inheritance.

matthewrsj: And that `Base` struct wouldn't necessarily have to be empty. Ideally it should have all the common members between all subclasses, if there is commonality.

rchiossi: Well, to be honest I'm not convinced with this solution. It seems to make it much more complicated than it needs to be. You get one extra layer of abstraction which makes the code harder to follow and get little benefit out of it.

My personal opinion is that it is better to call `SetProperty(m)` instead of having all this code to call `m.SetProperty()`.

tmarcu: Ok,  I understand the reasoning for keeping `SetPropery()` and `GetProperty() `separate and having the first parameter be a Config interface, that works fine, it means we can pass _any_ config into them. The `save` & `load` should be both declared in the interface though since they may have diff implementations, and the interface should be renamed to just `type Configuration interface`.  It sounds ok to name things "base" until you expand your implementation and support to where that name becomes ambiguous/wrong. You want to maintain an "is-a" relationship between objects, but not through their naming IMHO.

rchiossi:
New version is a rework of this patch, so I'm opening a new PR.

Issue #448, opened 2018-10-02
`mixer bundle edit` can create a bundle with invalid name


phmccarty:
Steps to reproduce (from a empty workspace):

```
$ mixer init
$ mixer bundle edit foo.bar
<edit bundle definition and save>
Error parsing bundle "foo.bar": Invalid bundle name "foo.bar" derived from file "/home/pmccarty/mix/local-bundles/foo.bar"
Would you like to edit as-is, revert and edit, or skip [Edit/Revert/Skip]?:
```

No matter what option I select (Edit, Revert, or Skip), I cannot get mixer out of this bad state. For example, if I choose Skip, I can successfully remove the `foo.bar` bundle, but mixer then gets confused about the .orig file it left behind:

```
$ mixer bundle remove --local foo.bar
Warning: Using Format=25 from mixer.state for this build.
Removing bundle file for "foo.bar" from local-bundles
Bundle "foo.bar" not found in mix; skipping
$ mixer bundle list
ERROR: Invalid bundle name "foo.bar.orig" derived from file "/home/pmccarty/mix/local-bundles/foo.bar.orig"
```

Perhaps mixer should not try to edit a freshly created bundle definition if the bundle name is invalid?

reaganlo:
@phmccarty @tmarcu Apologies, but this is going to be a long winded question.

Currently if a user creates a bundle using **Mixer** with an invalid name, the invalid bundle as well as its back up is created. I've done a quick fix where the invalid bundle as well as its backup does not get created.

But while testing a scenario where the user **manually** copies bundles from elsewhere with invalid names to the local-bundles folder, I came across another issue that "mixer bundle list" generates an invalid bundle name error. Ideally it should **not effect** this command because the invalid bundles are only in the local-bundles folder and not in the mix (mixbundles file). While working on a fix for this, I noticed that bundles are validated at various places in the code but not in a consistent manner. I can try to make it consistent which brings me to the main question - why do we need two levels of bundle validation (Basic and Strict). Can we just have one level (Strict) that is used consistently throughout the code.

tmarcu:
@reaganlo  ^

reaganlo:
Fixed and merged as part of https://github.com/clearlinux/mixer-tools/pull/490

PR #463, opened 2018-10-12
bundles: Fixes bundle edit issue (#448)

reaganlo:
Fixes bundle edit to not create bundle and backup if name is invalid.

### DIFF
--- a/builder/bundleset.go
+++ b/builder/bundleset.go
@@ -128,15 +128,17 @@ type ValidationLevel int
 // Enum of available validation levels
 const (
 	BasicValidation ValidationLevel = iota
+	BasicValidationIgnoreName
 	StrictValidation
 )
 
 func validateBundleFile(filename string, lvl ValidationLevel) error {
 	var errText string
 
-	// Basic Validation
-	if err := validateBundleFilename(filename); err != nil {
-		errText = err.Error() + "\n"
+	if lvl != BasicValidationIgnoreName {
+		if err := validateBundleFilename(filename); err != nil {
+			errText = err.Error() + "\n"
+		}
 	}
 
 	b, err := parseBundleFile(filename)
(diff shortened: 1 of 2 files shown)


builder/bundleset.go
```diff
@@ -128,15 +128,17 @@ type ValidationLevel int
...
 	StrictValidation
 )
 
 func validateBundleFile(filename string, lvl ValidationLevel) error {
 	var errText string
 
-	// Basic Validation
-	if err := validateBundleFilename(filename); err != nil {
-		errText = err.Error() + "\n"
+	if lvl != BasicValidationIgnoreName {
```
matthewrsj: why not `if lvl == BasicValidation {`? Imagine we add a new level called `OnlyValidateHeader`. In that case it would also `validateBundleFilename` which would not be what we want to happen.

builder/bundleset.go
```diff
@@ -145,22 +147,17 @@ func validateBundleFile(filename string, lvl ValidationLevel) error {
...
-		}
-		return nil
-	}
-
 	// Strict Validation
-	err = validateBundle(b)
-	if err != nil {
-		errText += err.Error() + "\n"
-	}
+	if lvl == StrictValidation {
```
matthewrsj: All this logic should be replaced with a
```go
switch {
case lvl >= basicValidation:
    // basic validation only
    fallthrough
case lvl >= validateName:
    // only validate the name of the bundle
    fallthrough
case lvl >= validateSomeOtherSpecificThing:
    // only validate some other specific thing
    fallthrough
case lvl >= strictValidation:
    // additional strict validation
default:
    // no validation?
}
```
Now this means you have actual *levels* instead of being able to select validation. If you want to let the user select validation you can essentially use the same switch just replace the `>=` with `==`.

builder/bundleset.go
reaganlo: `lvl != BasicValidationIgnoreName` was used instead of `lvl == BasicValidation` so that the definition of BasicValidation does not change.

builder/bundleset.go
reaganlo: Wouldn't `fallthrough` execute the first statement in the next `case`? E.g if lvl is `BasicValidation`, all the cases including `StrictValidation` will be executed.

matthewrsj: eh maybe I have the cases backwards oops. Yeah `StrictValidation` would be first and would `fallthrough` to the basic validation.

builder/bundleset.go
matthewrsj: Ok, but why do you call it `BasicValidationIgnoreName`? This is becoming a mess with different levels of `BasicValidation`. I think this needs to be rethought before you modify the concept of `BasicValidation` by providing a way to do it except for one of the checks. Next step I see happening is `StrictValidationPlusExtraCheck` or `StrictValidationIgnoreHeader`. If this is re-thought to either 1) have a menu of validation options or 2) have clear validation levels it will be much cleaner and more maintainable.

matthewrsj: my point is, you might not be modifying the in-code definition of `BasicValidation` but you are overloading the concept of basic validation. It ends up being confusing and takes a step away from maintainable code. Not claiming the code was particularly maintainable before, but we should takes steps in the right direction.

tmarcu:
@reaganlo Bump?

matthewrsj:
rebase as well...

reaganlo:
Closing this PR and instead opening PR #490 with increased scope

PR #464, opened 2018-10-16
AFB: also update version in /usr/share/clear/version

phmccarty:
The version numbers in /usr/lib/os-release and /usr/share/clear/version
should be identical, so overwrite the latter file with the correct
value.


tmarcu:
@phmccarty sorry this can't merge until the travis errors pass, so I will merge it once the linter fixes go in. Looks ok otherwise.

PR #465, opened 2018-10-16
lint: Fix unchecked return on Fprintf

rchiossi:
This patch adds error checking for all uses of Fprintf and changes
Warning messages that were using stderr to use stdout instead to keep
then consistent with the rest of the code.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -354,7 +354,7 @@ func (b *Builder) ReadVersions() error {
 	}
 	ver, err = ioutil.ReadFile(filepath.Join(b.Config.Builder.VersionPath, b.UpstreamURLFile))
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "WARNING: %s/%s does not exist, run mixer init to generate\n", b.Config.Builder.VersionPath, b.UpstreamURLFile)
+		fmt.Printf("Warning: %s/%s does not exist, run mixer init to generate\n", b.Config.Builder.VersionPath, b.UpstreamURLFile)
 		b.UpstreamURL = ""
 	} else {
 		b.UpstreamURL = strings.TrimSpace(string(ver))
@@ -1046,7 +1046,9 @@ func (b *Builder) ListBundles(listType listType, tree bool) error {
 			if _, exists := bundles[bundle]; !exists {
 				included = "(included)"
 			}
-			fmt.Fprintf(tw, "%s\t%s\t%s\n", bundle, location, included)
+			if _, err := fmt.Fprintf(tw, "%s\t%s\t%s\n", bundle, location, included); err != nil {
+				return err
+			}
 		}
 	case LocalList:
 		// Only print the top-level set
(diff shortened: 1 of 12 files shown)


rchiossi:
Yeah, I know... It is painful to watch.

jmb82:
Getting rid of lint problems sounds good.  I'm not so sure on having the errors going to stdout.  Why not leave errors in stderr?  That way errors can be viewed independently from other stuff.  If you want to save both to a file adding a 2>&1 in is not that hard.  Just my $0.02....

rchiossi:
Errors go to stderr, warnings to stdout. I'm not saying that is the best approach, but that is to keep it at least consistent since the great majority of warning messages go to stdout.

There is a lot of stuff in the legacy code that is inconsistent throughout the code. Printing is just one of them. Just for printing we have things going arbitrarily to stderr or stdout, prints using `fmt` package, using our own `helper.PrintError()` or using go's `log` package. A major overhaul will be required to normalize the code. That is the scope of a different PR though,

jmb82:
That's fair enough.  In general, I'd expect warnings and errors to go to stderr and all other output (progress, debug info, etc.) to go to stdout.  This type of tweaking can await some future overhaul, though.

tmarcu:
Why don't we just have a `PrintErr() `function that does the Fprintf to stderr and checking inside of that,  so our codebase doesn't get polluted completely as we will be adding even more fprintfs in the future. It will change all the fprintf lines to not have to do checks anymore, and they're shorter name-wise too.

matthewrsj:
+1 to that idea but maybe that should be another PR?

tmarcu:
@matthewrsj Sure, so what do you expect for this one currently then? Shall we merge it as is and then make the appropriate PrintErr() changes right after?

rchiossi:
> 1. that's not what this PR is for and

You could say that this PR should just fix the error and not change mixer's bahavior, but that is just not possible unless you ignore all the errors in all fprintf instances. Handling errors will necessarily change mixer's behavior and since if we wanna handle this right, it will have to change. Yes, I could just ignore all the errors and make make lint happy, but that will just make then harder to track in the future.

> 2. you aren't actually going through the whole codebase to make it consistent, so don't just redirect some warnings to stdout just because.

I actually looked thought the whole codebase. With the exception on the legacy code that uses go`s `log` package (which is only the `swupd/fullfiles.go`), every single Warning is now consistent. I can even change the 3 instances in fullfiles in this PR so everything is on `fmt`. 
Now, this change was far from "just because" since it was explained, discussed and agreed upon on IRC before it was implemented. You are free to disagree, just don't believe those things come out of the blue.

matthewrsj:
I agree with using the `log` package btw.

rchiossi:
> So ignore all the errors! Who cares if it fails to print to stderr. It shouldn't interrupt the continuing flow of mixer execution. We were ignoring the errors before, so now just explicitly do it by assigning it to the reserved variable _.

Not all instances of Fprintf are printing error messages. For the ones that do, they do get ignored. The ones that are writing to config file must be handled (so it crashes on creation, not on use) and that will change the behavior of mixer.

I'll rewrite this PR and rename it to something else. Fixing lint will just be a side effect.

matthewrsj:
I wasn't talking about ignoring those errors, I was talking about ignoring errors when writing to stderr. Obviously we should check errors writing to actual files and I see that as in-scope for the changes in the PR. What I don't see as in-scope is changing WARNINGs to print to stdout instead of stderr.

PR #466, opened 2018-10-17
Fprintf combo

rchiossi:
This PR add fixes and improvements to different scenarios where `Fprint*` is used in the code. Each commit can be viewed as a separate PR since they are handling different issues, but since they are needed together to pass lint testing and they are fairly simple individually, they can be review all at once.

The first 3 commits add proper error handling to different parts of the code. The last commit changes the code to use `log` package to handle error and warning messages.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -588,7 +588,7 @@ func (b *Builder) buildBundles(set bundleSet) error {
 
 	// Write INI files. These are used to communicate to the next step of mixing (build update).
 	var serverINI bytes.Buffer
-	fmt.Fprintf(&serverINI, `[Server]
+	_, _ = fmt.Fprintf(&serverINI, `[Server]
 emptydir=%s/empty
 imagebase=%s/image/
 outputdir=%s/www/
@@ -609,7 +609,7 @@ src=%s
 	// in bundleset to check for that. See also readGroupsINI in swupd package.
 	var groupsINI bytes.Buffer
 	for _, bundle := range set {
-		fmt.Fprintf(&groupsINI, "[%s]\ngroup=%s\n\n", bundle.Name, bundle.Name)
+		_, _ = fmt.Fprintf(&groupsINI, "[%s]\ngroup=%s\n\n", bundle.Name, bundle.Name)
 	}
 	err = ioutil.WriteFile(filepath.Join(b.Config.Builder.ServerStateDir, "groups.ini"), groupsINI.Bytes(), 0644)
 	if err != nil {
@@ -642,7 +642,7 @@ src=%s
 		// TODO: Should we embed this information in groups.ini? (Maybe rename it to bundles.ini)
 		var includes bytes.Buffer
 		for _, inc := range bundle.DirectIncludes {
-			fmt.Fprintf(&includes, "%s\n", inc)
+			_, _ = fmt.Fprintf(&includes, "%s\n", inc)
 		}
 		err = ioutil.WriteFile(filepath.Join(buildVersionDir, name+"-includes"), includes.Bytes(), 0644)
 		if err != nil {
@@ -833,7 +833,10 @@ func createVersionsFile(baseDir string, packagerCmd []string) error {
 	for _, e := range versions {
 		// TODO: change users of "versions" file to not rely on this exact formatting (version
 		// starting at column 51). E.g. this doesn't handle very well packages with large names.
-		fmt.Fprintf(w, "%-50s%s\n", e.name, e.version)
+		_, err = fmt.Fprintf(w, "%-50s%s\n", e.name, e.version)
+		if err != nil {
+			return err
+		}
 	}
 	return w.Flush()
 }
@@ -861,7 +864,7 @@ func fixOSRelease(filename, version string) error {
 		if strings.HasPrefix(text, "VERSION_ID=") {
 			text = "VERSION_ID=" + version
 		}
-		fmt.Fprintln(&newBuf, text)
+		_, _ = fmt.Fprintln(&newBuf, text)
 	}
 
 	err = scanner.Err()
--- a/builder/stopwatch.go
+++ b/builder/stopwatch.go
@@ -1,8 +1,8 @@
 package builder
 
 import (
-	"fmt"
 	"io"
+	"log"
 	"time"
 )
 
@@ -21,11 +21,12 @@ type stopWatchEntry struct {
 }
 
 func (sw *stopWatch) Start(name string) {
+	logger := log.New(sw.w, "", log.Ldate|log.Ltime)
 	if sw.w != nil {
 		if len(sw.entries) > 0 {
-			fmt.Println()
+			logger.Println(sw.w, "")
 		}
-		fmt.Fprintf(sw.w, "=> %s\n", name)
+		logger.Printf("=> %s\n", name)
 	}
 	sw.entries = append(sw.entries, stopWatchEntry{name: name})
 	sw.t = time.Now()
@@ -46,6 +47,7 @@ func (sw *stopWatch) Stop() {
 }
 
 func (sw *stopWatch) WriteSummary(w io.Writer) {
+	logger := log.New(w, "", log.Ldate|log.Ltime)
 	if len(sw.entries) == 0 {
 		return
 	}
@@ -56,10 +58,10 @@ func (sw *stopWatch) WriteSummary(w io.Writer) {
 		}
 	}
 	var sum time.Duration
-	fmt.Fprintf(w, "\nTIMINGS\n")
+	logger.Printf("\nTIMINGS\n")
 	for _, e := range sw.entries {
-		fmt.Fprintf(w, "  %-*s %s\n", max, e.name, e.d.Truncate(time.Millisecond))
+		logger.Printf("  %-*s %s\n", max, e.name, e.d.Truncate(time.Millisecond))
 		sum += e.d
 	}
-	fmt.Fprintf(w, "TOTAL: %s\n", sum.Truncate(time.Millisecond))
+	logger.Printf("TOTAL: %s\n", sum.Truncate(time.Millisecond))
 }
--- a/helpers/helpers.go
+++ b/helpers/helpers.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"math/big"
 	"net/http"
 	"os"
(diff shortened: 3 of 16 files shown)


helpers/helpers.go
```diff
@@ -321,22 +322,30 @@ func RunCommandOutputEnv(cmdname string, args []string, envs []string) (*bytes.B
...
 	cmd.Stderr = &errBuf
 	cmd.Env = append(os.Environ(), envs...)
-	err := cmd.Run()
+	runError := cmd.Run()
 
-	if err != nil {
+	if runError != nil {
 		var buf bytes.Buffer
-		fmt.Fprintf(&buf, "failed to execute %s", strings.Join(cmd.Args, " "))
+		if _, err := fmt.Fprintf(&buf, "failed to execute %s", strings.Join(cmd.Args, " ")); err != nil {
```
tmarcu: This will also just print several of the same warning to stderr as it goes down this function.
Since we return an error at the end of this function, I also don't care if writing to the buffer fails. If writing to a bytes.Buffer fails, probably the whole system is hosed anyway to actually continue mixing. I think writing to actual *Files* is the only place we really need to check for errors on writing, because those can fail for a variety of subtle reasons.

tmarcu: You could change all the `fmt.Fprintf()` calls to `log.Printf()` setting the output to be buf, or making a logger var using buf as the io.Writer, and then we get to remove all the "Warning" lines and change all the Fprintf() lines to 1 line since log.Prints don't return an error to check.

builder/stopwatch.go
```diff
@@ -23,9 +24,14 @@ type stopWatchEntry struct {
 func (sw *stopWatch) Start(name string) {
 	if sw.w != nil {
 		if len(sw.entries) > 0 {
-			fmt.Println()
+			if _, err := fmt.Fprintln(sw.w, ""); err != nil {
```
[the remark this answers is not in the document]
matthewrsj: Speaking of `log.SetOutput` what is the decision on sending warnings to `stderr` vs `stdout`?

helpers/helpers.go
```diff
@@ -39,19 +40,14 @@ import (
...
-
 // CreateCertTemplate will construct the template for needed openssl metadata
 // instead of using an attributes.cnf file
 func CreateCertTemplate() *x509.Certificate {
 	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
 	serialnumber, err := rand.Int(rand.Reader, serialNumberLimit)
 	if err != nil {
-		fmt.Println("ERROR: Failed to generate serial number")
-		PrintError(err)
+		log.Println("ERROR: Failed to generate serial number")
```
matthewrsj: I know this is unrelated to your change, but since you are here can you just print the error on a single line? Much easier when grepping output or log files.

helpers/helpers.go
```diff
@@ -73,8 +69,8 @@ func CreateCertTemplate() *x509.Certificate {
 func CreateKeyPair() (*rsa.PrivateKey, error) {
 	rootKey, err := rsa.GenerateKey(rand.Reader, 4096)
 	if err != nil {
-		fmt.Println("ERROR: Failed to generate random key")
-		PrintError(err)
+		log.Println("ERROR: Failed to generate random key")
```
matthewrsj: Same here

helpers/helpers.go
matthewrsj: > If writing to a bytes.Buffer fails, probably the whole system is hosed anyway to actually continue mixing.

That's a mighty big assumption to make. And if you are wrong...

> I think writing to actual Files is the only place we really need to check for errors on writing

writing to the file could actually pass but it may be writing a corrupted buffer.

matthewrsj: I personally like printing the errors as warnings here.

tmarcu: > That's a mighty big assumption to make. And if you are wrong...
We make it all the time in swupd! <function>_or_die()

> writing to the file could actually pass but it may be writing a corrupted buffer.

What does it mean to corrupt your buffer in Go?

builder/stopwatch.go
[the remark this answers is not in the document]
tmarcu: Since the default output is stderr with log, so we can just use that directly for anything that is a warning/err.

matthewrsj:
Generally this looks ok to me. There are a few things I think we should consider.

I still think we should print warnings and errors to stderr, progress to stdout. We've discussed this some and it would be easy to implement. Just create a global error logger
```go
errLogger := log.New(os.Stderr, ...)
```
and just use that everywhere you do a warning or an error. You can even get more explicit
```go
warnLogger := log.New(os.Stderr, "WARNING: ", 0)
errLogger := log.New(os.Stderr, "ERROR: ", 0)
```
and use each as appropriate.

To really improve our logging we could start using multiwriters to write both to a file and to stderr.
```go
mw := io.MultiWriter(os.Stderr, logFile)
warnLogger := log.New(mw, "WARNING: ", 0)
warnLogger := log.New(mw, "ERROR: ", 0)
```
and we can play with the flags as well to add time information.

matthewrsj:
note: I thought until just now that the default output for log was stdout, not stderr.

helpers/helpers.go
matthewrsj: > We make it all the time in swupd! function_or_die()

No we make the opposite assumption than you are. If a write to memory fails we just die. You are saying allow it to happen and just write to the file anyways.

Maybe we should go the swupd route and just fail immediately then? Is that what you are suggesting? I'd be ok with that.

> What does it mean to corrupt your buffer in Go?

idk, since I'm not sure why a buffer write would fail other than oom. Seems like a good argument for just failing at the buffer write to me.

builder/stopwatch.go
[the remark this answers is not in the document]
matthewrsj: what about for progress and information? All that should go to stderr as well?

tmarcu:
@matthewrsj me too, I only know cause I looked at the implementation the other day https://golang.org/src/log/log.go?s=8948:8992#L73 :sweat_smile:

matthewrsj:
@tmarcu I couldn't even find any mention of stderr in the godoc, I had to find the info on stack overflow :roll_eyes:

helpers/helpers.go
matthewrsj: Ok so digging through the go source code looks like an `Fprintf` on a `bytes.Buffer` will **never** return an `error`.

The `Fprintf` just calls the `Write` method of the passed `io.Writer`, in this case a `bytes.Buffer`:
https://golang.org/src/fmt/print.go?s=4979:5056#L176
```go
func Fprintf(w io.Writer, format string, a ...interface{}) (n int, err error) {
	p := newPrinter()
	p.doPrintf(format, a)
	n, err = w.Write(p.buf)
	p.free()
	return
}
```
You can see it returns the `err` from `w.Write`, but for a `bytes.Buffer.Write` `err` will always be `nil`:
https://golang.org/src/bytes/buffer.go?s=5775:5826#L160
```go
func (b *Buffer) Write(p []byte) (n int, err error) {
	b.lastRead = opInvalid
	m, ok := b.tryGrowByReslice(len(p))
	if !ok {
		m = b.grow(len(p))
	}
	return copy(b.buf[m:], p), nil
}
```

So we definitely can throw away the error when writing to a buffer.

rchiossi:
In this new revision I removed the warnings in stopwatch and helpers and used go's logger to handle prints. I had to create new logger objects since just setting `log.SetOutput` changes the output of the global logger. I also incorporated the one line print suggested by @matthewrsj .

As for the MultiWriter, it can be a good addition in the future, but I don't want to drag this PR any longer since it is blocking all other PRs. We can open an issue to keep track of it as a possible enhancement.

builder/bundles.go
```diff
@@ -588,7 +588,7 @@ func (b *Builder) buildBundles(set bundleSet) error {
 
 	// Write INI files. These are used to communicate to the next step of mixing (build update).
 	var serverINI bytes.Buffer
-	fmt.Fprintf(&serverINI, `[Server]
+	_, err = fmt.Fprintf(&serverINI, `[Server]
```
matthewrsj: Don't check this error. It is just a `bytes.Buffer` which will never fail, so you can just throw it out. See earlier discussion with @tmarcu

builder/bundles.go
```diff
@@ -609,7 +612,10 @@ src=%s
 	// in bundleset to check for that. See also readGroupsINI in swupd package.
 	var groupsINI bytes.Buffer
 	for _, bundle := range set {
-		fmt.Fprintf(&groupsINI, "[%s]\ngroup=%s\n\n", bundle.Name, bundle.Name)
+		_, err = fmt.Fprintf(&groupsINI, "[%s]\ngroup=%s\n\n", bundle.Name, bundle.Name)
```
matthewrsj: Same thing, throw out the error.

builder/bundles.go
```diff
@@ -642,7 +648,10 @@ src=%s
 		// TODO: Should we embed this information in groups.ini? (Maybe rename it to bundles.ini)
 		var includes bytes.Buffer
 		for _, inc := range bundle.DirectIncludes {
-			fmt.Fprintf(&includes, "%s\n", inc)
+			_, err = fmt.Fprintf(&includes, "%s\n", inc)
```
matthewrsj: throw out the error.

builder/bundles.go
```diff
@@ -833,7 +842,10 @@ func createVersionsFile(baseDir string, packagerCmd []string) error {
 	for _, e := range versions {
 		// TODO: change users of "versions" file to not rely on this exact formatting (version
 		// starting at column 51). E.g. this doesn't handle very well packages with large names.
-		fmt.Fprintf(w, "%-50s%s\n", e.name, e.version)
+		_, err = fmt.Fprintf(w, "%-50s%s\n", e.name, e.version)
```
matthewrsj: Interestingly, a `bufio.Writer` *can* return an error so you're doing the right thing by checking it here. That's interesting.

builder/bundles.go
```diff
@@ -861,7 +873,10 @@ func fixOSRelease(filename, version string) error {
 		if strings.HasPrefix(text, "VERSION_ID=") {
 			text = "VERSION_ID=" + version
 		}
-		fmt.Fprintln(&newBuf, text)
+		_, err = fmt.Fprintln(&newBuf, text)
```
matthewrsj: don't need to check bytes.buffer errors.

rchiossi:
I dislike making changes based on the internal implementation since it may change without notice, and I'd rather not do it, but since I don't want to delay this any longer, I added the changes.

matthewrsj:
@rchiossi you are right. That's a very good point. If you want to change it back you have my support.

PR #469, opened 2018-10-19
swupd: remove os-core-update-index file creation

### LINKED ISSUE #468: Stop creating `/usr/share/clear/os-core-update-index` file
This file is no longer used as the bundle-info themselves are published with much more information.

matthewrsj:
The file was no longer being used due to the actual bundle-info files
being published in the os-core-update-index bundle. Because this file is
very large and requires an update every version remove this file.

This commit also fixes a bug where the os-core-update-index bundle was
being created for every build, regardless of changes. This was not a
visibile bug when the os-core-update-index file itself was being created
because it had changes every time. Now that that file does not change
per release we should not be writing the manifest every time.

Fixes #468

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -444,13 +444,19 @@ func CreateManifests(version uint32, minVersion uint32, format uint, statedir st
 		return nil, err
 	}
 
-	osIdxPath := filepath.Join(verOutput, "Manifest."+osIdx.Name)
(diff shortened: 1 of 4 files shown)


matthewrsj:
travis errors will be fixed by #466

phmccarty:
In the second commit's commit message: s/visibile/visible

matthewrsj:
@phmccarty fixed.

matthewrsj:
travis error fixed.

PR #471, opened 2018-10-22
 swupd/format_switch: do not fail on 0 format

matthewrsj:
It is not the responsibility of the format_switch to check for non-zero
format, as this is an internal function and validation on the format
happens well before this function is called. It simplifies the caller so
it doesn't have to handle an error that will never occur and simplifies
the function itself so it doesn't have to explicitly check for
format == 0.

Based on #470


tmarcu:
merging 470 since it is identical

PR #472, opened 2018-10-22
New format bump process

tmarcu:
Implements new format bump process internally in mixer under the format-bum new/old commands.

### DIFF
--- a/bat/tests/06-test-format-bump/run.bats
+++ b/bat/tests/06-test-format-bump/run.bats
@@ -52,14 +52,11 @@ EOF
   [ "$result" -eq 0 ]
 
   #editors bundle should not exist in +20
-  if [ -f update/www/30/Manifest.editors ]; then
-    exit 1
-  fi
+  [ ! -f update/www/30/Manifest.editors ]
+
 
   #editors bundle should not exist in +20 MoM
-  if grep "editors" update/www/30/Manifest.MoM; then
-    exit 1
-  fi
+  ! grep "editors" update/www/30/Manifest.MoM
 
   #check if +10 previous version is set to +0
   awk '$1 == "previous:" { exit $2 != 10}' update/www/20/Manifest.MoM
@@ -118,14 +115,10 @@ EOF
 
 
   #editors bundle should not exist in +20
-  if [ -f update/www/30/Manifest.editors ]; then
-    exit 1
-  fi
+  [ ! -f update/www/30/Manifest.editors ]
 
   #editors bundle should not exist in +20 MoM
-  if grep "editors" update/www/30/Manifest.MoM; then
-    exit 1
-  fi
+  ! grep "editors" update/www/30/Manifest.MoM
 
   #check if +10 previous version is set to +0
   awk '$1 == "previous:" { exit $2 != 10}' update/www/20/Manifest.MoM
--- a/bat/tests/06-test-format-bump/run.bats
+++ b/bat/tests/06-test-format-bump/run.bats
@@ -29,10 +29,10 @@ EOF
 
   #Build +10
   format=$((format+1))
-  mixer-build-format-bump-old $format > $LOGDIR/build_formatbump_new.log
+  mixer-build-format-bump-old $format > $LOGDIR/build_formatbump_old.log
 
   #Build +20
-  mixer-build-format-bump-new $format > $LOGDIR/build_formatbump_old.log
+  mixer-build-format-bump-new $format > $LOGDIR/build_formatbump_new.log
 
   #check if LAST_VER is set to +20
   test $(< update/image/LAST_VER) -eq 30
@@ -48,8 +48,10 @@ EOF
   }' update/www/20/Manifest.MoM
 
   #editors bundle should have all files marked as deleted in +10 (0's)
-  result="$(grep -E "(10|20)" update/www/20/Manifest.editors | grep -Ev "(version|previous|000000000)" | wc -l)"
-  [ "$result" -eq 0 ]
+  # NOTE* This has to be updated if the number of fields in a manifest
+  # file entry changes from 4 which it currently is at.
+  awk 'NF == 4 { if ($2 !~ /^0{64}$/) exit 1 }' update/www/20/Manifest.editors
+
 
   #editors bundle should not exist in +20
   [ ! -f update/www/30/Manifest.editors ]
@@ -94,7 +96,7 @@ EOF
 
   #Build +10
   format=$((format+1))
-  mixer-build-format-bump $format > $LOGDIR/build_formatbump_new.log
+  mixer-build-format-bump $format > $LOGDIR/build_formatbump_one_step.log
 
   #check if LAST_VER is set to +20
   test $(< update/image/LAST_VER) -eq 30
@@ -110,8 +112,10 @@ EOF
   }' update/www/20/Manifest.MoM
 
   #editors bundle should have all files marked as deleted in +10 (0's)
-  result="$(grep -E "(10|20)" update/www/20/Manifest.editors | grep -Ev "(version|previous|000000000)" | wc -l)"
-  [ "$result" -eq 0 ]
+  # NOTE* This has to be updated if the number of fields in a manifest
+  # file entry changes from 4 which it currently is at.
+  awk 'NF == 4 { if ($2 !~ /^0{64}$/) exit 1 }' update/www/20/Manifest.editors
+
 
 
   #editors bundle should not exist in +20
--- a/bat/tests/06-test-format-bump/run.bats
+++ b/bat/tests/06-test-format-bump/run.bats
@@ -7,45 +7,137 @@ setup() {
   global_setup
 }
 
-# TODO: This test should run 'mixer build upstream-format' instead
-# once the tooling required to run custom test docker images for
-# is ready.
-@test "Perform upstream format bump on last format boundary" {
+@test "Perform format bump using individual build format-bump <new/old>" {
   upstreamver=$(get-last-format-boundary)
   mixer-init-stripped-down $upstreamver 10
-
-  #Create initial version in the old format
-  mixer-build-all > $LOGDIR/build_all.log
-
-  #Save initial format
   format=$(sed -n 's/[ ]*FORMAT[ ="]*\([0-9]\+\)[ "]*/\1/p' mixer.state)
 
-  #Update to new upstream in the new format
-  mixer-upstream-update $(($upstreamver + 10)) > $LOGDIR/upstream_update.log
+  # create deprecated bundle
+  cat << EOF > local-bundles/editors
+# [TITLE]: editors
+# [DESCRIPTION]: Run popular terminal text editors
+# [STATUS]: Deprecated
+# [CAPABILITIES]:
+# [MAINTAINER]: Test Master <email@example.com>
+EOF
+  mixer bundle add editors
 
-  #Build +20
-  mixer-build-format-bump-new > $LOGDIR/build_formatbump_new.log
+  #Create initial version in the old format
+  mixer-build-bundles > $LOGDIR/build_bundles.log
 
-  #Reset upstream for +10 build
-  echo -n "$upstreamver" > upstreamversion
+  mixer-build-update > $LOGDIR/build_update.log
 
   #Build +10
-  mixer-build-format-bump-old > $LOGDIR/build_formatbump_old.log
+  format=$((format+1))
+  mixer-build-format-bump-old $format > $LOGDIR/build_formatbump_new.log
 
-  #restore upstrean
-  mv upstreamversion.bump upstreamversion
+  #Build +20
+  mixer-build-format-bump-new $format > $LOGDIR/build_formatbump_old.log
 
   #check if LAST_VER is set to +20
   test $(< update/image/LAST_VER) -eq 30
 
+  #check if deprecated bundle is marked deleted in +10
+  awk '
+  $4 == "editors" {
+    found = 1; exit $3 != 20
+  }
+  END {
+  if (found != 1)
+    exit 1
+  }' update/www/20/Manifest.MoM
+
+  #editors bundle should have all files marked as deleted in +10 (0's)
+  result="$(grep -E "(10|20)" update/www/20/Manifest.editors | grep -Ev "(version|previous|000000000)" | wc -l)"
+  [ "$result" -eq 0 ]
+
+  #editors bundle should not exist in +20
+  if [ -f update/www/30/Manifest.editors ]; then
+    exit 1
+  fi
+
+  #editors bundle should not exist in +20 MoM
+  if grep "editors" update/www/30/Manifest.MoM; then
+    exit 1
+  fi
+
+  #check if +10 previous version is set to +0
+  awk '$1 == "previous:" { exit $2 != 10}' update/www/20/Manifest.MoM
+
   #check if +20 previous version is set to +0
   awk '$1 == "previous:" { exit $2 != 10}' update/www/30/Manifest.MoM
 
-  #check if +10 previous version is set to +0
-  awk '$1 == "previous:" { exit $2 != 10}' update/www/30/Manifest.MoM
+  #check if +20 format is set to formatN
+  awk -v f=$format '$1 == "MANIFEST" {exit $2 != f}' update/www/30/Manifest.MoM
 
-  #check if builder.conf has the +20 format
-  test $(sed -n 's/[ ]*FORMAT[ ="]*\([0-9]\+\)[ "]*/\1/p' mixer.state) -eq $(($format + 1))
+  format=$((format-1))
+  #check if +10 format is set to formatN+1
+  awk -v f=$format '$1 == "MANIFEST" {exit $2 != f}' update/www/20/Manifest.MoM
 }
 
+@test "Perform format bump using top level command" {
+  upstreamver=$(get-last-format-boundary)
+  mixer-init-stripped-down $upstreamver 10
+  format=$(sed -n 's/[ ]*FORMAT[ ="]*\([0-9]\+\)[ "]*/\1/p' mixer.state)
+
+  # create deprecated bundle
+  cat << EOF > local-bundles/editors
+# [TITLE]: editors
+# [DESCRIPTION]: Run popular terminal text editors
+# [STATUS]: Deprecated
+# [CAPABILITIES]:
+# [MAINTAINER]: Test Master <email@example.com>
+EOF
+  mixer bundle add editors
+
+  #Create initial version in the old format
+  mixer-build-bundles > $LOGDIR/build_bundles.log
+
+  mixer-build-update > $LOGDIR/build_update.log
+
+  #Build +10
+  format=$((format+1))
+  mixer-build-format-bump $format > $LOGDIR/build_formatbump_new.log
+
+  #check if LAST_VER is set to +20
+  test $(< update/image/LAST_VER) -eq 30
+
+  #check if deprecated bundle is marked deleted in +10
+  awk '
+  $4 == "editors" {
+    found = 1; exit $3 != 20
+  }
+  END {
+  if (found != 1)
+    exit 1
+  }' update/www/20/Manifest.MoM
+
+  #editors bundle should have all files marked as deleted in +10 (0's)
+  result="$(grep -E "(10|20)" update/www/20/Manifest.editors | grep -Ev "(version|previous|000000000)" | wc -l)"
+  [ "$result" -eq 0 ]
+
+
+  #editors bundle should not exist in +20
+  if [ -f update/www/30/Manifest.editors ]; then
+    exit 1
+  fi
+
+  #editors bundle should not exist in +20 MoM
+  if grep "editors" update/www/30/Manifest.MoM; then
+    exit 1
+  fi
+
+  #check if +10 previous version is set to +0
+  awk '$1 == "previous:" { exit $2 != 10}' update/www/20/Manifest.MoM
+
+  #check if +20 previous version is set to +0
+  awk '$1 == "previous:" { exit $2 != 10}' update/www/30/Manifest.MoM
+
+  #check if +20 format is set to formatN
+  awk -v f=$format '$1 == "MANIFEST" {exit $2 != f}' update/www/30/Manifest.MoM
+
+  format=$((format-1))
+  #check if +10 format is set to formatN+1
+  awk -v f=$format '$1 == "MANIFEST" {exit $2 != f}' update/www/20/Manifest.MoM
+}
 # vi: ft=sh ts=8 sw=2 sts=2 et tw=80
--- a/builder/metadata.go
+++ b/builder/metadata.go
@@ -15,11 +15,13 @@
 package builder
 
 import (
+	"bufio"
 	"fmt"
 	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -147,6 +149,60 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first, 
 	}
 
 	return format, first, latest, err
+}
+
+// UpdateFormatFile update the format number in the full-chroot format file
+func (b *Builder) UpdateFormatFile(version int) error {
+	formatFile := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer, "full/usr/share/defaults/swupd/format")
+	if _, err := os.Stat(formatFile); err == nil {
+		return err
+	}
+
+	return ioutil.WriteFile(formatFile, []byte(strconv.Itoa(version)), 0644)
+}
+
+// ModifyBundles goes through the bundle directory and performs an action when it finds
+// a Deprecated bundle
+func (b *Builder) ModifyBundles(action func(string) error) error {
+	path := b.Config.Mixer.LocalBundleDir
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return err
+	}
+
+	var scanner *bufio.Scanner
+	for _, file := range files {
+		fmt.Println("CHECKING FILE: " + file.Name())
+		fileToScan := filepath.Join(path, file.Name())
+		f, err := os.Open(fileToScan)
+		if err != nil {
+			return err
+		}
+
+		// Scan the files and find which bundle definitions are marked deprecated
+		scanner = bufio.NewScanner(f)
+		var str string
+		re := regexp.MustCompile("#\\s\\[STATUS\\]:\\s*Deprecated.*")
+		for scanner.Scan() {
+			str = scanner.Text()
+			fmt.Println("Scanning: " + str)
+			// Don't scan past header, stop once we have no more # comments
+			if str[0] == '#' {
+				if index := re.FindStringIndex(str); index != nil {
+					fmt.Println("Found deprecated bundle: " + fileToScan)
+					// Call the callback function we need on the file we're scanning
+					if err = action(file.Name()); err != nil {
+						return err
+					}
+				}
+			} else {
+				_ = f.Close()
+				break
+			}
+		}
+		_ = f.Close()
+	}
+	return nil
 }
 
 // PrintVersions prints the current mix and upstream versions, and the
--- a/builder/metadata.go
+++ b/builder/metadata.go
@@ -151,16 +151,6 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first, 
 	return format, first, latest, err
 }
 
-// UpdateFormatFile update the format number in the full-chroot format file
-func (b *Builder) UpdateFormatFile(version int) error {
-	formatFile := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer, "full/usr/share/defaults/swupd/format")
-	if _, err := os.Stat(formatFile); err == nil {
-		return err
-	}
-
-	return ioutil.WriteFile(formatFile, []byte(strconv.Itoa(version)), 0644)
-}
-
 // ModifyBundles goes through the bundle directory and performs an action when it finds
 // a Deprecated bundle
 func (b *Builder) ModifyBundles(action func(string) error) error {
@@ -172,7 +162,6 @@ func (b *Builder) ModifyBundles(action func(string) error) error {
 
 	var scanner *bufio.Scanner
 	for _, file := range files {
-		fmt.Println("CHECKING FILE: " + file.Name())
 		fileToScan := filepath.Join(path, file.Name())
 		f, err := os.Open(fileToScan)
 		if err != nil {
@@ -185,7 +174,6 @@ func (b *Builder) ModifyBundles(action func(string) error) error {
 		re := regexp.MustCompile("#\\s\\[STATUS\\]:\\s*Deprecated.*")
 		for scanner.Scan() {
 			str = scanner.Text()
-			fmt.Println("Scanning: " + str)
 			// Don't scan past header, stop once we have no more # comments
 			if str[0] == '#' {
 				if index := re.FindStringIndex(str); index != nil {
--- a/builder/metadata.go
+++ b/builder/metadata.go
@@ -151,9 +151,55 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first, 
 	return format, first, latest, err
 }
 
+// ReplaceInfoBundles wipes the <bundle>-info files and replaces them with
+// <bundle>/ directories that are empty. When the manifest creation happens it will
+// mark all files in that bundles as deleted.
+func (b *Builder) ReplaceInfoBundles(filenames []string) error {
+	for _, f := range filenames {
+		bdir := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer, f)
+		// This bundle is deprecated, remove the info file
+		if err := os.Remove(bdir + "-info"); err != nil {
+			return err
+		}
+		// Create the empty dir for update to mark all files as deleted
+		if err := os.MkdirAll(bdir, 0755); err != nil {
+			return errors.Wrapf(err, "Failed to create bundle directory: %s", bdir)
+		}
+	}
+	return nil
+}
+
+// RemoveBundlesGroupINI removes the bundles from groups.ini and mixbundles so they are not
+// tracked anymore and manifests do not get created for them
+func (b *Builder) RemoveBundlesGroupINI(bundles []string) error {
+	var groups []byte
+	filename := filepath.Join(b.Config.Builder.ServerStateDir, "groups.ini")
+	groups, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return err
+	}
+	groupsini := string(groups)
+	for _, bundle := range bundles {
+		bundleToRemove := fmt.Sprintf(`\[%s\]\ngroup=%s`, bundle, bundle)
+		re := regexp.MustCompile(bundleToRemove)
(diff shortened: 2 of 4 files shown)


builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
 	return ioutil.WriteFile(filepath.Join(b.Config.Builder.VersionPath, b.MixVerFile), []byte(b.MixVer), 0644)
 }
 
+// UpdateOsRelease sets the mix version in the full-chroot os-release file
+func (b *Builder) UpdateOsRelease(version string) error {
+	filename := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer, "full/usr/lib/os-release")
```
matthewrsj: Can you make `"full/usr/lib/os-release"` a const?

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+	if err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(filename, newBuf.Bytes(), 0644)
+}
+
+// UpdateFormatFile update the format number in the full-chroot format file
+func (b *Builder) UpdateFormatFile(version int) error {
+	formatFile := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer, "full/usr/share/defaults/swupd/format")
```
matthewrsj: Can you also make this a const?

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+	if _, err := os.Stat(formatFile); err == nil {
```
matthewrsj: if it exists we don't overwrite it? Won't it exist since this is called after `build bundles` finishes and this file is created during that step?

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+		return err
```
matthewrsj: you will only ever return `nil` here.

builder/builder.go
matthewrsj: why even check for existence? shouldn't we just blindly write to it?

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+func (b *Builder) ModifyBundles(action func(string) error) error {
+	path := b.Config.Mixer.LocalBundleDir
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return err
+	}
+
+	var scanner *bufio.Scanner
+	for _, file := range files {
+		fmt.Println("CHECKING FILE: " + file.Name())
```
matthewrsj: I'm assuming this is one of those debug prints and will go away.

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+		fileToScan := filepath.Join(path, file.Name())
+		f, err := os.Open(fileToScan)
+		if err != nil {
+			return err
+		}
+
+		// Scan the files and find which bundle definitions are marked deprecated
+		scanner = bufio.NewScanner(f)
+		var str string
+		re := regexp.MustCompile("#\\s\\[STATUS\\]:\\s*Deprecated.*")
```
matthewrsj: instead of using double quotes (") use back-tics (\`) so you don't have to escape your escapes.
This seems to work
```go
re := regexp.MustCompile(`#\s\[STATUS\]:\s*Deprecated.*`)
```
https://goplay.space/#hhjZcWAeWOW

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+		for scanner.Scan() {
+			str = scanner.Text()
+			fmt.Println("Scanning: " + str)
+			// Don't scan past header, stop once we have no more # comments
+			if str[0] == '#' {
```
matthewrsj: this might be easier to read as a guard clause
```go
if str[0] != '#' {
    break
}
// the rest of your code
```

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+	files, err := ioutil.ReadDir(path)
+	if err != nil {
+		return err
+	}
+
+	var scanner *bufio.Scanner
+	for _, file := range files {
+		fmt.Println("CHECKING FILE: " + file.Name())
+		fileToScan := filepath.Join(path, file.Name())
+		f, err := os.Open(fileToScan)
```
matthewrsj: you should defer your close here so you don't have to manually handle it below.

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
 	return ioutil.WriteFile(filepath.Join(b.Config.Builder.VersionPath, b.MixVerFile), []byte(b.MixVer), 0644)
 }
 
+// UpdateOsRelease sets the mix version in the full-chroot os-release file
+func (b *Builder) UpdateOsRelease(version string) error {
```
matthewrsj: Why do you make this a method with a `*Builder` receiver? it only needs the statedir and the version to write to it. Seems a little heavy-weight to be passing the huge builder obj around (conceptually, I know it's just a pointer).

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+	err = scanner.Err()
+	if err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(filename, newBuf.Bytes(), 0644)
+}
+
+// UpdateFormatFile update the format number in the full-chroot format file
+func (b *Builder) UpdateFormatFile(version int) error {
```
matthewrsj: same comment about this having a builder receiver.

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+	if _, err := os.Stat(formatFile); err == nil {
+		return err
+	}
+
+	return ioutil.WriteFile(formatFile, []byte(strconv.Itoa(version)), 0644)
+}
+
+// ModifyBundles goes through the bundle directory and performs an action when it finds
+// a Deprecated bundle
+func (b *Builder) ModifyBundles(action func(string) error) error {
```
matthewrsj: This uses the builder receiver even less than the others.

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
 		if err != nil {
 			fail(err)
 		}
+		// Update mixer to build version +20, to build the bundles with +20 inside not +10
 		if err = b.UpdateMixVer(ver + 20); err != nil {
 			failf("Couldn't update Mix Version")
```
matthewrsj: not in your diff but this is `failf()` with no formatting and not actually passing the error in.

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
...
+		// be part of the mixbundles list and the groups.ini
+		if err = buildBundles(b, buildFlags.noSigning, buildFlags.clean); err != nil {
 			fail(err)
 		}
-		setWorkers(b)
 
-		fmt.Println(" Backing up full groups.ini")
-		// Back up groups.ini in case we have deprecated bundles to delete
-		if err = b.CopyFullGroupsINI(); err != nil {
+		// Wipes the <bundle>-info files and replaces them with
```
matthewrsj: update this comment to say it is wiping the bundle-info files of the deprecated bundles only.

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
...
+		// <bundle>/ directories that are empty. When the manifest creation happens it will
+		// mark all files in that bundles as deleted.
+		f := func(filename string) error {
+			bdir := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer, filename)
+			// This bundle is deprecated, remove the info file
+			if err = os.RemoveAll(bdir + "-info"); err != nil {
```
matthewrsj: maybe just `os.Remove`?

* bundle `foo` is deprecated
* bundle `foo-info` is not
* bundle `foo-info` has a chroot to provide the content instead of a `foo-info-info` file.
* bundle `foo` *also* has a chroot to provide the content instead of a `foo-info` file.
* this will recursively remove the `foo-info` chroot instead of the `foo` chroot.

We should just use `os.Remove` so we don't accidentally remove bundle content we don't intend to. There's probably some other sanity checking that needs to happen, like checking to make sure `foo-info` isn't a bundle itself.

note that `mixer build update` will fail if there is a `foo-info` and a `foo` bundle that both use a chroot instead of an `info` file, so this isn't a huge deal, but still definitely don't want to be removing that content on accident.

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
...
+				return err
+			}
+			// Create the empty dir for update to mark all files as deleted
+			if err = os.MkdirAll(bdir, 0755); err != nil {
```
matthewrsj: any reason to not just use `os.Mkdir`? We expect and need all parent dirs to exist already.

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
...
+		// rare).
+		ver, err = strconv.Atoi(b.MixVer)
+		if err != nil {
 			fail(err)
 		}
+		ver -= 10
+		source := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer)
+		dest := filepath.Join(b.Config.Builder.ServerStateDir, "image", strconv.Itoa(ver))
+		fmt.Println(" Copying +20 bundles to +10 bundles")
+		if err = helpers.RunCommandSilent("cp", "-al", source, dest); err != nil {
```
matthewrsj: I wish there was a way to do https://golang.org/pkg/os/#Link recursively that didn't mean implementing it yourself.

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
...
+		// underlying content. The only things that might change are the manifests
+		// (potentially the pack and full-file formats as well, though this is very
+		// rare).
+		ver, err = strconv.Atoi(b.MixVer)
+		if err != nil {
 			fail(err)
 		}
+		ver -= 10
+		source := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer)
+		dest := filepath.Join(b.Config.Builder.ServerStateDir, "image", strconv.Itoa(ver))
```
matthewrsj: can we not hard-code the original version to be `b.MixVer - 10`? there has to be somewhere else we can dynamically grab that info...

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
...
+		b.State.Mix.Format = oldFormat
 
+		// Update mixer to build version +10, the version we want to output for old format
 		ver, err = strconv.Atoi(b.MixVer)
 		if err != nil {
 			fail(err)
 		}
-
-		// Build the +20 update so we don't have to switch tooling in between
+		if err = b.UpdateMixVer(ver - 10); err != nil {
```
matthewrsj: Even if you do want to keep the hardcoded +-10 increments/decrements, at least save the versions to their own variables so I don't have to keep figuring out where `ver` is at with all the `ver -= 10` and `ver - 10`....

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
...
-			fail(err)
-		}
-		prevVersion -= 10
-		source := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer)
-		dest := filepath.Join(b.Config.Builder.ServerStateDir, "image", strconv.Itoa(prevVersion))
-		fmt.Println(" Copying +20 chroots to +10 chroots")
-		if err = helpers.RunCommandSilent("cp", "-al", source, dest); err != nil {
-			failf("Failed to copy +20 chroots to +10: %s\n", err)
+		// Write the +0 version to LAST_VER so that we reference in both +10 and +20 manifests as the 'previous:'
+		if err = ioutil.WriteFile(filepath.Join(b.Config.Builder.ServerStateDir, "image", "LAST_VER"), []byte(strconv.Itoa(ver-20)), 0644); err != nil {
```
matthewrsj: wait `ver-20`? isn't `ver` already at the +10 version, so this would put it at the -10 version?

ohhhh no you reread `ver` from `b.MixVer` and just did `ver - 10` not `ver -= 10`. This is confusing lol.

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
...
+		f := func(bundle string) error {
+			var groups []byte
+			filename := filepath.Join(b.Config.Builder.ServerStateDir, "groups.ini")
+			groups, err = ioutil.ReadFile(filename)
+			if err != nil {
+				return err
+			}
+			groupsini := string(groups)
+
+			bundleToRemove := fmt.Sprintf("\\[%s\\]\ngroup=%s", bundle, bundle)
```
matthewrsj: if you use backtics you don't have to double-escape.

mixer/cmd/build.go
```diff
@@ -199,113 +213,163 @@ var buildFormatNewCmd = &cobra.Command{
...
+			re := regexp.MustCompile(bundleToRemove)
+			groupsini = re.ReplaceAllString(groupsini, "")
+
+			// We are rarely removing more than a couple bundles, so it is ok
```
matthewrsj: ha.

my code is slow but is ok i promis

builder/builder.go
```diff
@@ -1311,6 +1312,102 @@ func (b *Builder) UpdateMixVer(version int) error {
...
+	path := b.Config.Mixer.LocalBundleDir
```
matthewrsj: also need to do this for upstreambundledir!

builder/builder.go
tmarcu: CopyPasta-ish reason for this check :) should be gone.

mixer/cmd/build.go
tmarcu: shush

matthewrsj:
Can you add a bat test for doing the whole format bump in one step?

builder/metadata.go
```diff
@@ -149,6 +151,60 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first,
 	return format, first, latest, err
 }
 
+// UpdateFormatFile update the format number in the full-chroot format file
+func (b *Builder) UpdateFormatFile(version int) error {
+	formatFile := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer, "full/usr/share/defaults/swupd/format")
```
matthewrsj: `"full/usr/share/defaults/swupd/format"` should be a const.

builder/metadata.go
```diff
@@ -149,6 +151,60 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first,
...
+	if _, err := os.Stat(formatFile); err == nil {
```
matthewrsj: I think all the old issues are still here. I'll hold off on reviewing.

tmarcu: Yes it's not updated, just rebased to builder breakage.

mixer/cmd/build.go
tmarcu: Done!

builder/builder.go
tmarcu: yes :D

mixer/cmd/build.go
tmarcu: hehe done sorry

builder/builder.go
matthewrsj: Also I like that link... `WAeWOW` like it's impressed with my code. :rofl:

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,146 @@ setup() {
...
+  fi
+
+  #editors bundle should not exist in +20
+  run grep "00000000000000" update/www/30/Manifest.editors
+  if [[ $status -lt 1 ]]; then
+    exit 1
+  fi
+
+  #editors bundle should not exist in +20 MoM
+  run grep "editors" update/www/30/Manifest.MoM
```
matthewrsj: You don't need to run these with  `run` you can just 
```bash
if grep "editors" update/www/30/Manifest.MoM; then
    # grep found something
    exit 1
fi
```

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,146 @@ setup() {
...
+  format=$(sed -n 's/[ ]*FORMAT[ ="]*\([0-9]\+\)[ "]*/\1/p' mixer.state)
+
+  # create deprecated bundle
+  cat << EOF > local-bundles/editors
+# [TITLE]: editors
+# [DESCRIPTION]: Run popular terminal text editors
+# [STATUS]: Deprecated
+# [CAPABILITIES]:
+# [MAINTAINER]: William Douglas <william.douglas@intel.com>
+# Run time requirement as vim is compiled to use libpython
```
matthewrsj: remove email, remove unrelated next line.

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,146 @@ setup() {
...
+  $4 == "editors" {
+    found = 1; exit $3 != 20
+  }
+  END {
+  if (found != 1)
+    exit 1
+  }' update/www/20/Manifest.MoM
+
+  #editors bundle should have tracking file marked as deleted in +10 (0's)
+  run grep "00000000000000" update/www/20/Manifest.editors
```
matthewrsj: See above comment about testing that all files are deleted.

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,146 @@ setup() {
...
+  if [[ $status -gt 0 ]]; then
+    exit 1
+  fi
+
+  #editors bundle should not exist in +20
+  run grep "00000000000000" update/www/30/Manifest.editors
```
matthewrsj: Just check that the file doesn't exist.

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,146 @@ setup() {
...
+  if [[ $status -lt 1 ]]; then
+    exit 1
+  fi
+
+  #editors bundle should not exist in +20 MoM
+  run grep "editors" update/www/30/Manifest.MoM
```
matthewrsj: Don't need to use `run`

matthewrsj:
Requesting @phmccarty to review the awk testing.

builder/metadata.go
```diff
@@ -149,6 +151,48 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first,
...
+		fileToScan := filepath.Join(path, file.Name())
+		f, err := os.Open(fileToScan)
+		if err != nil {
+			return err
+		}
+
+		// Scan the files and find which bundle definitions are marked deprecated
+		scanner = bufio.NewScanner(f)
+		var str string
+		re := regexp.MustCompile("#\\s\\[STATUS\\]:\\s*Deprecated.*")
```
matthewrsj: backtics to avoid double escapes and make it easier to read.

builder/builder.go
matthewrsj: NEVER MIND this is in a loop XD

builder/metadata.go
```diff
@@ -149,6 +151,48 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first,
...
+		for scanner.Scan() {
+			str = scanner.Text()
+			// Don't scan past header, stop once we have no more # comments
+			if str[0] == '#' {
```
matthewrsj: Still think this would be much easier to read as a guard clause. You end up going 6 indentation levels deep (7 counting the error handling).

builder/metadata.go
```diff
@@ -149,6 +151,48 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first,
...
+				if index := re.FindStringIndex(str); index != nil {
```
matthewrsj: Instead of `FindStringIndex` here (which requires you to check `index` against `nil`) just look for a `re.Match`

```go
if re.Match(str) {
    // code
}
```

https://golang.org/pkg/regexp/#Regexp.Match

mixer/cmd/build.go
```diff
@@ -107,6 +109,7 @@ var buildBundlesCmd = &cobra.Command{
 			fail(err)
 		}
 		setWorkers(b)
+
```
matthewrsj: unrelated whitespace change?

builder/metadata.go
matthewrsj: Also, you should definitely move this to before the for loop. You are compiling the regexp on every iteration yet it doesn't change.

mixer/cmd/build.go
```diff
@@ -166,31 +169,47 @@ var buildFormatBumpCmd = &cobra.Command{
...
+		}
 		b, err := builder.NewFromConfig(configFile)
 		if err != nil {
 			fail(err)
 		}
-		cmdToRun := strings.Split("mixer build format-bump new", " ")
-		if err := b.RunCommandInContainer(cmdToRun); err != nil {
-			fail(err)
+		// Silence linter...
+		_ = b
```
matthewrsj: If you don't use the builder why are you creating one `NewFromConfig`? Just don't create it anymore.

mixer/cmd/build.go
matthewrsj: I still think this comment should be updated, it currently makes it sound like we are wiping all the bundle-info files.

mixer/cmd/build.go
matthewrsj: bump about the `os.Remove` bug I mentioned in my previous review.

mixer/cmd/build.go
matthewrsj: bump with the question about `os.Mkdir`

mixer/cmd/build.go
```diff
@@ -199,113 +218,158 @@ var buildFormatNewCmd = &cobra.Command{
...
+			if err = os.RemoveAll(bdir + "-info"); err != nil {
+				return err
+			}
+			// Create the empty dir for update to mark all files as deleted
+			if err = os.MkdirAll(bdir, 0755); err != nil {
+				return errors.Wrapf(err, "Failed to create bundle directory: %s", bdir)
+			}
+			return nil
+		}
+		fmt.Println("Removing deprecated bundle-info files...")
```
matthewrsj: This is a debug print, the user doesn't need to know we are removing internal data files.

mixer/cmd/build.go
```diff
@@ -199,113 +218,158 @@ var buildFormatNewCmd = &cobra.Command{
...
 		}
-		// Fill this in w/Update bundle definitions
-		// if err := UpdateBudlesForFormatBump(); err != nil {...}
 
-		// Build the +20 (first build in new format) bundles
-		if err = buildBundles(b, buildFlags.noSigning, buildFlags.clean); err != nil {
+		// Wipes the <bundle>-info files and replaces them with
+		// <bundle>/ directories that are empty. When the manifest creation happens it will
+		// mark all files in that bundles as deleted.
+		f := func(filename string) error {
```
matthewrsj: I think declaring this inline makes this long function even longer than it needs to be. Any reason you don't just define it at package scope?

mixer/cmd/build.go
```diff
@@ -199,113 +218,158 @@ var buildFormatNewCmd = &cobra.Command{
...
 
+		// Link the +20 bundles to the +10 so we are building the update with the same
+		// underlying content. The only things that might change are the manifests
+		// (potentially the pack and full-file formats as well, though this is very
+		// rare).
 		ver, err = strconv.Atoi(b.MixVer)
 		if err != nil {
 			fail(err)
 		}
+		if err = b.UpdateMixVer(ver - 10); err != nil {
```
matthewrsj: You are still dancing with the `ver`. It would be much easier to follow if you had `oldFormatVer` and `newFormatVer` or similar.

mixer/cmd/build.go
```diff
@@ -199,113 +218,158 @@ var buildFormatNewCmd = &cobra.Command{
...
-		if err != nil {
-			fail(err)
-		}
-		prevVersion -= 10
-		source := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer)
-		dest := filepath.Join(b.Config.Builder.ServerStateDir, "image", strconv.Itoa(prevVersion))
-		fmt.Println(" Copying +20 chroots to +10 chroots")
-		if err = helpers.RunCommandSilent("cp", "-al", source, dest); err != nil {
-			failf("Failed to copy +20 chroots to +10: %s\n", err)
+		zeroVer := strconv.Itoa(ver - 20)
```
matthewrsj: this could be set to `oldFormatVer - 10` if it was defined, but would probably be even easier to follow if you initiated it earlier on using the last version of the mix before the format bump started or something.

mixer/cmd/build.go
```diff
@@ -199,113 +218,158 @@ var buildFormatNewCmd = &cobra.Command{
...
+		// Set mixversion to +20 build now, we only need to build update since we
+		// already have the bundles created
+		ver, err := strconv.Atoi(b.MixVer)
 		if err != nil {
 			fail(err)
 		}
-		prevFormat--
-		if err = b.UpdateFormatVersion(strconv.Itoa(prevFormat)); err != nil {
-			fail(err)
+		if err = b.UpdateMixVer(ver + 10); err != nil {
```
matthewrsj: You initiate `ver` to `b.MixVer` then immeidately start using it at another value. This is still difficult to follow.

All this logic could really use a new type
```go
type fbVersions struct {
    zeroVer, oldFormatVer, newFormatVer int
    oldFormat, newFormat int
}
```
and use that everywhere.

mixer/cmd/build.go
matthewrsj: you know, you could just pass this a slice of strings and modify the caller (ModifyBundles) to utilize that
```go
func(bundles string...) error {
}
```
Move the matching logic into the `action` func itself so it can handle multiple matches. Then you don't have to write the file a ton of times.

mixer/cmd/build.go
tmarcu: 'cause por que no

mixer/cmd/build.go
tmarcu: Huh you're right.

mixer/cmd/build.go
matthewrsj: because
> declaring this inline makes this long function even longer than it needs to be

mixer/cmd/build.go
tmarcu: hmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmmm

builder/metadata.go
tmarcu: ahhh thx

builder/metadata.go
```diff
@@ -151,16 +151,64 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first,
...
 }
 
+// replaceInfoBundles wipes the <bundle>-info files and replaces them with
+// <bundle>/ directories that are empty. When the manifest creation happens it will
+// mark all files in that bundles as deleted.
+func (b *Builder) ReplaceInfoBundles(filenames []string) error {
+	for _, f := range filenames {
+		bdir := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer, f)
+		// This bundle is deprecated, remove the info file
+		if err := os.RemoveAll(bdir + "-info"); err != nil {
```
matthewrsj: Can you just do an `os.Remove` related to the bug I pointed out in an earlier review?

builder/metadata.go
```diff
@@ -151,16 +151,64 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first,
...
+	if err != nil {
+		return err
+	}
+	groupsini := string(groups)
+	for _, bundle := range bundles {
+		bundleToRemove := fmt.Sprintf(`\[%s\]\ngroup=%s`, bundle, bundle)
+		re := regexp.MustCompile(bundleToRemove)
+		groupsini = re.ReplaceAllString(groupsini, "")
+	}
+	err = ioutil.WriteFile(filename, []byte(groupsini), 0644)
```
matthewrsj: \o/ you only write once

builder/metadata.go
```diff
@@ -151,16 +151,64 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first,
...
+func (b *Builder) ModifyBundles(action func([]string) error) error {
 	path := b.Config.Mixer.LocalBundleDir
 	files, err := ioutil.ReadDir(path)
 	if err != nil {
 		return err
 	}
 
 	var scanner *bufio.Scanner
+	var filesGroup []string
+	re := regexp.MustCompile("#\\s\\[STATUS\\]:\\s*Deprecated.*")
```
matthewrsj: ``

mixer/cmd/build.go
matthewrsj: bump

builder/metadata.go
tmarcu: fffffffffff

tmarcu:
@matthewrsj I thought we shoudl leave that removing deprecated bundle print in, so that we can actually see it in debug output if something goes wrong with removing them. What do you think?

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,137 @@ setup() {
...
+  if (found != 1)
+    exit 1
+  }' update/www/20/Manifest.MoM
+
+  #editors bundle should have all files marked as deleted in +10 (0's)
+  result="$(grep -E "(10|20)" update/www/20/Manifest.editors | grep -Ev "(version|previous|000000000)" | wc -l)"
+  [ "$result" -eq 0 ]
+
+  #editors bundle should not exist in +20
+  if [ -f update/www/30/Manifest.editors ]; then
```
matthewrsj: drop the `if` and just test it `[ ! -f update/www/30/Manifest.editors ]`

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,137 @@ setup() {
...
+    exit 1
+  fi
+
+  #editors bundle should not exist in +20 MoM
+  if grep "editors" update/www/30/Manifest.MoM; then
```
matthewrsj: You can drop the if here as well.

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,137 @@ setup() {
...
+    exit 1
+  }' update/www/20/Manifest.MoM
+
+  #editors bundle should have all files marked as deleted in +10 (0's)
+  result="$(grep -E "(10|20)" update/www/20/Manifest.editors | grep -Ev "(version|previous|000000000)" | wc -l)"
+  [ "$result" -eq 0 ]
+
+
+  #editors bundle should not exist in +20
+  if [ -f update/www/30/Manifest.editors ]; then
```
matthewrsj: just `[ test ]` it

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,137 @@ setup() {
...
+    exit 1
+  fi
+
+  #editors bundle should not exist in +20 MoM
+  if grep "editors" update/www/30/Manifest.MoM; then
```
matthewrsj: you can drop the if here like you did with one of the greps above.

tmarcu:
@matthewrsj ok updated the bat tests and it's passing!

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,137 @@ setup() {
...
+  mixer-build-bundles > $LOGDIR/build_bundles.log
 
-  #Reset upstream for +10 build
-  echo -n "$upstreamver" > upstreamversion
+  mixer-build-update > $LOGDIR/build_update.log
 
   #Build +10
-  mixer-build-format-bump-old > $LOGDIR/build_formatbump_old.log
+  format=$((format+1))
+  mixer-build-format-bump-old $format > $LOGDIR/build_formatbump_new.log
```
jwakre: The output of the old format bump is stored in build_formatbump_**new**.log

s/new/old  ?

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,137 @@ setup() {
...
 
-  #restore upstrean
-  mv upstreamversion.bump upstreamversion
+  #Build +20
+  mixer-build-format-bump-new $format > $LOGDIR/build_formatbump_old.log
```
jwakre: New format bump output stored in old log

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,130 @@ setup() {
...
+    found = 1; exit $3 != 20
+  }
+  END {
+  if (found != 1)
+    exit 1
+  }' update/www/20/Manifest.MoM
+
+  #editors bundle should have all files marked as deleted in +10 (0's)
+  result="$(grep -E "(10|20)" update/www/20/Manifest.editors | grep -Ev "(version|previous|000000000)" | wc -l)"
+  [ "$result" -eq 0 ]
```
phmccarty: Optional: if you want to use AWK instead, you could consolidate the above two lines into:

```
awk 'NF == 4 { if ($2 !~ /^0{64}$/) exit 1 }' update/www/20/Manifest.editors
```

bat/tests/06-test-format-bump/run.bats
```diff
@@ -7,45 +7,130 @@ setup() {
...
```
phmccarty: Optional: if you want to use AWK instead, see my previous comment for a one-liner.

bat/tests/06-test-format-bump/run.bats
tmarcu: ahh ty

bat/tests/06-test-format-bump/run.bats
tmarcu: yes this is a nice improvement.

builder/metadata.go
```diff
@@ -149,6 +151,92 @@ func (b *Builder) getUpstreamFormatRange(version string) (format string, first,
...
+		for scanner.Scan() {
+			str = scanner.Text()
+			// Don't scan past header, stop once we have no more # comments
+			if str[0] != '#' {
+				break
+			}
+			if !re.Match([]byte(str)) {
+				continue
+			}
+			fmt.Println("Found deprecated bundle: " + fileToScan)
```
matthewrsj: is this a debug print? this is intended to be a generic function that performs `action()`, not necessarily removes deprecated bundles.

mixer/cmd/build.go
matthewrsj: https://github.com/matthewrsj/copy :)

builder/metadata.go
tmarcu: Hmm you're right, it wasn't a generalized one before. I'll remove it for future proofing it and maybe add this to the part of the format-bump command that it's called at so we can still confirm things are getting deprecated.

PR #475, opened 2018-10-22
builder: improve error message when listing repos

### LINKED ISSUE #323: "mixin repo list" should perform "mixin repo init" if user hasn't added a repo
If a user hasn't added a repo, "$ mixin repo list" returns: 

     ERROR: stat /usr/share/mix/.yum-mix.conf: no such file or directory

I think it would be nicer if mixin simply performs "init" automatically and return the default repos like this: 

     Clear	https://download.clearlinux.org/releases/$releasever/clear/x86_64/os/
     Local	file:///usr/share/mix/local

matthewrsj:
Fixes #323
Provide a better error message when listing repositories. The error now
gives a hint to the user to initialize their workspace before running
the command.


matthewrsj:
This is related to a `mixin` issue, so labeling as such.

PR #477, opened 2018-10-23
Create DNF conf as part of mixer init

### LINKED ISSUE #452: mixer init must create .yum-mix.conf
This should be shifted left from being create in build bundles, to being created during mixer init phase, with a --repo flag that sets the default repo in it to something other than download.clearlinux. A user may want to edit this file before building their first mix because they are a full upstream building completely offline, so make it easier for them to do so from the beginning.

tmarcu:
One may wish to add or edit repos before building any mixes, but the DNF
(.yum-mix.conf) is not available until build-bundles is run.

Fixes #452

### DIFF
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -1,0 +1,51 @@
+package builder
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+)
+
+func mustExist(t *testing.T, name string) {
+	t.Helper()
+	_, err := os.Stat(name)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func cleanup(path string) {
+	_ = os.RemoveAll(path)
+}
+
+func TestNewDNFConfOnInit(t *testing.T) {
+	testDir, err := ioutil.TempDir("", "dnf-testdir-")
+	if err != nil {
+		t.Fatalf("couldn't create temporary directory to write test cases: %s", err)
+	}
+	defer cleanup(testDir)
+	conf := testDir + "/builder.conf"
+
+	b := New()
+	b.Config.LoadDefaultsForPath(testDir)
+
+	if err = b.Config.InitConfigPath(conf); err != nil {
+		t.Errorf("Failed to init default config: %s", err)
+	}
+
+	if err = b.Config.SaveConfig(); err != nil {
+		t.Errorf("Failed to create default config: %s", err)
+	}
+
+	if err = b.Config.LoadConfig(conf); err != nil {
+		t.Errorf("Failed to load default builder.conf: %s", err)
+	}
+	Offline = true
+
+	err = b.InitMix("10", "10", false, false, true, "https://example.com", false)
+	if err != nil {
+		t.Errorf("Failed to initialize mix: %s", err)
+	}
+
+	mustExist(t, testDir+"/.yum-mix.conf")
+}
--- a/config/config.go
+++ b/config/config.go
@@ -121,7 +121,7 @@ func (config *MixConfig) CreateDefaultConfig() error {
 		return err
 	}
 
-	err := config.initConfigPath("")
+	err := config.InitConfigPath("")
 	if err != nil {
 		return err
 	}
@@ -184,7 +184,7 @@ func (config *MixConfig) SetProperty(propertyStr string, value string) error {
 // LoadConfig loads a configuration file from a provided path or from local directory
 // is none is provided
 func (config *MixConfig) LoadConfig(filename string) error {
-	if err := config.initConfigPath(filename); err != nil {
+	if err := config.InitConfigPath(filename); err != nil {
 		return err
 	}
 	if err := config.parseVersionAndConvert(); err != nil {
@@ -271,7 +271,7 @@ func (config *MixConfig) validate() error {
 
 // Convert parses an old config file and converts it to TOML format
 func (config *MixConfig) Convert(filename string) error {
-	if err := config.initConfigPath(filename); err != nil {
+	if err := config.InitConfigPath(filename); err != nil {
 		return err
 	}
 
@@ -296,12 +296,14 @@ func (config *MixConfig) Print() error {
 	return nil
 }
 
-func (config *MixConfig) initConfigPath(path string) error {
-	if path != "" {
-		config.filename = path
+// InitConfigPath sets the main config name to what was passed in,
+// or defaults to the current working directory + builder.conf
+func (config *MixConfig) InitConfigPath(fullpath string) error {
+	if fullpath != "" {
+		config.filename = fullpath
 		return nil
 	}
-
+	// Create a builder.conf in the current directory if none is passed in
 	pwd, err := os.Getwd()
 	if err != nil {
 		return err
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -222,6 +222,11 @@ func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allU
 		}
 	}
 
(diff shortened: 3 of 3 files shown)


rchiossi:
I wonder if we could rename `.yum-mix.conf` to `dnf.conf` (or anything else without the initial dot). I dislike the fact that the fole doesn't go away on a `rm -rf *` in the mix directory because it is hidden. I wasted time more than once trying to figure out what was going wrong just to figure out the problem was that I was using an old `.yum-mix.conf`

matthewrsj:
Here's an example: https://github.com/clearlinux/mixer-tools/pull/459/files#diff-58e7b3a6ac3d44b45b15c4d86cf8f3caR36

You don't need to implement a whole `testFilesystem` like in swupd.

tmarcu:
@matthewrsj thanks that works, looks like I will have to chdir into that temp dir for it to work though, as default creation uses the current WD, which is why I thought of the testFilesystem.

builder/builder_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func mustExist(t *testing.T, name string) {
+	t.Helper()
+	_, err := os.Stat(name)
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func cleanup(file string) {
```
matthewrsj: nitpick - you call this on a directory, not a file (ok directories are still files, but still...)

builder/builder_test.go
```diff
@@ -0,0 +1,44 @@
...
+	if err = b.Config.CreateCustomConfig(false, testDir); err != nil {
+		t.Errorf("Failed to create default config: %s", err)
+	}
+
+	if err = b.Config.LoadConfig(testDir + "/builder.conf"); err != nil {
+		t.Errorf("Failed to load default builder.conf: %s", err)
+	}
+	Offline = true
+
+	err = b.InitMix("25550", "10", false, false, true, "https://download.clearlinux.org", false)
```
matthewrsj: since this is an offline unit test can we make it more explicit that these are test values. so 25550 -> 10 and "https://download.clearlinux.org" -> "https://example.com"

builder/builder_test.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		t.Errorf("Failed to initialize mix: %s", err)
+	}
+
+	mustExist(t, testDir+"/.yum-mix.conf")
```
matthewrsj: Does `.yum-mix.conf` exist as a const anywhere you can take advantage of? especially with all the chatter around changing the name of that file.

config/config.go
```diff
@@ -134,6 +134,19 @@ func (config *MixConfig) CreateDefaultConfig(localrpms bool) error {
 	return config.SaveConfig()
 }
 
+// CreateCustomConfig creates a default builder.conf using the pathPrefix
+// directory as base path for the variables values.
+func (config *MixConfig) CreateCustomConfig(localrpms bool, pathPrefix string) error {
```
matthewrsj: This is completely unused? I expect you intended to call it somewhere or expose it via cmdline.

config/config.go
```diff
@@ -301,9 +314,16 @@ func (config *MixConfig) Print() error {
 	return nil
 }
 
-func (config *MixConfig) initConfigPath(path string) error {
-	if path != "" {
-		config.filename = path
+func (config *MixConfig) initConfigPath(pathPrefix string) error {
+	if pathPrefix != "" {
+		// This must be a file, so use the path name directly
```
matthewrsj: instead of overloading `pathPrefix` to *also* mean `fullPath` just remove all this logic and require the argument to actually be the path prefix. If you need both capabilities for some reason add a `fullPath` argument or something.

config/config.go
```diff
@@ -301,9 +314,16 @@ func (config *MixConfig) Print() error {
...
+		if strings.HasSuffix(pathPrefix, ".conf") {
+			config.filename = pathPrefix
+		} else {
+			fmt.Println("writing " + pathPrefix + "/builder.conf")
```
matthewrsj: debug print?

builder/builder_test.go
tmarcu: I don't understand this question.

config/config.go
tmarcu: It's used in the test instead of percolating the changes all over the code base to support a pathPrefix.

config/config.go
tmarcu: It should support passing in just "builder.conf" when you're in the current working directory though, passing a full path in that case would be annoying.

config/config.go
tmarcu: yes :D

config/config.go
matthewrsj: so just make that the default if it is an empty string. Otherwise the argument does two different things based entirely on the suffix of the prefix...

tmarcu: Renamed it to not path prefix :)

config/config.go
```diff
@@ -301,9 +314,15 @@ func (config *MixConfig) Print() error {
...
 
-func (config *MixConfig) initConfigPath(path string) error {
-	if path != "" {
-		config.filename = path
+func (config *MixConfig) initConfigPath(filename string) error {
+	if filename != "" {
+		// This must be a file, so use the path name directly
+		if strings.HasSuffix(filename, ".conf") {
+			config.filename = filename
+		} else {
```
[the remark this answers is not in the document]
matthewrsj: I agree with @rchiossi here, and my other reservations still stand. You may have renamed it from `pathPrefix` to `filename` buy you are *still* overloading the same variable with different purposes.

config/config.go
matthewrsj: You resolved this comment but didn't respond to my suggestion.

tmarcu:
@rchiossi @matthewrsj I haven't found a cleaner way to introduce passing around a path prefix to make it generic without editing nearly the entire code base. It's all hardcoded in many functions to use curr dir and doesn't easily propagate. I'll take another look to see what I can do.

config/config.go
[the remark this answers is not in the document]
tmarcu: What it's called just _name_ ;)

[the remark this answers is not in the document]
matthewrsj: That may address @rchiossi's comment, but not mine.

tmarcu:
@matthewrsj @rchiossi Ok i looked and pretty much everywhere I can find we are passing the full path in every time for the builder.conf so... I don't mind just asking for a fullpath to the file. Simplifies the code a lot.

config/config.go
tmarcu: removed it :)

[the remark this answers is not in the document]
tmarcu: You're right, removed it and using the other func!

builder/builder_test.go
```diff
@@ -0,0 +1,51 @@
...
+func TestNewDNFConfOnInit(t *testing.T) {
+	testDir, err := ioutil.TempDir("", "dnf-testdir-")
+	if err != nil {
+		t.Fatalf("couldn't create temporary directory to write test cases: %s", err)
+	}
+	defer cleanup(testDir)
+	conf := testDir + "/builder.conf"
+
+	b := New()
+	b.Config.LoadDefaultsForPath(false, testDir)
```
phmccarty: The signature of this function is `func (config *MixConfig) LoadDefaultsForPath(path string)`, so I think the lint warnings here do make sense...

tmarcu: Ok rebased and now this is fixed...this was a weird scenario...

PR #478, opened 2018-10-23
builder: Only check online to update upstreamver

### LINKED ISSUE #282: "mixer versions update" fails without a network connection to the upstream server
Incrementing mixversion with "mixer versions update" doesn't work unless there's a network connection.  Even if I add the "--offline" option, it still won't work.

tmarcu:
Updating mixver should not require a network connection, nor reach out
to the server to find out any more information than is needed.

Fixes #282


tmarcu:
A little hard to test until we find some tool that can simulate lack of network, but here's my output (after removing my eth device from my computer & thus my network):
```
[tmarcu]~/mix mixer versions update --mix-version 50
Warning: Using Format=1 from mixer.state for this build.
Old mix:      43
Old upstream: 25750 (format: 1)

New mix:      50
New upstream: 25750 (format: )

Wrote mixversion.
Wrote upstreamversion.

```

rchiossi:
This is how I'm testing offline mode in my working offline branch: https://gist.github.com/rchiossi/6a1cea867fac547e84d3fef44e8f29db

Basically just set an invalid proxy and see if it works.

rchiossi:
This part of `mixer versions` is one of the things that held me back a little. I really want to get that print out of `builder` and move it to `cmd`, but as it is right now it would require passing a lot of info back. Do we really need all this info printed? I wonder if we could just go with the versions after they've been set, so the info would be accessible from `cmd` without any changes.

tmarcu:
@rchiossi I don't necessarily think it should print all that info, at most perhaps just print out the version of the value you changed, i.e just the mixversion. 
And I guess we could just set a bogus proxy to make it fail, you're right. I'm gonna check it out :D

tmarcu:
@matthewrsj rebased to master with new builder changes

bat/tests/09-update-mixver-offline/description.txt
```diff
@@ -0,0 +1,4 @@
+02-update-mixver-offline
```
matthewrsj: copypasta! s/02/09/

bat/tests/09-update-mixver-offline/run.bats
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bats
+
+# shared test functions
+load ../../lib/mixerlib
+
+setup() {
+  global_setup
+}
+
+@test "Create mix 10 with real clear bundle" {
```
matthewrsj: update test description.

bat/tests/09-update-mixver-offline/run.bats
```diff
@@ -0,0 +1,17 @@
...
+  mixer-init-stripped-down 25770 10
```
matthewrsj: should you export the proxies before this, and make this completely offline?

builder/metadata.go
```diff
@@ -172,9 +172,20 @@ Latest upstream in format: %d
 // upstream version is 0, then the latest upstream version in the current
 // upstream format will be taken instead.
 func (b *Builder) UpdateVersions(nextMix, nextUpstream uint32) error {
-	format, _, latest, err := b.getUpstreamFormatRange(b.UpstreamVer)
-	if err != nil {
-		return err
+	var format string
+	var nextFormat string
```
matthewrsj: can you please put this closer to where it is actually used on line 202? Unless you *like* C89 variable declaration requirements...

builder/metadata.go
```diff
@@ -187,18 +198,20 @@ func (b *Builder) UpdateVersions(nextMix, nextUpstream uint32) error {
 
 	nextUpstreamStr := strconv.FormatUint(uint64(nextUpstream), 10)
 
-	nextFormat := format
-	if nextUpstream > latest {
-		nextFormat, err = b.getUpstreamFormat(nextUpstreamStr)
-		if err != nil {
-			return err
+	if checkUpstream {
+		nextFormat = format
```
matthewrsj: `nextFormat` is only set if you are going to `checkUpstream` so it will be printed (on line 217 for now) as its zero-value. You need to initialize it to the current format.

builder/metadata.go
tmarcu: I mean, this could also be C99 style ;)

bat/tests/09-update-mixver-offline/run.bats
tmarcu: It's a commonly used function, I don't want to edit it just to move the proxies above by 1 line, I think this works just fine.

builder/metadata.go
tmarcu: It has to happen on 200 because it's used out of that scope, moved it though.

builder/metadata.go
tmarcu: That's invalid though, the right answer is to not print it at all because the nextFormat is not guaranteed to be curr +1.

tmarcu: Updated to check if it's empty.

matthewrsj: No, next format is guaranteed to be current format if you aren't changing your upstream, why not display that? You are still printing what the `nextUpstream` is even if it didn't change but for some reason you are deciding not to print what the format of that upstream is? So I'll say again (feel free to give me a thumbs down again if you wish): you should initialize it to the current format and always print.

builder/metadata.go
```diff
@@ -187,26 +197,32 @@ func (b *Builder) UpdateVersions(nextMix, nextUpstream uint32) error {
...
+		}
 	}
 
 	fmt.Printf(`Old mix:      %d
 Old upstream: %d (format: %s)
 
-New mix:      %d
-New upstream: %d (format: %s)
-`, b.MixVerUint32, b.UpstreamVerUint32, format, nextMix, nextUpstream, nextFormat)
+New mix:      %d\n
```
matthewrsj: Why did you decide to add an extra newline here?

This should really be a template instead of a `Printf` like this, but I guess that's out of scope for what you are trying to do.

bat/tests/09-update-mixver-offline/run.bats
matthewrsj: Then we should have another explicit bat test making sure the offline capabilities work. In that case we will need to either modify that function or provide a new one.

builder/metadata.go
tmarcu: It doesn't make sense to me to say the next format = my current format, that's why there's 2 format variables to print here. If I can't check for format and my curr format is 25, my next format is not 25, it's (likely) 26, but who knows, it could be 30. I would rather not print information than print misleading info.

builder/metadata.go
tmarcu: It's the same output as before, the original was 
```
`
%s
`
```

which is the same as
`"%s\n"` because of the backticks.
yeah more templates more better at some point.

builder/metadata.go
matthewrsj: but you are just printing what the format the next upstream version is. If the version itself hasn't changed why would the upstream format change?

bat/tests/09-update-mixver-offline/run.bats
tmarcu: This isn't that test nor the scope/requirement of this test to introduce though... I can add a new function but it's starting to add tests that should be opened in another PR.

builder/metadata.go
tmarcu: Ok I think we were not on the same page, I think I've implemented what you're stating above now ^

builder/metadata.go
```diff
@@ -172,9 +172,19 @@ Latest upstream in format: %d
 // upstream version is 0, then the latest upstream version in the current
 // upstream format will be taken instead.
 func (b *Builder) UpdateVersions(nextMix, nextUpstream uint32) error {
-	format, _, latest, err := b.getUpstreamFormatRange(b.UpstreamVer)
-	if err != nil {
-		return err
+	var format string
+	var latest uint32
+	var err error
+	// Only check online if an upstream version was passed in
```
phmccarty: Can you update this comment to reflect what the code is actually doing? Or remove the comment, since the code below seems pretty understandable.

tmarcu: sure I think we can remove this

PR #480, opened 2018-10-23
battest: Get format boundary from latest

rchiossi:
Instead of getting first version of the current format and subtracting,
use the previous format and get the latest version directly.

### DIFF
--- a/bat/lib/mixerlib.bash
+++ b/bat/lib/mixerlib.bash
@@ -99,9 +99,7 @@ download-rpm() {
 get-last-format-boundary() {
   latest=$(curl https://download.clearlinux.org/latest)
   format=$(curl https://download.clearlinux.org/update/$latest/format)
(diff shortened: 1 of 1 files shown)


bat/lib/mixerlib.bash
```diff
@@ -99,9 +99,7 @@ download-rpm() {
 get-last-format-boundary() {
   latest=$(curl https://download.clearlinux.org/latest)
   format=$(curl https://download.clearlinux.org/update/$latest/format)
-  first=$(curl https://download.clearlinux.org/update/version/format$format/first)
-
-  echo $(($first-10))
+  echo $(curl https://download.clearlinux.org/update/version/format$(($format-1))/latest)
```
matthewrsj: I guess this is ok since it is getting it directly from upstream. If format was a default `1` this wouldn't work of course.

tmarcu:
Fixing upstream, and mixer should start creating this file.

PR #69, opened 2018-01-04
Improve test helpers for swupd library

matthewrsj:
* Add CreateManifests test helpers to helpers_test.go
Add several test helpers to helpers_test.go to aid in generation of test
directories and other inputs. Convert existing tests to use these helper
functions. Move existing helper functions to helpers_test.go and remove
unused helpers.

* Remove unused CreateManifests test input directories.
The previous commit removed the need for these manually created test
directories.

* Move remaining test helpers to helpers_test.go
Consolidate remaining test helper functions in helpers_test.go

PR #484, opened 2018-10-25
Builder tests and improvements based on the tests

matthewrsj:
* Start to add bundle_control unit tests 
Start to add some unit tests for the bundle_control file. There is much
work to still be done as far as testing goes for these methods.

* builder: read upstream-bundles from mix workspace
Instead of requiring upstream bundles be read from a child directory
under where mixer is currently run read them from underneath the mixer
workspace (builder.Config.Mixer.VersionPath) the same way local-bundles
is handled.

* builder: improve isLocalBundle checks based on testing
Some testcases showed these checks were incomplete and not handling
corner cases. Improve the logic of these checks to comply with the new
tests.

### DIFF
--- a/builder/bundle_control_test.go
+++ b/builder/bundle_control_test.go
@@ -1,0 +1,432 @@
+package builder
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+
+	"github.com/clearlinux/mixer-tools/config"
+)
+
+func TestGetUpstreamBundlesVerDir(t *testing.T) {
+	testCases := []struct {
+		ver string
+		exp string
+	}{
+		{"0", "clr-bundles-0"},
+		{"000001", "clr-bundles-000001"},
+		{"non-numeric", "clr-bundles-non-numeric"},
+		{"", "clr-bundles-"},
+		{"25660", "clr-bundles-25660"},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.ver, func(t *testing.T) {
+			actual := getUpstreamBundlesVerDir(tc.ver)
+			if actual != tc.exp {
+				t.Errorf("expected %s on input %s but got %s", tc.exp, tc.ver, actual)
+			}
+		})
+	}
+}
+
+func TestGetUpstreamBundlesPath(t *testing.T) {
+	testCases := []struct {
+		ver string
+		exp string
+	}{
+		{"0", "test/upstream-bundles/clr-bundles-0/bundles"},
+		{"000001", "test/upstream-bundles/clr-bundles-000001/bundles"},
+		{"non-numeric", "test/upstream-bundles/clr-bundles-non-numeric/bundles"},
+		{"", "test/upstream-bundles/clr-bundles-/bundles"},
+		{"25660", "test/upstream-bundles/clr-bundles-25660/bundles"},
+	}
+	b := New()
+	b.Config = config.MixConfig{}
+	b.Config.LoadDefaultsForPath("test")
+	b.Config.Builder.VersionPath = "test"
+	for _, tc := range testCases {
+		t.Run(tc.ver, func(t *testing.T) {
+			actual := b.getUpstreamBundlesPath(tc.ver)
+			if actual != tc.exp {
+				t.Errorf("expected %s on input %s but got %s", tc.exp, tc.ver, actual)
+			}
+		})
+	}
+}
+
+func TestGetLocalPackagesPath(t *testing.T) {
+	b := New()
+	// internally this function requires some sub-objects of builder to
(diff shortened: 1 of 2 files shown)


builder/bundle_control_test.go
```diff
@@ -0,0 +1,432 @@
...
+		t.Run(tc.ver, func(t *testing.T) {
+			actual := getUpstreamBundlesVerDir(tc.ver)
+			if actual != tc.exp {
+				t.Errorf("expected %s on input %s but got %s", tc.exp, tc.ver, actual)
+			}
+		})
+	}
+}
+
+func TestGetUpstreamBundlesPath(t *testing.T) {
```
tmarcu: Hmm I like that this increases coverage, but this is actually just testing if `LoadDefaultsForPath()` and `filepath.Join()` works, & the only test case that would actually succeed in a real run is the 25660 one. The rest will fail to download bundles (granted later on in a real mixer run), so I think it's a little over-engineered :) if it writes a good version w/ a good prefix, I think we can safely say it works.

builder/bundle_control_test.go
```diff
@@ -0,0 +1,432 @@
...
+		{"25660", "upstream-bundles/clr-bundles-25660/packages"},
+	}
+
+	b := New()
+	for _, tc := range testCases {
+		t.Run(tc.ver, func(t *testing.T) {
+			b.UpstreamVer = tc.ver
+			actual := b.getUpstreamPackagesPath()
+			if actual != tc.exp {
+				t.Errorf("expected %s but got %s", tc.exp, actual)
```
tmarcu: i.e I'd say doing for this what you did on line #69 is totally enough IMHO.

builder/bundle_control_test.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+}
+
+func TestGetUpstreamBundles(t *testing.T) {
+	b := New()
+	Offline = true
+	if b.getUpstreamBundles("", false) != nil {
+		t.Error("returned error when in offline mode")
+	}
+	// TODO: mock network
```
tmarcu: Maybe we can mock this with setting the env at beginning of test with http_proxy=0.0.0.0 like we do in BATS, it should set it for the whole run and just work maybe? Since we don't care about slow network tests (right now) just lack of network.

matthewrsj: No there are actual mock packages out there that mock network responses.

That also doesn't "mock network". The TODO is so we can actually test the rest of `getUpstreamBundles`. It already tests offline mode.

builder/bundle_control_test.go
matthewrsj: Yes, right now it only tests `filepath.Join` because `getUpstreamBundlesPath` is just a wrapper for that function. This is a unit test - part of the value of a unit test is to future-proof any changes we make to that function. We are now testing what we expect from that function and if we change that function that causes it to deviate from our expectations we will be alerted.

Also, what does it cost us to have this unit test? We gain value (validate future changes) and it costs us nothing.

builder/bundle_control_test.go
matthewrsj: I think if either test should be modified it should be the one on L69 to provide more cases. I don't see the value in *taking away* test cases that have already been written.

tmarcu: _ok then_
I just want to reiterate having _useful_ test cases instead of just extra lines of code going forward.

builder/bundle_control_test.go
```diff
@@ -0,0 +1,432 @@
...
+	}{
+		{"0", "test/upstream-bundles/clr-bundles-0/bundles"},
+		{"000001", "test/upstream-bundles/clr-bundles-000001/bundles"},
+		{"non-numeric", "test/upstream-bundles/clr-bundles-non-numeric/bundles"},
+		{"", "test/upstream-bundles/clr-bundles-/bundles"},
+		{"25660", "test/upstream-bundles/clr-bundles-25660/bundles"},
+	}
+	b := New()
+	b.Config = config.MixConfig{}
+	b.Config.LoadDefaultsForPath(false, "test")
```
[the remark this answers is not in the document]
matthewrsj: fixed and new version pushed.

builder/bundle_control_test.go
matthewrsj: Well I'd argue that all of those test-cases are useful. This is the difference between *white-box testing* and *black-box testing*. When black-box testing we can't assume we know the internals of the function at all *or* that the internals will remain the same. Those test cases are essentially black-box tests to make sure we are always testing the expected result even if the internals of the function changes. Black box tests are not "extra lines of code" and are valuable even if you personally know the internal workings of the functions under test.

tmarcu:
@phmccarty bump

Issue #275, opened 2018-05-04
Eliminate the "--local-rpms" options with "mixer init" and just automatically create the local-rpms and local-yum directories


bktan8:
Why not just automatically create the local-rpms and local-yum directories and add their paths into the builder.conf file without the user having to specify the "--local-rpms" option as part of the "mixer init"?

If the user misses the "--local-rpms" options at the beginning, s/he has to manually create the these "local" workspaces and modify the builder.conf which can lead to errors if not modified correctly.

tmarcu:
Sure!

bktan8:
awesome!

PR #486, opened 2018-10-26
init: Always set localrpms

rchiossi:
The existence of local repo and local rpms folder has no impact if they
are not used in the mix. This patch removes the `--local-rpms` flag for
`mixer init` and make it always true. This way both folders are always
created and configured in builder.conf during init.

This patch also update the test cases for config convert to address this
change since.

### DIFF
--- a/builder/init.go
+++ b/builder/init.go
@@ -36,17 +36,13 @@ func (b *Builder) getLatestUpstreamVersion() (string, error) {
 // initDirs creates the directories mixer uses
 func (b *Builder) initDirs() error {
 	// Create folder to store local rpms if defined but doesn't already exist
-	if b.Config.Mixer.LocalRPMDir != "" {
-		if err := os.MkdirAll(b.Config.Mixer.LocalRPMDir, 0777); err != nil {
-			return errors.Wrap(err, "Failed to create local rpms directory")
-		}
+	if err := os.MkdirAll(b.Config.Mixer.LocalRPMDir, 0777); err != nil {
+		return errors.Wrap(err, "Failed to create local rpms directory")
 	}
 
 	// Create folder for local dnf repo if defined but doesn't already exist
-	if b.Config.Mixer.LocalRepoDir != "" {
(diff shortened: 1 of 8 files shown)


rchiossi:
This PR addresses issue #275

builder/init.go
```diff
@@ -36,17 +36,13 @@ func (b *Builder) getLatestUpstreamVersion() (string, error) {
 // initDirs creates the directories mixer uses
 func (b *Builder) initDirs() error {
 	// Create folder to store local rpms if defined but doesn't already exist
-	if b.Config.Mixer.LocalRPMDir != "" {
-		if err := os.MkdirAll(b.Config.Mixer.LocalRPMDir, 0777); err != nil {
-			return errors.Wrap(err, "Failed to create local rpms directory")
-		}
+	if err := os.MkdirAll(b.Config.Mixer.LocalRPMDir, 0777); err != nil {
```
matthewrsj: I'm not sure I like failing the entire init process if this fails, since before it could be circumvented (or never even seen) if a user did not set the local-rpm conf or pass the flag. Now it is a requirement that it gets made. Not sure how I would resolve that, maybe print a warning and keep going?

rchiossi: Since the path is fixed on init and points to $pwd, this would also means that the creation of `local-bundles` and `upstream-bundles` folders would also fail (not to mention all the other artifacts created on $pwd during init).

matthewrsj: so hard-fail on `local-bundles` since that is the only thing that is technically necessary for a mix right?

rchiossi: not really, you can have a mix with an empty `local-bundles`, so I assume you don't even need that folder. Still, you do need the other artifacts that are created on pwd (upstreamurl, upstreamversion, mixversion and mixer.state) so if if fails to create a folder it will also fail to create those files.

matthewrsj: We need to stop creating those files based on pwd and put them under the `VersionPath`

matthewrsj: Anyways, I don't see the hard failure as a big deal, I just wanted to bring it up.

PR #487, opened 2018-10-26
Bat Improvements

rchiossi:
Small improvements to battest. See commit message for details.

### DIFF
--- a/bat/lib/mixerlib.bash
+++ b/bat/lib/mixerlib.bash
@@ -15,11 +15,8 @@ global_setup() {
 }
 
 localize_builder_conf() {
-  if $(mixer $MIXARGS config set Mixer.LOCAL_RPM_DIR $BATS_TEST_DIRNAME/local-rpms --native=true); then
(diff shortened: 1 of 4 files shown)


matthewrsj:
in second commit message, `s/rellying/relying`

bat/lib/mixerlib.bash
```diff
@@ -86,19 +86,58 @@ mixer-bundle-remove() {
...
+}
+
+get-last-format-boundary() {
+  format=$(get-current-format)
   first=$(curl https://download.clearlinux.org/update/version/format$format/first)
 
   echo $(($first-10))
 }
 
+setup-dnf() {
```
matthewrsj: this gets called for every time you `download_rpm`. Can you put this in a shared location and check for existence so you don't have to write the same file again and again? It doesn't change.

rchiossi: You mean something like create a `bat/config` folder and keep it there instead of creating it all the time? It sounds better indeed.

matthewrsj: Yep!

rchiossi:
I pushed again because I forgot to fix the typo. =(

PR #488, opened 2018-10-26
Cli Rework

rchiossi:
This might not look like, but it is part of my offline patch. It was so complicated to track all code paths that I end up cleaning up the root code path on the cli module.

This PR includes a couple cleanups and refactor in `cmd` as well as a major rework of root's preRun function.

### DIFF
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -17,6 +17,7 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -29,6 +30,8 @@ import (
 
 	"github.com/spf13/cobra"
 )
+
+const bumpMarker = "format-bump"
 
 type buildCmdFlags struct {
 	format        string
@@ -72,6 +75,33 @@ func setWorkers(b *builder.Builder) {
 var buildCmd = &cobra.Command{
 	Use:   "build",
 	Short: "Build various pieces of OS content",
+
+	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		// Since this function overwrites root's PersistentPreRun, call it manually
+		root := cmd
+		for root.Parent() != nil {
+			root = root.Parent()
+		}
+		if err := root.PersistentPreRunE(cmd, args); err != nil {
+			return err
+		}
+
+		b, err := builder.NewFromConfig(configFile)
+		if err != nil {
+			return err
+		}
+
+		if err := printFormatMismatch(b); err != nil {
+			return err
+		}
+
+		// Skip format bump check for format-bump commands
+		if hasMarker(cmd, bumpMarker) {
+			return nil
+		}
+
+		return checkFormatBump(b)
+	},
 }
 
 func buildBundles(builder *builder.Builder, signflag, cleanFlag bool) error {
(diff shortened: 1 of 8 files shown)


matthewrsj:
in "cmd: Break root preRun into smaller parts" `s/threated/treated/`

matthewrsj:
Also in that commit message:
- conteiner -> container
- executtion -> execution
- excusive -> exclusive

mixer/cmd/build.go
```diff
@@ -71,6 +73,33 @@ func setWorkers(b *builder.Builder) {
...
+		if err != nil {
+			return err
+		}
+
+		if err := checkFormatMismatch(b); err != nil {
+			return err
+		}
+
+		// Skip format bump check for format-bump commands
+		if marker := cmd.Annotations[bumpMarker]; marker != "" {
```
matthewrsj: Do you need to actually get the value or can you just check if the key exists?
```go
if _, ok := cmd.Annotations[bumpMarker]; ok {
```

mixer/cmd/build.go
```diff
@@ -508,23 +537,45 @@ func setUpdateFlags(cmd *cobra.Command) {
 	_ = cmd.Flags().MarkDeprecated("keep-chroots", "this flag is ignored by the update builder")
 }
 
-var buildCmds = []*cobra.Command{
+var containerCmds = []*cobra.Command{
```
matthewrsj: ah very nice way to split this up.

mixer/cmd/build.go
```diff
@@ -508,23 +537,45 @@ func setUpdateFlags(cmd *cobra.Command) {
...
+	buildFormatOldCmd,
 }
 
 func init() {
-	for _, cmd := range buildCmds {
+	for _, cmd := range containerCmds {
+		if cmd.Annotations == nil {
+			cmd.Annotations = map[string]string{}
+		}
+		cmd.Annotations[containerMarker] = "true"
```
matthewrsj: I think for your use-case here you can just set it to the empty string and check for key existence instead. I don't like default strings like this.

mixer/cmd/build.go
```diff
@@ -508,23 +537,45 @@ func setUpdateFlags(cmd *cobra.Command) {
...
+	for _, cmd := range nativeCmds {
+		buildCmd.AddCommand(cmd)
+	}
+
+	for _, cmd := range bumpCmds {
+		if cmd.Annotations == nil {
+			cmd.Annotations = map[string]string{}
+		}
+		cmd.Annotations[containerMarker] = "true"
+		cmd.Annotations[bumpMarker] = "true"
```
matthewrsj: same comment about these.

mixer/cmd/build.go
```diff
@@ -568,3 +619,39 @@ func init() {
...
+
+	hostFormat, upstreamFormat, err := b.GetHostAndUpstreamFormats()
+	if err != nil {
+		return err
+	}
+
+	if hostFormat == "" {
+		fmt.Println("Warning: Unable to determine host format. Running natively may fail.")
+	} else if hostFormat != upstreamFormat {
+		fmt.Println("Warning: The host format and mix upstream format do not match.",
```
matthewrsj: We should `log` these so they go to `stderr` like other warnings.

mixer/cmd/build.go
```diff
@@ -568,3 +619,39 @@ func init() {
...
+			"Mixer may be incompatible with this format; running natively may fail.")
+	}
+
+	return nil
+}
+
+func checkFormatBump(b *builder.Builder) error {
```
matthewrsj: so, `checkFormatMismatch` prints warnings and returns `nil` but this function doesn't print anything and returns `err`s. Any reason for the inconsistency?

mixer/cmd/root.go
```diff
@@ -33,12 +33,15 @@ import (
...
 
+const containerMarker = "container"
+
 // RootCmd represents the base command when called without any subcommands
 var RootCmd = &cobra.Command{
 	Use:  "mixer",
 	Long: `Mixer is a tool used to compose OS update content and images.`,
 
 	PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+		//CPU Profiling
```
matthewrsj: For consistency with the rest of the codebase can you add a space after the `//` please?

mixer/cmd/root.go
```diff
@@ -59,85 +62,16 @@ var RootCmd = &cobra.Command{
...
-			if !builder.Native {
-				if err := b.RunCommandInContainer(reconstructCommand(cmd, args)); err != nil {
-					fail(err)
-				}
-				// Cancel native run and return
-				cancelRun(cmd)
-				return nil
-			}
+		// Execute command in container if needed
+		if useContainer := cmd.Annotations[containerMarker]; useContainer != "" && !builder.Native {
```
matthewrsj:
```go
if _, ok := cmd.Annotations[containerMarker]; ok && !builder.Native {
```

mixer/cmd/build.go
rchiossi: yep, that works! I'll update.

mixer/cmd/build.go
rchiossi: Indeed! I wrote this code before the log change, I let this one slip. Good catch!

mixer/cmd/build.go
rchiossi: Those code chunks where extracted from the older root preRun(). They have similar naming, but they don't perform the same function. The format check main purpose is to inform the user, it only fails on `GetHostAndUpstreamFormats()` because this is an early way to stop the build if there is no connection on offline was not used. We could ignore the error and continue.
On the other hand, the `checkFormatBump()` is an actual check. If the check fails, the execution should stop.
Maybe I could rename `checkFormatMismatch()` to `informFormatMismatch()` if that would make it less confusing.

tmarcu: I think either names work, if you want more options `PrintFormatMismatch` works too IMO.

mixer/cmd/init.go
```diff
@@ -0,0 +1,89 @@
...
+	upstreamURL string
+	git         bool
+	format      string
+}
+
+var initFlags initCmdFlags
+
+var initCmd = &cobra.Command{
+	Use:   "init",
+	Short: "Initialize the mixer and workspace",
```
tmarcu: > "Initialize the mixer and workspace",

This would make more sense saying just "Initialize the mixer workspace" or "initialize workspace and confs"
I know this is old & not your wording btw, but this is a great time to update it :)

mixer/cmd/root.go
```diff
@@ -362,6 +211,26 @@ func checkAllDeps() bool {
 	return ok
 }
 
+func containerExecute(cmd *cobra.Command, args []string) error {
+	if builder.Offline {
+		fmt.Println("Warning: Unable to determine upstream format in --offline mode, " +
```
tmarcu: For consistency i'd `_, _ = fmt.Fprintf(os.Stderr, ...)` this one so you don't have to pull in the log package just for here.

rchiossi: Log is already included in build.go, so I used log instead.

mixer/cmd/init.go
rchiossi: done

rchiossi:
Fixed rebase conflicts and added @tmarcu suggestions.

rchiossi:
Updated with @matthewrsj suggestion. I changed `annotateCmd()` to `addMarker()` since annotation has a different meaning in cobra.Command.

PR #490, opened 2018-10-31
bundles: Remove editor option for `bundle edit`

reaganlo:
`bundle edit` will no longer open the editor. It will just create new bundles or copy existing bundles.
This command will locate the bundle (looking first in local-bundles, then in upstream-bundles).
If the bundle is only found upstream, the bundle file will be copied to the local-bundles directory.
If the bundle is not found anywhere, a blank template will be created with the correct name.

This commit also fixes issue #448 and increases test coverage for `bundle validate`.

### DIFF
--- a/builder/bundle_control.go
+++ b/builder/bundle_control.go
@@ -15,12 +15,10 @@
 package builder
 
 import (
-	"bufio"
 	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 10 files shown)


builder/bundle_control.go
```diff
@@ -831,53 +752,35 @@ func createBundleFile(bundle string, path string) error {
...
 
 // EditBundles copies a list of bundles from upstream-bundles to local-bundles
-// (if they are not already there) or creates a blank template if they are new,
-// and launches an editor to edit them. Passing true for 'suppressEditor' will
-// suppress the launching of the editor (and just do the copy or create, if
-// needed), and 'add' will also add the bundles to the mix.
-func (b *Builder) EditBundles(bundles []string, suppressEditor bool, add bool, git bool) error {
+// (if they are not already there) or creates a blank template if they are new.
+// 'add' will also add the bundles to the mix.
+func (b *Builder) EditBundles(bundles []string, add bool, git bool) error {
```
tmarcu: This name doesn't really make sense anymore, since we no longer offer edit capabilities. Maybe we can use a different name, like CopyBundles or something? /me is open to suggestions.

builder/bundle_control.go
```diff
@@ -831,53 +752,35 @@ func createBundleFile(bundle string, path string) error {
...
 	for _, bundle := range bundles {
 		path, _ := b.getBundlePath(bundle)
 		if !b.isLocalBundle(path) {
 			localPath := filepath.Join(b.Config.Mixer.LocalBundleDir, bundle)
 
 			if path == "" {
-				// Bunlde not found upstream, so create new
-				if err = createBundleFile(bundle, localPath); err != nil {
+				// Bundle not found upstream, so create new
+				if err := createBundleFile(bundle, localPath); err != nil {
```
tmarcu: You can just declare `var err error` on line 762 and not have to edit any of these lines below.

mixer/cmd/bundles.go
```diff
@@ -155,31 +155,19 @@ var bundleListCmd = &cobra.Command{
...
-	add      bool
-	git      bool
+	add bool
+	git bool
 }
 
 var bundleEditFlags bundleEditCmdFlags
 
 var bundleEditCmd = &cobra.Command{
 	Use:   "edit <bundle> [<bundle>...]",
```
tmarcu: create <bundle>  sounds better than edit; the use of edit is now not supported so we should use something more meaningful to what actually happens.

mixer/cmd/bundles.go
```diff
@@ -155,31 +155,19 @@ var bundleListCmd = &cobra.Command{
...
-EDITOR. If neither are set, the tool defaults to nano. If nano is not installed,
-the tool will skip editing, and act as if '--suppress-editor' had been passed.
-
-Passing '--suppress-editor' will suppress launching the editor, and will thus
-only copy the bundle file to local-bundles (if it is only found upstream), or
-create the blank template (if it was not found anywhere). This can be useful if
-you want to add a bundle to local-bundles, but wish to edit it at a later time.
+	Short: "Create new bundles or copy existing bundles",
+	Long: `Create new bundles or copy existing bundles. This command will locate the
+bundle (looking first in local-bundles, then in upstream-bundles). If the bundle is only found upstream,
```
tmarcu: > This command will locate the bundle (looking first in local-bundles, then in upstream-bundles).
Sounds better written as:

`This command will locate the bundle by first looking in local-bundles, and then in upstream-bundles.`

mixer/cmd/bundles.go
```diff
@@ -278,7 +266,6 @@ func init() {
 
 	bundleListCmd.Flags().BoolVar(&bundleListFlags.tree, "tree", false, "Pretty-print the list as a tree.")
 
-	bundleEditCmd.Flags().BoolVar(&bundleEditFlags.copyOnly, "suppress-editor", false, "Suppress launching editor (only copy to local-bundles or create template)")
```
tmarcu: Actually we should not remove this right now, can you just make it a no-op and also mark it as hidden, so it doesn't show up in the list anymore, but it also won't crash anyone who already is calling this in a script.

mixer/cmd/bundles.go
reaganlo: @tmarcu Yes, I agree 'edit' does not make sense anymore. 'create' would sound better. I wasn't sure if we should change it right away and maybe could be the next iteration. Also, what are your thoughts on reducing the impact of this PR. E.g. --suppress-editor is removed. Should we have it as deprecated for a while and then remove it completely?

reaganlo:
@phmccarty While trying to fix #448, we realized that there were too many bundle validations happening in mixer; `bundle edit` being one of them. And the validation was nested in the "editor" logic. Also there was logic that created backup of a bundle being edited even when the bundle was incorrect.
Also we felt that it was an overkill for mixer to have this interactive editor feature.

phmccarty:
@reaganlo IMO, removing the interactive editor feature negatively impacts user experience. If a `mixer bundle create` command just copies a blank template (non-interactively), then is the user supposed to know where the bundle definitions live? One reason I like `mixer bundle edit` is because the user does not need to know the location of the bundle definitions being modified. It also freely permits mixer to change the the underlying implementation of bundle management in the future without changing the cli.

tmarcu:
@phmccarty would you like that after copying, it prints out "Created bundle <foo> in $MIX_WORKSPACE/local-bundles, where mixworkspace would be an actual path.

phmccarty:
@tmarcu Yes, I would be okay with that. I still think that interactive editing is better, but I can live with a solution that informs the user about what to do next.

reaganlo:
@phmccarty @tmarcu Added the message to inform user about the bundle creation.

mixer/cmd/bundles.go
rchiossi: Look for `MarkHidden()` and `MarkDeprecated()` in the code for exemples.

reaganlo: @rchiossi Its already been marked hidden and deprecated. You see this line as removed because the flag has been renamed from bundleEditCmd to bundleCreateCmd.

matthewrsj:
Needs a rebase for that test file (we aren't numbering tests anymore)

reaganlo:
@matthewrsj Rebased and added test for `mixer bundle edit --suppress-editor` as suggested.

PR #492, opened 2018-11-02
bundles: Build image with default ister template file (#58)

reaganlo:
When building an image without providing a template, if the default image template
is not present in the mix directory, mixer will do the following:
 - Copy image template from ister template "/usr/share/defaults/ister/release-image-config.json".
 - Update image template with the bundle list from "mixbundles" file.
   In case the "mixbundles" file is empty or not present, use ister template as is.
 - Inform the user about the above steps.
 - Continue the existing build process.


bat/tests/09-ister-config/run.bats
```diff
@@ -0,0 +1,34 @@
...
+  run sudo mixer $MIXARGS build image --native
+  [[ "$status" -eq 1 ]]
+  [[ "$output" =~ "release-image-config.json not found" ]]
+  [[ "$output" =~ "Copying image template" ]]
+  [[ "$output" =~ "Updating image bundle list based on mixbundles" ]]
+}
+
+@test "Update ister config file with mix bundle list" {
+  run cat release-image-config.json
+  [[ "$status" -eq 0 ]]
```
phmccarty: optional: the above two lines can be replaced with `cat release-image-config.json`

bat/tests/09-ister-config/run.bats
```diff
@@ -0,0 +1,34 @@
...
+  [[ "$output" =~ os-core\" ]]
```
phmccarty: is `\"` really needed in the regex?

builder/builder.go
```diff
@@ -26,11 +31,6 @@ import (
 	"path/filepath"
 	"runtime"
 	"sync"
-
-	"github.com/clearlinux/mixer-tools/config"
-	"github.com/clearlinux/mixer-tools/helpers"
-	"github.com/clearlinux/mixer-tools/swupd"
-	"github.com/pkg/errors"
```
phmccarty: Not sure about this reordering of imports... normally, the mixer-tools and third-party package imports are imported like this to visually separate the imports from the standard library. See other mixer-tools source files as a reference.

reaganlo: From what I recollect 'gofmt -w' did this.

bat/tests/09-ister-config/run.bats
reaganlo: Yes. This is done in-order to differentiate between os-core and os-core-update.

bat/tests/09-ister-config/run.bats
reaganlo: I'm executing it with the `run` command so that the output can be accessed using  `$output`

bat/tests/09-ister-config/run.bats
phmccarty: Okay, then I would prefer the regex to be `\"os-core\"`

builder/builder.go
phmccarty: Hmm, interesting, I tried running `gofmt -w builder/builder.go` on the current master branch, but see no modifications to the file...

bat/tests/09-ister-config/run.bats
phmccarty: Okay, thanks for clarifying. I don't have strong opinions either way, but it seems cleaner to me to rewrite this test to use `grep` instead, avoiding any use of `$status` and `$output`. Something like:

```
config=release-image-config
[ -f $config ]
for b in os-core os-core-update kernel-native bootloader editors; do
  grep '"'$b'"' $config
done
```

bat/tests/09-ister-config/run.bats
reaganlo: Fair enough. In fact, I'm taking it a step further and doing it for all the bundles.

builder/builder.go
reaganlo: This happens when you "gofmt -w" after importing "encoding/json" in the empty line [29](https://github.com/clearlinux/mixer-tools/blob/master/builder/builder.go#L29) between "sync" and "github.com/clearlinux/mixer-tools/config"

Anyways, I have changed it as per your comment.

tmarcu: Formatting works in chunks w/no empty lines in between, so if you clump "fmt" and "github.com/.." it will sort them in the same paragraph, but if it's separated with an empty line it will treat paragraphs/chunks of includes as separate and sort them individually.

bat/tests/09-ister-config/run.bats
tmarcu: I'll go 1 step further and say 
`bundles="os-core os-core-update kernel-native bootloader editors"

for b in $bundles; do
  echo '"'$b'"'
done`

is a little clearer too.

PR #496, opened 2018-11-07
build: Add increment flag support to build all

### LINKED ISSUE #445: Mixer build delta-packs won't work after a build all
Running mixer commands in this order will fail:

    $ sudo mixer build all
    $ sudo mixer build delta-packs --previous-versions=1

mixer build all updates mixversion and this information is used by delta-packs to check what's the current version to be used.

If build bundles and build update is used, delta-packs works fine because mixversion isn't updated:

    $ sudo mixer build bundles
    $ sudo mixer build update
    $ sudo mixer build delta-packs --previous-versions=1

jwakre:
Now the 'mixer build all' command supports the --increment flag. When the
--increment flag is passed to 'mixer build all', the mixversion will be
incremented. Otherwise, the mixversion will remain unchanged. This fixes
an issue where delta packs could not be created after using 'mixer build all'.

Fixes #445

### DIFF
--- a/bat/tests/09-build-all-delta-packs/description.txt
+++ b/bat/tests/09-build-all-delta-packs/description.txt
@@ -1,0 +1,4 @@
+09-build-all-delta-packs
+==================
+This test case tests the 'mixer build delta-packs' command when generating
+content with 'mixer build all'.
--- a/bat/tests/09-build-all-delta-packs/Makefile
+++ b/bat/tests/09-build-all-delta-packs/Makefile
@@ -1,0 +1,9 @@
+.PHONY: check clean
+
+check:
+	bats ./run.bats
+
+CLEANDIRS = ./update ./test-chroot ./logs ./.repos ./bundles ./update ./mix-bundles ./clr-bundles ./local-yum ./results ./repodata ./local-rpms ./upstream-bundles ./local-bundles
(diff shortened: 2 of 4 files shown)


phmccarty:
Hmm, thinking of other options here, would it make sense for `mixer build bundles` or `mixer build all` to perform the mixversion increment before starting the build, instead of afterward?

bat/tests/09-build-all-delta-packs/description.txt
```diff
@@ -0,0 +1,4 @@
+09-build-all-delta-packs
```
reaganlo: You may have to rename this test to "10-build-all-delta-packs" if https://github.com/clearlinux/mixer-tools/pull/492 is merged before this.

jwakre: Yes, I saw a variety of PRs adding tests that would have conflicting version numbers. I think it is easiest to pick the next available number and resolve conflicts when rebasing.

jwakre:
@phmccarty I think that incrementing the mixversion before building bundles rather than when building an update makes sense. I don't know the initial design decisions and I think that this would need some form of deprecation to minimize breakage. @tmarcu Thoughts?

PR #498, opened 2018-11-08
Remove /etc/dnf tree in full chroot

phmccarty:
DNF creates two additional state directories in the full chroot if they missing: `/etc/dnf/` and `/etc/dnf/modules.d`.

Update the DNF state path slice so that the `/etc/dnf` directory tree is removed from the full chroot after full chroot creation. This avoids adding those two directories to Manifest.full.

### DIFF
--- a/bat/tests/12-no-state-variables-in-full/run.bats
+++ b/bat/tests/12-no-state-variables-in-full/run.bats
@@ -1,0 +1,21 @@
+#!/usr/bin/env bats
+
+# shared test functions
+load ../../lib/mixerlib
+
+setup() {
+  global_setup
+}
+
+@test "Test that state paths do not exist in the full chroot" {
+  upstreamver=$(get-last-format-boundary)
+  mixer-init-stripped-down $upstreamver 10
+
+  #Create initial version in the old format
+  mixer-build-bundles > $LOGDIR/build_bundles.log
+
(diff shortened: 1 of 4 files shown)


matthewrsj:
@tmarcu should this have a related test? I could probably write one tomorrow if @phmccarty doesn't have the bandwidth.

phmccarty:
@matthewrsj I would appreciate the help with adding a test... not sure when I will have a chance to work on that.

tmarcu:
Yes, this is something that should have a test. I don't mind helping out, I think this is a simple ish test to put out.

tmarcu:
@phmccarty  Please confirm this is what you want to see in the test and and if not, make any appopriate changes. Testing w/o this patch fails the test, testing with this patch passes.

bat/tests/12-no-state-variables-in-full/run.bats
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bats
+
+# shared test functions
+load ../../lib/mixerlib
+
+setup() {
+  global_setup
+}
+
+# TODO: This test should run 'mixer build upstream-format' instead
```
matthewrsj: take out this copypasta comment

phmccarty:
@tmarcu +1, thanks

bat/tests/12-no-state-variables-in-full/run.bats
tmarcu: thx

PR #502, opened 2018-11-14
bat: renumber conflicting tests

matthewrsj:
We had three 09-* tests, renumber them so we can prove we actually know
how to count when people look at our test directory.

EDIT: just remove all the numbers, they were silly.


tmarcu:
merging after travis ty again

PR #504, opened 2018-11-15
mixin: fix MoM verification with openssl 1.1

phmccarty:
The "-purpose any" option is required now under openssl 1.1.

NOTE: needs testing

### DIFF
--- a/mixin/build.go
+++ b/mixin/build.go
@@ -206,7 +206,7 @@ func buildMix(prepNeeded bool) error {
 	cert := "/usr/share/ca-certs/Swupd_Root.pem"
 
 	err = helpers.RunCommandSilent("openssl", "smime", "-verify", "-in", "Manifest.MoM.sig",
-		"-inform", "der", "-content", "Manifest.MoM", "-CAfile", cert)
+		"-inform", "der", "-content", "Manifest.MoM", "-purpose", "any", "-CAfile", cert)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rchiossi:
Do you have any details on why this works? From what I could understand from documentation is that if `-purpose` is not provided, this check is skipped during verification, so it seems weird that you need it for openssl to work.

phmccarty:
@rchiossi This is a compatibility fix for openssl 1.1.x. The verification worked without the option in openssl 1.0.x. swupd-client needed a compat fix too: https://github.com/clearlinux/swupd-client/pull/682

As to "why", no idea. Because the "any" purpose is the required value for validating the current Swupd_Root.pem, it's possible openssl 1.0.x inferred the purpose, whereas 1.1.x does not.

phmccarty:
@rchiossi Were you reading the documentation from https://www.openssl.org/docs/manmaster/man1/verify.html ?

A couple of observations: the "any" purpose isn't documented on that page, so it may require different handling than the listed purposes. Also, the Swupd_Root.pem is self-signed, not part of a cert chain, so verify may treat the purpose differently than what is documented there.

rchiossi:
Yes, this was my reference. Indeed it is not mentioned, so they might have changed the default from skip to some of the options in 1.1.
Note that I just wanted to understand what happened, I'm not against this PR if it solves the issue.

PR #505, opened 2018-11-16
Convert to go modules for dependency management

matthewrsj:
Go modules are the official implementation of dependency management in
go so convert from go dep to go mod.

This means we can develop mixer outside of any gopath so we no longer
need local gopath setup in the Makefile.


matthewrsj:
Need to leave vendor dir

PR #506, opened 2018-11-16
Convert to using go modules for dependency management

matthewrsj:
This means we can remove local gopath preparation from the Makefile and
will be using the officially sanctioned dependency management system for
Go. We have to leave the vendor directory so we can create the dist
tarball.

Remove required package argument to 'make checkcoverage'. We are on Go 1.11 now and -coverprofile supports multiple packages.

Update travis environment to run go 1.11


matthewrsj:
I considered squashing these commits together but I actually like how they are separated out right now.

Makefile
```diff
@@ -88,11 +62,7 @@ lint: gopath
 	./...
 
 clean:
-ifeq (,${LOCAL_GOPATH})
-	go clean -i -x
-else
-	rm -rf ${LOCAL_GOPATH}
-endif
+	go clean -i -x ./...
```
rchiossi: Just curious about why you need `./...` here

matthewrsj: Because without it `go clean` errors out with
```
can't load package: package github.com/clearlinux/mixer-tools: unknown import path "github.com/clearlinux/mixer-tools": cannot find module providing package github.com/clearlinux/mixer-tools
```
since mixer-tools itself doesn't provide a go package (its subdirectories do)

Issue #489, opened 2018-10-30
`mixer bundle add` does not add valid bundle if invalid bundle is also provided


reaganlo:
`mixer bundle add <valid bundle> <invalid bundle>` does not even add the `<valid bundle>` to the mix.

E.g.
$`mixer bundle add foobar foo.bar`
_Adding bundle "foobar" from local bundles
2018/10/30 19:43:01 ERROR: Invalid bundle name "foo.bar" derived from file "/home/reaganlo/mix/local-bundles/foo.bar"_

$`cat mixbundles`
_bootloader
kernel-native
os-core
os-core-update_

Expected Output:
Ideally, one would expect that the valid bundle be added and only the invalid bundle be skipped.

reaganlo:
Fixed and merged as part of PR https://github.com/clearlinux/mixer-tools/pull/507

PR #507, opened 2018-11-21
bundles: Add valid bundles and skip invalid ones

reaganlo:
`bundle add` will now add valid bundles and skip only the invalid ones,
thereby fixing issue #489.

### DIFF
--- a/bat/tests/bundle-commands/run.bats
+++ b/bat/tests/bundle-commands/run.bats
@@ -96,6 +96,20 @@ setup() {
   rm -f local-bundles/foocar
 }
 
+@test "Skip invalid bundles from mix" {
+  mixer $MIXARGS bundle create foocar foo.car foojar --add      # Create and add valid as well as invalid bundles
+
+  mixer $MIXARGS bundle list | grep -q foocar     # "foocar" bundle is in the mix
+  mixer $MIXARGS bundle list | grep -q foojar     # "foojar" bundle is in the mix
+
+  run $(mixer $MIXARGS bundle list | grep foo.car)              # "foo.car" is an invalid bundle and is not in the mix
+  [[ ${#lines[@]} -eq 0 ]]
+
+  # Delete bundle from local-bundles and mix (test case clean up)
+  mixer $MIXARGS bundle remove foocar foojar
+  rm -f local-bundles/foocar local-bundles/foojar local-bundles/foo.car
+}
+
 @test "Validate a bundle" {
   echo "package" >> $BATS_TEST_DIRNAME/local-bundles/foobar
   mixer $MIXARGS bundle create foo.bar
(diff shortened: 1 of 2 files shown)


bat/tests/bundle-commands/run.bats
```diff
@@ -96,6 +96,20 @@ setup() {
   rm -f local-bundles/foocar
 }
 
+@test "Skip invalid bundles from mix" {
+  mixer $MIXARGS bundle create foocar foo.car foojar --add      # Create and add valid as well as invalid bundles
+
+  mixer $MIXARGS bundle list | grep -q foocar     # "foocar" bundle is in the mix
+  mixer $MIXARGS bundle list | grep -q foojar     # "foojar" bundle is in the mix
```
matthewrsj: you also need to validate that `foo.car` is *not* in the mix.

bat/tests/bundle-commands/run.bats
```diff
@@ -96,6 +96,20 @@ setup() {
...
+
+  run $(mixer $MIXARGS bundle list | grep foo.car)              # "foo.car" is an invalid bundle and is not in the mix
+  [[ ${#lines[@]} -eq 0 ]]
```
reaganlo: @matthewrsj I have it here.

matthewrsj: ha, so you do.

matthewrsj:
Not sure if that is the right way to clean up after your test case, pinging @rchiossi

builder/bundle_control.go
```diff
@@ -375,10 +375,12 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool,
 
 		bundle, err := b.getBundleFromName(bName)
 		if err != nil {
-			return err
+			fmt.Println(err.Error() + "; skipping")
```
rchiossi: This looks more like a warning that a normal message, so I believe it would be better to use the log API. Something like `log.Println("Warning: " + err.Error() + "; skipping" )`.

builder/bundle_control.go
```diff
@@ -375,10 +375,12 @@ func (b *Builder) AddBundles(bundles []string, allLocal bool, allUpstream bool,
...
+			continue
 		}
 		if err = validateBundleName(bName); err != nil {
-			return err
+			fmt.Println(err.Error() + "; skipping")
```
rchiossi: Same as above

phmccarty:
@rchiossi Github shows diffs for force pushes now (see the "force-pushed" links in the PR), so we don't lose context for reviewing anymore :)

builder/bundle_control.go
reaganlo: @rchiossi Done.

PR #510, opened 2018-11-28
mixin: prevent removal of clear repo

### LINKED ISSUE #326: mixin cannot add package from repo if the "clear" repo is removed
If you remove the default "clear" repo, "mixin package add" fails to find and add packages in a private repo.

Steps to repo:
1) mkdir -p /home/user/mixin-repo
2) cp helloworldb* /home/user/mixin-repo
3) cd /home/user/mixin-repo
4) sudo createrepo_c .
5) sudo mixin repo add myrepo file:///home/user/mixin-repo/
6) sudo mixin repo remove clear
7) sudo -E mixin package add helloworldb

Should get:

      Initializing mix version 22600000 from upstream version 22600
      ERROR: no such package

jwakre:
The clear repo is mandatory for mixin. This change prevents mixin from
removing the clear repo.

Fixes #326

### DIFF
--- a/bat/tests/mixin-repo-commands/run.bats
+++ b/bat/tests/mixin-repo-commands/run.bats
@@ -1,0 +1,18 @@
+#!/usr/bin/env bats
+
+# shared test functions
+load ../../lib/mixerlib
+
+setup() {
+  global_setup
+}
+
+@test "Verify clear repo is unable to be removed by mixin" {
+  run sudo mixin repo remove clear
+
+  [[ "$status" -eq 1 ]]
+  [[ "$output" =~ "The clear repo is mandatory and cannot be removed" ]]
+  [[ ! "$output" =~ "Removed clear repo." ]]
+}
+
+# vi: ft=sh ts=8 sw=2 sts=2 et tw=80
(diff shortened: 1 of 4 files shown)


bat/tests/mixin-repo-commands/run.bats
```diff
@@ -0,0 +1,18 @@
...
+setup() {
+  global_setup
+}
+
+@test "Verify clear repo is unable to be removed by mixin" {
+  run sudo mixin repo remove clear
+
+  [[ "$status" -eq 1 ]]
+  [[ "$output" =~ "The clear repo is mandatory and cannot be removed" ]]
+  [[ ! "$output" =~ "Removed clear repo." ]]
```
rchiossi: I don't think there will ever be a case where the line above passes and this one fails, but on the other hand I see no harm in having them both here, so this is ok.

matthewrsj:
+1 for implementation.

I'm late to the party on this, but I'd rather allow it to be removable. Mixin should work on downstreams of clear as well as on clear itself. As @bktan8 mentioned in #326 "clear" is *just another repo*.

I know @tmarcu disagrees with me here though.

matthewrsj:
`¯\_(ツ)_/¯`

matthewrsj:
`¯\_(ツ)_/¯`

PR #512, opened 2018-12-03
Persistent Offline Mode

rchiossi:
Hey guys,

This patch is an aggregation of all the changes required to get offline mode out. Despite all the patches, the main feature of this PR is that it gives the user the ability to set offline mode during init and mixer will use that value as default for subsequent runs.
Some highlight changes of this patch:
- `Mix.OFFLINE` property was added to `mixer.state` to persist offline mode.
- `mixer.state` version was bumped to 1.1 and automatic conversion was enabled for that file
- `mixer config get` subcommand was added to allow properties to be read via mixer's CLI.
- `mixer config set` now also manipulates `mixer.state`. This allows the persistent value of offline mode to be set via command line - `mixer config set Mix.OFFLINE true`
- `mixer init` now works properly in offline mode
- `mixer versions` now works properly in offline mode

There is a lot of dependency among the commits in this PR but I tried to organize all the code in such a way that we can integrate early sections of it if they are good to go.

The first couple commits are a rework of the old #460 PR. They create an interface shared between MixConfig and MixState to allow code reuse. They enable the `config get/set` changes.

The second part expands the Configuration interface to enable reuse of the versioning code that existed in MixConfig and enables automatic conversion of `mixer.state` file.

The third part enables the Persistent Offline Mode. It will fetch the value from mixer.state early on execution and use it as default for `builder.Offline`. Note that mixer checks if the `--offline` flag was provided by the user. If so, the value of the flag is always used, despite its value.

The last part is an aggregation of fixes to some codepaths that ignored offline. Those were identified by the Offline battest that is included in the very last commit.

Those parts can be integrated separately, but the order must be preserved due to dependencies.


rchiossi:
As a side note to this PR regarding the structure suggested by @matthewrsj in #460, I explored a version where I had a `Configuration` interface, a `TOMLConfig` file which implemented that interface and had all the shared methods and `MixConfig`/`MixState` which had `TOMLConfig` as an embedded structure.

Although this structure might work for "normal" cases, since the shared methods rely on reflection, when I set the their receiver to TOMLConfig I lost visibility on all the properties of `MixConfig` and `MixState`. I tried reassigning the object to a `Configuration` interface variable but it also didn't help. Then I tried to use the reflection API to find the parent of the embedded structure, but at this point the complexity of the shared methods was getting out of control. That is when I decided it was best to just drop this middle class and roll back to the original design.

mixer/cmd/config.go
```diff
@@ -73,6 +73,17 @@ var configSetCmd = &cobra.Command{
 	the existence of the provided property, but will not validate the value provided.`,
 	Args: cobra.ExactArgs(2),
 	Run: func(cmd *cobra.Command, args []string) {
+		// Look for the value in mixer.state first
+		var ms config.MixState
+		err := ms.Load("")
+		if err == nil {
```
tmarcu: if err != nil it sounds like mixer.state doesn't exist, which is an error with mixer now right? I'm trying to re-think this logic a bit, along the lines of
```
	if err != nil {
		fail(err)
	}
	if err = config.SetProperty(&ms, args[0], args[1]) == nil {
		return
	}
```

mixer/cmd/config.go
```diff
@@ -96,11 +98,49 @@ var configSetCmd = &cobra.Command{
...
+	either the value set in the config file or the default value if the property has not
+	been defined in the file.`,
+	Args: cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		var value string
+
+		// Look for the value in mixer.state first
+		var ms config.MixState
+		err := ms.Load("")
+		if err == nil {
```
tmarcu: Same comment as previous, are we sure we want to ignore not being able to load the mixer.state?

mixer/cmd/config.go
```diff
@@ -96,11 +98,49 @@ var configSetCmd = &cobra.Command{
...
+			value, err = config.GetProperty(&ms, args[0])
+			if err == nil {
+				fmt.Println(value)
```
tmarcu: It'd be useful to know this value came from` [mixer.state]:` or `[builder.conf]:`

rchiossi: I believe devops want to use the output of this command for automation, which is much simpler to do if it just returns the value itself.
But I agree that for using it manually it is nice to see where it came from.

mixer/cmd/config.go
tmarcu: Then I don't get the separation between mixer.state and builder.conf.. which one matters? Might as well have a for loop that picks the first available config file in the directory at that point, which doesn't make sense. We chose to have 2 separate confs for a reason right, since it's a failure for either to not exist now.

mixer/cmd/config.go
tmarcu: @George can you confirm this?

george: I think you've got the wrong @george 😉

gtkramer: @tmarcu, I would assume that the point of the CLI is for DevOps shops to not need to know where the values come from?  I see it as an implementation behind the hood.

gtkramer:
Overall, this looks good and is a welcome change for DevOps shops

mixer/cmd/config.go
jwakre: I am also confused about these configuration files and what they should contain, especially mixer.state. The mixer.state file name doesn't make me think of a configuration file, but rather a file that contains information about the current state of the mix. It would make sense to me if mixer.state was read by mixer to gather information about the current mix and updated by mixer commands that change the values in mixer.state.

mixer/cmd/config.go
```diff
@@ -96,11 +98,49 @@ var configSetCmd = &cobra.Command{
 	},
 }
 
+var configGetCmd = &cobra.Command{
+	Use:   "get <property>",
+	Short: "Get a property in the config file",
```
jwakre: The phrasing of the Short message makes it sound like there is only one config file, but there are two. Also, I think it would be useful to explicitly write which files are config files in the help message. It may be confusing to users that mixer.state is a config file because it does not use the .conf suffix.

mixer/cmd/config.go
```diff
@@ -73,6 +73,17 @@ var configSetCmd = &cobra.Command{
 	the existence of the provided property, but will not validate the value provided.`,
 	Args: cobra.ExactArgs(2),
 	Run: func(cmd *cobra.Command, args []string) {
+		// Look for the value in mixer.state first
```
jwakre: Based on my feedback for the 'get' command, I think the help message should be updated.

mixer/cmd/config.go
rchiossi: The idea behind the introduction of `mixer.state` was to have a single file that included values that are either set by mixer or that could change between mixer versions, while builder.conf would be a "base recipe" with values provided by the user. The plan is to have files like mixver, upstreamver and other auxiliary files alike to be part of mixer.state. The main reason why this change has not happened yet is that the file was introduced right before we started working closely with devops and our focus changed to keep the process stable. Since changing those files also require updating devops release process, the change was put on hold.

Also note that `mixer.state` was never intended to have a get/set command, I always saw it as something that would be solely manipulated internally by mixer. But again, devops...

mixer/cmd/config.go
rchiossi: Indeed, that is the short version from the old PR. I'll update.

mixer/cmd/config.go
rchiossi: Also, there was a discussion not long ago about were the OFFLINE property should live. If it is in `mixer.state`, you can see it as "The mix is currently offline", while if it is in `builder.conf` you can see it as "This is an offline mix". The answer is whether users would shift between modes or set and forget. Either way, both solutions seems fine.
Now, even if we move it to `builder.conf` we still need a way to change values inside `mixer.state` because devops wants a way to set format without manually editing the file (and format is a value that for normal users should not change directly).

otaviobp:
Nice series. Patches looks good to me (but I haven't tested that)

builder/metadata.go
```diff
@@ -264,7 +279,7 @@ func (b *Builder) UpdateVersions(nextMix, nextUpstream uint32) error {
 	var latest uint32
 	var err error
 
-	checkUpstream := (nextUpstream != b.UpstreamVerUint32)
+	checkUpstream := (nextUpstream != b.UpstreamVerUint32) && !Offline
```
tmarcu: While this logic makes sense, it's becoming clear we have a growing issue with readability & code structure by needing these kinds of conditional statements. It would be useful to go over the code and see how we could restructure things to not need complex conditionals for things like offline. An option would be to move the offline check further into the actual function being called, i.e getUpstreamFormatRange().

mixer/cmd/config.go
tmarcu: @rchiossi I get what you're saying above but it doesn't really answer my question, we have a separation of content between confs, FORMAT is not parsed from builder.conf now, so missing mixer.state is an error, which mixer will try to solve by just picking a format for you. It has to fail if someone tries SetProperty() with format and mixer.state isn't there. Or am I misunderstanding?

rchiossi: Considering a better future where `mixer.state` is not auto-generated if it is missing (which is what we want), if mixer.state is missing and the user tries to set FORMAT, it will fail since Mix.FORMAT will not be found.

builder/metadata.go
rchiossi: I agree. Offline has the same issue that Native had before my CLI patch. In my opinion, the ideal solution would be isolating all commands that require network access under a separate network module and handle lack of network in there. This would make online/offline transparent for the rest of the code, pretty much like Native is right now.
With that said, I'd rather change this after we get offline working since it will is a pretty significant change to the code.

bat/tests/offline/Makefile
```diff
@@ -0,0 +1,9 @@
+.PHONY: check clean
```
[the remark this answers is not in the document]
rchiossi: both*
Good catch!

bat/tests/offline/run.bats
```diff
@@ -0,0 +1,127 @@
...
+  mixer help  --offline
+
+  #version
+  mixer --version --offline
+
+  # init
+  mixer init --clear-version $version --format $format --mix-version 10 --no-default-bundles --offline
+
+  # Prep for build
+  sed -i 's/os-core-update/os-core/' $BATS_TEST_DIRNAME/builder.conf
```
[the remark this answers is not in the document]
rchiossi: I agree, a mixer-init-minimal (or something like this) would be nice in the lib.

rchiossi:
Opening a new PR with the new version since it does not follow the same structure as this one.

PR #513, opened 2018-12-04
Makefile: use vendored dependencies when building

matthewrsj:
Allows for offline builds


Dockerfile
```diff
@@ -2,4 +2,4 @@ FROM clearlinux/mixer-ci:latest
 ENV LC_ALL="en_US.UTF-8"
```
rchiossi: Can't you set it here as well? Or do you need it set for that one command only?

matthewrsj: still trying to just get travis/docker to pass. I'll comment when this is actually ready for review again.

matthewrsj:
closing due to revert

PR #517, opened 2018-12-05
Enable iterative manifests with 0 file count

### LINKED ISSUE #516: Iterative Manifests: Build update fails when including bundle without file change
When a bundle is updated to add a new bundle to its include list without any file changes, iterative manifests generation will fail. This is because the list of files to be included in the iterative manifests would be empty. In this situation mixer outputs the following failure:

ERROR: Couldn't build update: failed to create update metadata: manifest has a zero file count

jwakre:
When a bundle is updated to only modify which bundles are included, an
iterative manifest can be generated with 0 files. This change adds a
type field to the Manifest struct to determine when an iterative
manifest is used and allows the creation of iterative manifests with
a 0 file count.

Fixes #516

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -82,6 +82,7 @@ func initBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error) {
 					TimeStamp: ui.timeStamp,
 				},
 				Name: bundleName,
+				Type: ManifestBundle,
 			}
 
 			if bundleName == "full" {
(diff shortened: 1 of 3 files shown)


rchiossi:
Ah, one small comment. This is a bit of personal preference, but I believe it is better to keep issue numbers out of the commit message since they don't really make sense if you are browsing git log. It is better if you are able to understand what is going on just by reading the commit history. You can always add extra info like issue number on the PR message when submitting it.

tmarcu:
You should always link to an appropriate issue # if it exists instead of making people hunt them down.

rchiossi:
Sure, you should link on github, but not on git. If we ever move the project somewhere else, this number is meaningless.

PR #518, opened 2018-12-10
builder: Experimental bundles

### LINKED ISSUE #497: Add to bundle Manifests header a flag saying that a bundle is experimental
Swupd needs that information in order to show users that information.

reaganlo:
From format 27 onwards, Mixer will recognize bundles with "[Status] : Experimental"
as experimental bundles and mark their status flag as "e" in the Manifest.MoM.

i.e. These bundles will include an "e" in the 2nd position of the manifest flags,
as shown here:

Me.. 4b91c3122e7e32f1e3edb597c6f89ff32cfffd977afb5e5 10 some-bundle

Fixes #497

### DIFF
--- a/builder/bundleset.go
+++ b/builder/bundleset.go
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+
+	"github.com/clearlinux/mixer-tools/swupd"
 )
 
 var (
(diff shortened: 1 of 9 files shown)


castulo:
As far as I can tell this looks good to me

otaviobp:
@reaganlo helped me and I could generate that here. And it's working fine with swupd code.

+1 then

builder/bundleset.go
```diff
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+
+	"github.com/clearlinux/mixer-tools/swupd"
```
tmarcu: This is out of scope of this PR, but I think the better thing to do would be to break out bundleset.go and bundle_control.go into their own package now, and include it in swupd and builder where needed, instead of pulling all of swupd in just for a header here.

swupd/format_switch.go
```diff
@@ -61,6 +61,17 @@ includes:	{{.Name}}
 `,
 }
 
+const statusExperimental = "Experimental"
```
rchiossi: Where does this string comes from? Swupd?

reaganlo: It comes from bundle header.
"[Status] : Experimental"

rchiossi: I mean, who sets that value?

reaganlo: Whoever creates/edits the bundle.

PR #526, opened 2018-12-18
create_manifests: Don't leak iterative manifests on minversions

### LINKED ISSUE #523: Iterative manifests do not adhere to minversions
Iterative manifests can leak across minversion updates. For example, in version 26780 the minversion is 26720 and some iterative manifests exist at older versions.

jwakre:
During a minversion update, iterative manifests older than the new
minversion were leaked into the new MoM. Also, format bumps perform
minversion updates, so iterative manifests from an old format could
leak into new formats. This change prevents forwarding iterative
manifests with versions older than the minversion into the new MoM.

Fixes #523

### DIFF
--- a/swupd/create_manifests.go
+++ b/swupd/create_manifests.go
@@ -279,6 +279,11 @@ func addUnchangedManifests(appendTo *Manifest, appendFrom *Manifest, bundles []s
 
 		bundleName := f.Name
 		if f.Type == TypeIManifest {
+			// Don't add iterative manifests across minversion updates
+			if f.Version < appendTo.Header.MinVersion {
+				continue
+			}
+
 			// Only the most recent iterative manifest for a bundle is valid,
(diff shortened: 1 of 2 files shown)


rchiossi:
If I understand correct, if there is a new version or a format bump iterative manifests are not created, but if they already existed previously, then they leak. Correct?
If so, patch looks good to me.

jwakre:
> If I understand correct, if there is a new version or a format bump iterative manifests are not created, but if they already existed previously, then they leak. Correct?
> If so, patch looks good to me.

Yes, format bumps do not create new iterative manifests, but existing ones are forwarded into the new MoM.

otaviobp:
btw, I haven't seen your comment about adding tests. sorry for saying it again.

rchiossi:
@otaviobp can you revise your review so we can merge this?

PR #530, opened 2019-01-04
WIP - Remove docker

### LINKED ISSUE #528: Remove support for running mixer commands inside container
Mixer is no longer required to run commands inside of a container and removing this feature reduces mixer's complexity. After this change, mixer would no longer need the '--native' flag because it will always run natively.

jwakre:
Mixer is no longer required to run commands inside a container. This PR removes does the following:

* Remove support for running commands in a container
* Deprecate native flag and remove references to it
* Remove DockerImgPath from builder.conf and bump the config file's minor version

Closes #528


jwakre:
@rchiossi based on our discussion on #528 I need to re-work this PR, so review/testing can wait.

jwakre:
This can be reopened when we are more ready to remove docker, so closing for now.

aaronovz1:
Can someone explain why docker is no longer needed now? I'm worried this will be problematic for use in a Yocto environment using meta-swupd due to the pseudo requirement and needing to be able to run commands as root so files can be created as other users/groups. Please see some of the original discussion here https://github.com/clearlinux/swupd-server/issues/111

I recently updated meta-swupd to use the mixer docker container so it'll be frustrating if I have to figure out an entirely new solution again.

rchiossi:
> Can someone explain why docker is no longer needed now? I'm worried this will be problematic for use in a Yocto environment using meta-swupd due to the pseudo requirement and needing to be able to run commands as root so files can be created as other users/groups. Please see some of the original discussion here [clearlinux/swupd-server#111](https://github.com/clearlinux/swupd-server/issues/111)
> 
> I recently updated meta-swupd to use the mixer docker container so it'll be frustrating if I have to figure out an entirely new solution again.

Hi @aaronovz1,

The fact that you could use mixer's format docker image to handle compatibility issues was a byproduct of the downstream format bump feature and was never the main purpose of having docker in mixer. Since mixer doesn't need multiple tooling versions to handle format bumps anymore, having docker just adds an unnecessary overlay that adds complexity to the code where it is not need.

We are aware that our users use this feature to  work around environment setup, but we believe that a dedicated docker image for that sole purpose would be better than the current images we have for format bumps. Also, for that use case in particular, docker can be called directly and does not need to be built in inside mixer. We are still evaluating what would be be best approach to tackle this use case though, so if you have any inputs on requirements that you may have, please do share with us so we can have a more informed decision on this change.

aaronovz1:
Hi @rchiossi,

As you suggested, I think as long as we have a simple docker container with mixer inside it then it will support the needs of the Yocto meta-swupd layer. The container allows the build environment to get around the need to run commands as root, which is required for creating files as other users/groups.

Cheers

Issue #467, opened 2018-10-18
Create a latest_format file in version/


otaviobp:
There's no way to know the last available version of Clearlinux just looking at content of meta-data generated by mixer.

Currently swupd knows checks the format file in the system or uses the default format file to get the ${SERVER}/version/format${FORMAT_NUMBER}/latest and check what's the latest version to be installed. If we don't know the latest format there's no way to proceed.

With that file we could have the following command in swupd:

    $ swupd verify --install -m latest -F latest --path /my_root_dir/

That is particularly useful to create images from custom built mixers.

This issue is particularly important for swupd right now because we need this information to automate some test scripts.

otaviobp:
What if the mixer was published in an http server and not in my system? Isn't this a use case?

And what about the official clearlinux content? How do I know the last format?

otaviobp:
My use case: I want to install the last available release published in a server in a chroot.

Is there a way to do that without the lastest_format file?

phmccarty:
@otaviobp I'm not thinking of a good approach to solve the problem you are outlining.

swupd-client can only understand a single format, so if it was able to consume a `${SERVER}/version/latest_format` when `latest_format` is greater than the client format, and there were some backwards incompatible changes in `latest_format`, swupd-client might not be able to consume the update content, depending on the nature of the breaking changes in the new format.

In other words, swupd-client would need to be forward compatible with any potential "breaking" changes to the update content produced by mixer.

otaviobp:
> swupd-client can only understand a single format, so if it was able to consume a ${SERVER}/version/latest_format when latest_format is greater than the client format, and there were some backwards incompatible changes in latest_format, swupd-client might not be able to consume the update content, depending on the nature of the breaking changes in the new format.

You have a point here. Maybe we should then have in swupd codebase any information about the format swupd supports. We don't have this right now.

I'll close this issue because you convinced me.

otaviobp:
I found a use case for this request, so I am reopening the ticket.
In swupd it would be nice to have the last_format/last_version listed in swupd check-update. This was requested in https://github.com/clearlinux/swupd-client/issues/482 and the best way to implement that is having that file.

cmarcelo:
I support having either the latest_format or the latest_version available *somewhere* would be useful for a couple of cases for me:

[...]

PR #533, opened 2019-01-16
builder: create a www/latest_version file

cmarcelo:
If publishing an update, also update the www/latest_version.  This is
useful for various consumers of swupd data: swupd-client can easily
discover the latest version (and not only the latest version in a
given format); swupd-inspector and swupd-extract will be able to
automatically figure the latest version of a mix given the update URL.

If --no-publish is used, the file will not be updated like the other
related files.  Clear Linux itself uses that so it can decide later on
whether to publish or not an update to the end users.

Related to #467.  With latest_version is possible to peek at
www/<latest_version>/format and figure its format.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -364,13 +364,17 @@ func (b *Builder) BuildUpdate(params UpdateParameters) error {
 		}
 	}
 
-	// Publish. Update the latest version both in the format (used by update itself) and in LAST_VER
-	// (used to create newer manifests).
+	// Publish. Update the latest version file in various locations.
 	if !params.Publish {
 		return nil
 	}
 
 	fmt.Printf("Setting latest version to %s\n", b.MixVer)
+
+	err = ioutil.WriteFile(filepath.Join(b.Config.Builder.ServerStateDir, "www", "version", "latest_version"), []byte(b.MixVer), 0644)
+	if err != nil {
+		return errors.Wrapf(err, "couldn't update the latest_version file")
+	}
 
 	err = ioutil.WriteFile(filepath.Join(formatDir, "latest"), []byte(b.MixVer), 0644)
 	if err != nil {


cmarcelo:
@rchiossi @otaviobp could you take a look?

rchiossi:
So, if --no-plublish is used, that means that the file must be manually updated afterwards?

cmarcelo:
Yes. We treat this file the same way we treat the other "latest version" file. If --no-publish is used, is up to the caller to fix up those files. If you later call publishing, the file will be updated.

For Clear Linux, we'll probably update the release scripts to ensure that -- they already set a "latest" outside update/ similar to this file.

rchiossi:
I see. Since this means that this feature would have to be implemented manually on devops side, I'll talk to them first for extra input.
It seems to me that for most use cases --no-publish would be used which means creating this file by hand afterwards. I wonder if there is value to add this to mixer without also providing a way to use mixer to publish the version once validation is done.
@mbelluzzo isn't this also true for your usecase?

cmarcelo:
Note: the situation is not that different than before (think about the versions/<format>/latest file), when people use `--no-publish` is up to them either: make a new published release (what a mix might do) or touch the files by hand (what Clear does).

builder/builder.go
```diff
@@ -364,14 +364,18 @@ func (b *Builder) BuildUpdate(params UpdateParameters) error {
...
-	// Publish. Update the latest version both in the format (used by update itself) and in LAST_VER
-	// (used to create newer manifests).
+	// Publish. Update the latest version file in various locations.
 	if !params.Publish {
 		return nil
 	}
 
 	fmt.Printf("Setting latest version to %s\n", b.MixVer)
 
+	err = ioutil.WriteFile(filepath.Join(b.Config.Builder.ServerStateDir, "www", "latest_version"), []byte(b.MixVer), 0644)
```
otaviobp: for me it makes more sense to write that file to www/version/ because all formats and latest from formats are in that folder

cmarcelo: Changed to write the file to www/version/ directory. @otaviobp  Could you take another look?

otaviobp: it's better, but we need to address the issue raised by @rchiossi

mbelluzzo:
> I see. Since this means that this feature would have to be implemented manually on devops side, I'll talk to them first for extra input.
> It seems to me that for most use cases --no-publish would be used which means creating this file by hand afterwards. I wonder if there is value to add this to mixer without also providing a way to use mixer to publish the version once validation is done.
> @mbelluzzo isn't this also true for your usecase?

I actually dont like this publish/no publish on the mixer side. That should be up to devops. Mixer IMHO should always produce an output that is "publishable" as is: IOW always create files and always update it. I control the actual publishing by having mixer to output in a stagging place, once that is approved, I sync the content with the folder that holds the actual published content. For me, that is a more sane way to operate. Others may disagree.

otaviobp:
I found a real use case for a latestversion file. Right now, from swupd perspective, if we don't know which format we are in, there isn't any file that we are sure that will always be in a mix.
So to check if the SSL connection is working we make a GET request to '/'. This works fine for CL servers, but there are some webservers that won't respond to '/'. 
If we had a file always present in the mix we could use that instead.

rchiossi:
This is now part of our Q1 release target. I'll coordenate with devops to get this change in.

cmarcelo:
> This is now part of our Q1 release target. I'll coordenate with devops to get this change in.

Ping?  (we are in about 2/3 of Q1 :-)

rchiossi:
Sorry for the delay. I got a meeting with devops today, I'll make sure I get an update on this.

mbelluzzo:
This needs to respect '--no-publish' because of the way releases are synced today.

cmarcelo:
> This needs to respect '--no-publish' because of the way releases are synced today.

Could you take another look? The patch already respects --no-publish.

gtkramer:
I agree with @mbelluzzo 's thoughts.  Publishing is more a DevOps concern more than a mixer concern, so the publish/no publish options don't make much sense to be in mixer.  Mixer should always "publish" and a more sane way of exposing content should be put in place by DevOps when validation is also taken into account.  That's more of a tangential thought with respect to this PR...

This PR solves the aforementioned problem for downstreams.  DevOps has solved it for upstream.  We have https://download.clearlinux.org/latest.  From there, format can be resolved from the update folder.  Automation could point here, but I also think it makes sense to have this be put in the update folder as well since it's swupd-related metadata.  We can then consolidate swupd automation to use this file, as mentioned, so that sounds good too.  This just means that DevOps will also need to produce/manage two latest files: one forwards compatible with downstream mixer users, and one backwards compatible for existing automation that's been built up.

rchiossi:
@gtkramer @mbelluzzo I see your point and I understand the actions that need to be taken on devops side to make this work, but what we want to know here is what needs to be done in this PR, if anything, so we can enable this on mixer side.

mbelluzzo:
Due to format staging, this also introduces yet another artifact that devops needs to control by hand, because we cannot have 'latest_version' pointing to a version that is only at staging.

IMHO correct solution is to mixer realize that the distro is not only the content that exists inside update/. If it wants to generate the latest file, it needs to be in the equivalent place where it exists today.

If that is impossible from mixer side, DevOps will end up have to care about yet another pointer when doing releases. To avoid 'latest' to diverge from 'update/latest_version' as well to avoid 'latest_version' to apear for users when a version is still only pointed by format staging.

So, unless there is a change in mindset about what mixer can output in terms of the expected file tree for the distro and start to consider that its 'www' should be equivalent to 'download.clearlinux.org/' instead of 'download.clearlinux.org/update/', there is nothing that DevOps can contribute to this discussion other than warn that this only adds to the number of artifacts that needs to be handcraft during every 'build & release'.

gtkramer:
Maybe something less controversial, @cmarcelo @otaviobp @rchiossi could the file be named latest-version, rather than using an underscore?

rchiossi:
@gtkramer Any particular reason why you prefer `lastest-version` rather than `latest_version`? I don't see a problem in the name change. I think i like the underscore more, but changing it won't  be a problem at all.

PR #534, opened 2019-01-16
Offline fix

rchiossi:
Those 3 patches fix code patches where `--offline` was ignored. They are required for downstream mixes to perform a manual format bump.


rchiossi:
@mbelluzzo this is the fix I talked to you about yesterday. With this in you should be able to bump without recompiling mixer.

mbelluzzo:
> @mbelluzzo this is the fix I talked to you about yesterday. With this in you should be able to bump without recompiling mixer.

Sweet! I think that would work without even requiring changes on my side.

Issue #531, opened 2019-01-10
Build bundles command can fail to download packages


jwakre:
When building bundles, sometimes Mixer fails to download a package and exits unsuccessfully. The error is easier to reproduce when building with a large set of bundles.

**Steps to reproduce:**

$ mixer init --all-upstream
$ sudo mixer bulid bundles --native

**Error snippets from output:**

[MIRROR] net-tools-man-1.60-21.x86_64.rpm: Curl error (56): Failure when receiving data from the peer for https://download.clearlinux.org/releases/27150/clear/x86_64/os/Packages/net-tools-man-1.60-21.x86_64.rpm []
[FAILED] net-tools-man-1.60-21.x86_64.rpm: No more mirrors to try - All mirrors were already tried without success
.
.
.
Error: Error downloading packages:
  Cannot download Packages/net-tools-man-1.60-21.x86_64.rpm: All mirrors were tried

jwakre:
When building the full chroot, sometimes dnf install fails to download the desired rpm. Subsequent calls to install the package usually succeed, but since there is no retry code in place, any failure is fatal.

PR #535, opened 2019-01-17
builder: Retry RPM downloads for build bundles

jwakre:
Download failures can occur when downloading large sets of RPMs from a
server with a poor connection. This change adds retry functionality to
attempt RPM downloads multiple times.

Fixes: #531

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -436,17 +436,51 @@ func genUpdateBundleSpecialFiles(chrootDir string, b *Builder) error {
 	return ioutil.WriteFile(filepath.Join(swupdDir, "format"), []byte(b.State.Mix.Format), 0644)
 }
 
-func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *bundle) error {
+func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int) error {
+	var downloadErr error
+
+	if maxRetries < 0 {
+		return errors.Errorf("maxRetries value < 0 for RPM downloads")
+	}
+
+	args := merge(packagerCmd, "--installroot="+baseDir, "install", "--downloadonly")
+	args = append(args, rpmList...)
+
+	for attempts := 0; attempts <= maxRetries; attempts++ {
+		_, downloadErr = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
+		if downloadErr == nil {
+			return downloadErr
+		}
+
+		fmt.Printf("RPM download attempt %d failed. Maximum of %d attempts.\n", attempts+1, maxRetries+1)
+		if err := clearDNFCache(packagerCmd); err != nil {
+			return err
+		}
+	}
+	return downloadErr
+}
+
+func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *bundle, downloadRetries int) error {
 	var err error
 	baseDir := filepath.Join(buildVersionDir, "full")
-	args := merge(packagerCmd, "--installroot="+baseDir, "install")
+	rpmList := []string{}
+
 	if len(bundle.AllPackages) > 0 {
 		// There were packages directly included for this bundle so
 		// install to full chroot. This check is necessary so we don't
 		// call dnf install with no package listed.
 		for p := range bundle.AllPackages {
-			args = append(args, p)
+			rpmList = append(rpmList, p)
 		}
+
+		// Retry RPM downloads to avoid timeout failures due to slow network
+		err = downloadRpms(packagerCmd, rpmList, baseDir, downloadRetries)
+		if err != nil {
+			return err
+		}
+
+		args := merge(packagerCmd, "--installroot="+baseDir, "install")
+		args = append(args, rpmList...)
 		_, err = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
 		if err != nil {
 			return err
@@ -502,7 +536,7 @@ func rmDNFStatePaths(fullDir string) {
 	}
 }
 
-func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVersionDir, version string) error {
+func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVersionDir, version string, downloadRetries int) error {
 	fmt.Println("Cleaning DNF cache before full install")
 	if err := clearDNFCache(packagerCmd); err != nil {
 		return err
@@ -522,7 +556,7 @@ func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVers
 			}
 		}
 
-		if err := installBundleToFull(packagerCmd, buildVersionDir, bundle); err != nil {
+		if err := installBundleToFull(packagerCmd, buildVersionDir, bundle, downloadRetries); err != nil {
 			return err
 		}
 
@@ -556,7 +590,7 @@ func writeBundleInfo(bundle *bundle, path string) error {
 	return ioutil.WriteFile(path, b, 0644)
 }
 
-func (b *Builder) buildBundles(set bundleSet) error {
+func (b *Builder) buildBundles(set bundleSet, downloadRetries int) error {
 	var err error
 
 	if b.Config.Builder.ServerStateDir == "" {
@@ -697,7 +731,7 @@ src=%s
 	}
 
 	// install all bundles in the set (including os-core) to the full chroot
-	err = buildFullChroot(b, &set, packagerCmd, buildVersionDir, version)
+	err = buildFullChroot(b, &set, packagerCmd, buildVersionDir, version, downloadRetries)
 	if err != nil {
 		return err
 	}
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -33,17 +33,21 @@ import (
 
 const bumpMarker = "format-bump"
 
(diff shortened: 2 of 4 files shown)


builder/bundles.go
```diff
@@ -447,6 +447,22 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
 		for p := range bundle.AllPackages {
 			args = append(args, p)
 		}
+
+		// Retry RPM downloads to avoid timeout failures due to slow network
```
otaviobp: I prefer to have a function downloadRpms(packagerCmd, baseDir, max_retries) to encapsulate that.

builder/bundles.go
```diff
@@ -447,6 +447,22 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
...
+		retries := 0
+		downloadArgs := merge(args, "--downloadonly")
+
+	DOWNLOAD_RPMS:
+		_, err = helpers.RunCommandOutputEnv(downloadArgs[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
+		if err != nil {
+			if retries < 2 {
+				retries++
+				fmt.Println("Retrying after failed RPM download")
+				goto DOWNLOAD_RPMS
```
otaviobp: I don't like gotos with the only exception for going to the end of a function on error handling.
It's a lot easier to read if you have

    while (retry < max_retries) {
         ...
         if (err == nil) break; //Or return if you create another function
    }

jwakre:
It looks like there is a problem with the webserver hosting download content.

builder/bundles.go
jwakre: sure

builder/bundles.go
jwakre: I was going back-and-forth on this. I can switch it to a loop implementation.

rchiossi: I'm with otavio here, specially considering that feels out of place compared to the rest of the code base.

builder/bundles.go
```diff
@@ -436,17 +436,43 @@ func genUpdateBundleSpecialFiles(chrootDir string, b *Builder) error {
...
+func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int) error {
+	var err error
+
+	args := merge(packagerCmd, "--installroot="+baseDir, "install", "--downloadonly")
+	args = append(args, rpmList...)
+
+	for retries := 0; retries < maxRetries; retries++ {
+		_, err = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
+		if err == nil {
+			return err
```
rchiossi: Doesn't really matter here, but you could also call `break` to end the `for` loop. No need to change it though.

builder/bundles.go
```diff
@@ -436,17 +436,43 @@ func genUpdateBundleSpecialFiles(chrootDir string, b *Builder) error {
...
 		// There were packages directly included for this bundle so
 		// install to full chroot. This check is necessary so we don't
 		// call dnf install with no package listed.
 		for p := range bundle.AllPackages {
-			args = append(args, p)
+			rpmList = append(rpmList, p)
+		}
+
+		// Retry RPM downloads to avoid timeout failures due to slow network
+		err = downloadRpms(packagerCmd, rpmList, baseDir, 3)
```
rchiossi: It would be nicer to have a constant defined with the number of retries instead of having a magic number on the function call. I also think it would be nice if this parameter could be set via command line, so users with troublesome connection could thinker with a better value.

jwakre: Ok, I can add this.

otaviobp:
I was looking at DNF manual (https://dnf.readthedocs.io/en/latest/conf_ref.html) and the standard number of retries in dnf is 10. Are we overriding this or we are going to retry dnf command after dnf failed 10 times?

And if we are overriding this, can't we just set the retry in dnf instead of adding that code to swupd?

I think it makes sense to retry on errors, but it's weird to have a failure at first. I'd prefer to have this problem investigated instead of just implementing a retry strategy.

Another question. Have you tried to reproduce that using dnf directly?

jwakre:
> I was looking at DNF manual (https://dnf.readthedocs.io/en/latest/conf_ref.html) and the standard number of retries in dnf is 10. Are we overriding this or we are going to retry dnf command after dnf failed 10 times?

The DNF command is using the .yum-conf from the Mixer workspace. I played with the timeout and retries parameters in this file, but I couldn't get rid of the failure this way. Something feels strange about the default mirror, but I couldn't find the root cause.

> And if we are overriding this, can't we just set the retry in dnf instead of adding that code to swupd?

I think this is default of 10.

> I think it makes sense to retry on errors, but it's weird to have a failure at first. I'd prefer to have this problem investigated instead of just implementing a retry strategy.

This is fair, I was just trying to unblock users.

> Another question. Have you tried to reproduce that using dnf directly?

No, but I can try it.

builder/bundles.go
```diff
@@ -436,17 +436,51 @@ func genUpdateBundleSpecialFiles(chrootDir string, b *Builder) error {
...
+	args = append(args, rpmList...)
+
+	for retries := 0; retries < maxRetries; retries++ {
+		_, err = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
+		if err == nil {
+			return err
+		}
+
+		fmt.Printf("RPM download attempt %d failed. Retrying up to %d times.\n", retries+1, maxRetries)
+		if err := clearDNFCache(packagerCmd); err != nil {
```
jwakre: I added the clearDNFCache function to remove potentially corrupted data from the DNF cache.

rchiossi:
If I understand correctly this last changes are to address the issue with embargo. @mbelluzzo is there any way you can verify if this fix your issue before we merge?

builder/bundles.go
```diff
@@ -436,17 +436,51 @@ func genUpdateBundleSpecialFiles(chrootDir string, b *Builder) error {
 	return ioutil.WriteFile(filepath.Join(swupdDir, "format"), []byte(b.State.Mix.Format), 0644)
 }
 
-func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *bundle) error {
+func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int) error {
+	var downloadErr error
```
jwakre: Renamed err -> downloadErr so that it does not get overwritten by err := clearDNFCache(

mbelluzzo:
I can try this PR during this week on my infra and see how it does. The error is not predictable, so an one-shot try wont be enough. I can provide an update on this later this week.

mbelluzzo:
I left this running for the entire week and besides a single error I had on the first attempt (which never occurred again) the retry saved me a handful of builds so I'm happy with that. Even that we haven't root caused the real issue, this works as a workaround.

rchiossi:
If that is the case, I'll include it in our next release. Thanks for testing it out @mbelluzzo .

builder/bundles.go
```diff
@@ -436,17 +436,51 @@ func genUpdateBundleSpecialFiles(chrootDir string, b *Builder) error {
...
+
+	if maxRetries <= 0 {
```
rchiossi: I believe the name is a bit confusing here since `retry` imply trying it again, but from this code, If I don't want mixer to retry a download if it fails I need to set retry to 1 instead of 0 which can be misleading. I think it would be better to check if `maxRetries >= 0` and look from `0` to `1 + maxRetries` instead, since we always wanna do at least one loop.

jwakre: Good catch! I'll update.

jwakre:
@rchiossi I changed the logic so that retries sets the number of download retries instead of download attempts. Also, I renamed the retries variable to attempts which is more accurate.

mixer/cmd/build.go
```diff
@@ -585,6 +592,12 @@ func init() {
 
 	RootCmd.AddCommand(buildCmd)
 
+	buildAllCmd.Flags().IntVar(&buildFlags.downloadRetries, "retries", retriesDefault, "Number of retry attempts to download RPMs")
+	buildBundlesCmd.Flags().IntVar(&buildFlags.downloadRetries, "retries", retriesDefault, "Number of retry attempts to download RPMs")
+	buildFormatBumpCmd.Flags().IntVar(&buildFlags.downloadRetries, "retries", retriesDefault, "Number of retry attempts to download RPMs")
+	buildFormatOldCmd.Flags().IntVar(&buildFlags.downloadRetries, "retries", retriesDefault, "Number of retry attempts to download RPMs")
+	buildUpstreamFormatCmd.Flags().IntVar(&buildFlags.downloadRetries, "retries", retriesDefault, "Number of retry attempts to download RPMs")
```
rchiossi: Shouldn't it be a top level flag instead like bundle-workers?

jwakre: There is already a bundle-workers flag that specifies the number of parallel workers when building bundles

rchiossi: I know, I used it as an exemple of a top level flag.

rchiossi: I'm suggesting that you replace all this code by:
`buildCmd.PersiustentFlags().IntVar(&buildFlags.downloadRetries, "retries", retriesDefault, "Number of retry attempts to download RPMs")`

jwakre: Oh, I see what you mean now... I can do that. It just seems weird to me that the option would be exposed to ~3 commands that don't use it.

jwakre: Updated the PR to make it a persistent flag

PR #539, opened 2019-01-25
Increment Clear repo default timeout

jwakre:
Since the default Clear repo mirror changed to the cdn, dnf timeouts
occur more frequently. Increasing the timeout to 45 seconds from the
default timeout of 30 seconds.


jwakre:
Since the default mirror switch to the CDN, I have hit DNF's default 30 second timeout 3 times. Raising the timeout makes this less likely to happen.

PR #543, opened 2019-02-08
Enable create Manifest deltas in mixer

bryteise:
swupd-server used to support building bsdiff based deltas of manifest
files. This change adds that ability back for bundle manifests with
the 'mixer build delta-manifests' command.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -609,3 +609,114 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
 	}
 	return nil
 }
+
+// BuildDeltaManifests between two versions of the mix.
+func (b *Builder) BuildDeltaManifests(from, to uint32) error {
+	var err error
+
+	if to == 0 {
+		to = b.MixVerUint32
+	} else if to > b.MixVerUint32 {
+		return errors.Errorf("--to version must be at most the latest mix version (%d)", b.MixVerUint32)
+	}
+	if from == to {
+		fmt.Println("the --from version matches the --to version, nothing to do")
+		return nil
+	} else if from > to {
+		return errors.Errorf("the --from version must be smaller than the --to version")
+	}
+
+	outputDir := filepath.Join(b.Config.Builder.ServerStateDir, "www")
+
+	toManifest, err := swupd.ParseManifestFile(filepath.Join(outputDir, fmt.Sprint(to), "Manifest.MoM"))
+	if err != nil {
+		return errors.Wrapf(err, "couldn't find manifest of target version")
+	}
+
+	fromManifest, err := swupd.ParseManifestFile(filepath.Join(outputDir, fmt.Sprint(from), "Manifest.MoM"))
+	if err != nil {
+		return errors.Wrapf(err, "couldn't find manifest of from version")
+	}
+
+	fmt.Printf("Using %d workers\n", b.NumDeltaWorkers)
+	fmt.Printf("Creating Manifest delta files from %d to %d\n", from, to)
+	deltas, err := swupd.CreateManifestDeltas(b.Config.Builder.ServerStateDir, fromManifest, toManifest, b.NumDeltaWorkers)
+	if err != nil {
+		log.Printf("  %s\n", err)
+	} else {
+		created := 0
+		for _, delta := range deltas {
+			if delta.Error == nil {
+				created++
+			}
+		}
+		fmt.Printf("  Created %d Manifest delta files\n", created)
+	}
+
+	return nil
+}
+
+// BuildDeltaManifestsPreviousVersions builds manifests to version from up to
+// prev versions. It walks the Manifest "previous" field to find those from versions.
+func (b *Builder) BuildDeltaManifestsPreviousVersions(prev, to uint32) error {
+	var err error
+
+	if to == 0 {
+		to = b.MixVerUint32
+	} else if to > b.MixVerUint32 {
+		return errors.Errorf("--to version must be at most the latest mix version (%d)", b.MixVerUint32)
+	}
+
+	outputDir := filepath.Join(b.Config.Builder.ServerStateDir, "www")
+	toManifest, err := swupd.ParseManifestFile(filepath.Join(outputDir, fmt.Sprint(to), "Manifest.MoM"))
+	if err != nil {
+		return errors.Wrapf(err, "couldn't find manifest of target version")
+	}
+
+	var previousManifests []*swupd.Manifest
+	cur := toManifest.Header.Previous
+	for i := uint32(0); i < prev; i++ {
+		if cur == 0 {
+			break
+		}
+		var m *swupd.Manifest
+		m, err = swupd.ParseManifestFile(filepath.Join(outputDir, fmt.Sprint(cur), "Manifest.MoM"))
+		if err != nil {
+			log.Printf("Warning: Could not find manifest for previous version %d, skipping...\n", cur)
+			continue
+		}
+		// do not create delta-manifests over format bumps since clients can't update
+		// past the boundary anyways. Only check for inequality, if the format
+		// goes down that should be checked elsewhere.
+		if m.Header.Format != toManifest.Header.Format {
+			log.Println("Warning: skipping delta-pack creation over format bump")
+			break
+		}
+		previousManifests = append(previousManifests, m)
+		cur = m.Header.Previous
+	}
+
+	fmt.Printf("Found %d previous versions\n", len(previousManifests))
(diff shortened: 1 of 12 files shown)


builder/builder.go
```diff
@@ -609,3 +609,104 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
...
+		var m *swupd.Manifest
+		m, err = swupd.ParseManifestFile(filepath.Join(outputDir, fmt.Sprint(cur), "Manifest.MoM"))
+		if err != nil {
+			log.Printf("Warning: Could not find manifest for previous version %d, skipping...\n", cur)
+			continue
+		}
+		// do not create delta-packs over format bumps since clients can't update
+		// past the boundary anyways. Only check for inequality, if the format
+		// goes down that should be checked elsewhere.
+		if m.Header.Format != toManifest.Header.Format {
```
rchiossi: shouldn't this check also exist in `BuildDeltaManifests`?

builder/builder.go
```diff
@@ -609,3 +609,104 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
...
+
+	outputDir := filepath.Join(b.Config.Builder.ServerStateDir, "www")
+	toManifest, err := swupd.ParseManifestFile(filepath.Join(outputDir, fmt.Sprint(to), "Manifest.MoM"))
+	if err != nil {
+		return errors.Wrapf(err, "couldn't find manifest of target version")
+	}
+
+	var previousManifests []*swupd.Manifest
+	cur := toManifest.Header.Previous
+	for i := uint32(0); i < prev; i++ {
```
rchiossi: Could we have a separate function to generate the manifest instead? It seems to me that the logic in the loop plus the loop below is the same as the one in `BuildDeltaManifests`, so it would be nice if we could reuse the same code for both.

docs/mixer.build.1
```diff
@@ -201,6 +201,45 @@ Generate packs targeting a specific \fIto\fP \fIversion\fP\&.
...
+if it exists and apply that on the bundle manifest file for the version
+installed on their system (if it exists). This can save a large amount of
+content being downloaded in the case of few files changing in a manifest.
+Because the client can fall back to the full manifest file if a delta is not
+available, delta manifests are not necessary for a functional update. In
+addition to the global options \fBmixer build delta\-manifests\fP takes the
+following options.
+.INDENT 0.0
+.IP \(bu 2
+\fB\-c, \-\-config {path}\fP
```
rchiossi: this is now a global flag so it doesnt need to be listed here. This will also change on the other commands as well, I'm working on updating the docs.

docs/mixer.build.1.rst
```diff
@@ -162,6 +162,40 @@ SUBCOMMANDS
...
+    ``swupd`` client update runs, it will first try to get a delta manifest file
+    if it exists and apply that on the bundle manifest file for the version
+    installed on their system (if it exists). This can save a large amount of
+    content being downloaded in the case of few files changing in a manifest.
+    Because the client can fall back to the full manifest file if a delta is not
+    available, delta manifests are not necessary for a functional update. In
+    addition to the global options ``mixer build delta-manifests`` takes the
+    following options.
+
+    - ``-c, --config {path}``
```
rchiossi: same as above

bryteise: Will fix.

docs/mixer.build.1
bryteise: Will fix.

bryteise:
This change is being submitted as an alternative to the other approach which has more risk due to the correctness of the deltas needing more testing as opposed to this where the correctness is assumed based on the full manifest (since the hashes would be equal in order to be used by swupd-client). This change could be replaced by the other method but the lack of delta usage is really hurting update performance and this is a semi urgent fix.

builder/builder.go
bryteise: Ah indeed, also fixing the comment.

bryteise: Hrm so I basically took the implementation from BuildDeltaPacks which doesn't do this check. Should it be added there too? (I'll probably not do that though via this PR).

rchiossi: I can't think of a situation where we would need deltas across formats so it might make sense there as well.

rchiossi:
I see. Still, I`ll wait for @otaviobp 's input on this one since he knows better what are the plans for this feature.

builder/builder.go
bryteise: So this is a similar pattern to the BuildDeltaPacks and BuildDeltaPacksPreviousVersions. Is that needing a similar refactor? I'm a bit unlucky on the code I stole from lol. I think these are a little awkward to combine though since one is using a specific version and the other is going back a count from the current. I would likely have just written it as one function that checks which argument is valid and then created a single or multiple element list and processed that the same way. Do you want me to go that route even though it diverges from how the pack functions are laid out?

rchiossi: Well, if that is the case, I'm fine keeping it this way for now. If the others need a refactor, I'm ok with keeping the current patern for now do it all at once later on. This is also out of scope of this PR anyway.

jwakre:
@bryteise Do we plan on keeping the current delta manifest implementation indefinitely in Mixer or are we removing it soon? If we intend to keep both implementations, the naming terminology could cause confusion in the future. Both delta manifests and manifest deltas use the term 'delta manifest', but for different features.

builder/builder.go
```diff
@@ -609,3 +609,104 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
...
 }
+
+// BuildDeltaManifests between two versions of the mix.
+func (b *Builder) BuildDeltaManifests(from, to uint32) error {
+	var err error
+
+	if to == 0 {
+		to = b.MixVerUint32
+	} else {
+		if to > b.MixVerUint32 {
```
jwakre: Nit: Could be elise if

builder/builder.go
```diff
@@ -609,3 +609,104 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
...
+		if cur == 0 {
+			break
+		}
+		var m *swupd.Manifest
+		m, err = swupd.ParseManifestFile(filepath.Join(outputDir, fmt.Sprint(cur), "Manifest.MoM"))
+		if err != nil {
+			log.Printf("Warning: Could not find manifest for previous version %d, skipping...\n", cur)
+			continue
+		}
+		// do not create delta-packs over format bumps since clients can't update
```
otaviobp: s/delta-packs/delta-manifest/

builder/builder.go
```diff
@@ -609,3 +609,104 @@ func (b *Builder) BuildDeltaPacksPreviousVersions(prev, to uint32, printReport b
...
+		// past the boundary anyways. Only check for inequality, if the format
+		// goes down that should be checked elsewhere.
+		if m.Header.Format != toManifest.Header.Format {
+			log.Println("Warning: skipping delta-pack creation over format bump")
+			continue
```
otaviobp: Can't you break here? If we found a format-bump all other previous version are going to be in a format bump region.

otaviobp:
We could have an output similar to delta-pack creation too. I didn't know if the command was successful or not at the end.

jwakre:
When manifest deltas are generated, they create delta files for every bundle, even when the bundle was not updated (manifest deltas from bundles that did not change will be duplicated from prior versions). For instance, if one bundle was changed in a new version, there will be manifest deltas generated for every bundle.

Typically, manifests without changes are not copied forward to the update content for newer versions. Instead, Swupd should be smart enough to find the manifest deltas in older versions.

builder/builder.go
bryteise: Ah indeed, copypasta'd logic and didn't think too much about the extra if, fixing.

builder/builder.go
bryteise: Yea, this was copypasta logic from the delta-pack (obviously given the above X( ). I'll fix.

builder/builder.go
bryteise: Will fix.

rchiossi:
Just reiterating what @otaviobp said, we also need automated tests in this PR in order to integrate it. Since a couple monts ago, we changed our policy to require any new feature to also includes automated tests, either to bat tests or to unity test.

bryteise:
Alrighty I think I've addressed all the comments on the PR with my latest changes. I added two bat tests, corrected when generating manifest delta files to only if the bundle version in the to mom is the same as the to version and added some output for the processing of manifest delta files.

PR #553, opened 2019-03-03
builder: don't use logger date/time

cmarcelo:
Logger default date/time are prefixes and don't play well when
printing multiple lines.  What usually matters is the amount of time
elapsed since the start, so print that instead (when each new entry
that is not the first is starting).


rchiossi:
I rebased locally and pushed the change.

PR #555, opened 2019-03-04
Add MCA support

jwakre:
MCA collects RPM metadata and compares it against bundle manifests to verify that the manifests were generated correctly.

TODO:
  - Add tests

Example:

  - mixer build validate --from 10 --to 30


jwakre:
@rchiossi @phmccarty 

With the exception of bundle deletes, MCA is working to identify manifest errors and print package statistics. Currently, I am working to add bundle delete support, parallelism, and tests. I can include any feedback provided in the next revision.

Also, I am open to suggestions on where to store the MCA code and how it should be called by Mixer.

builder/build_validate.go
```diff
@@ -0,0 +1,911 @@
...
+}
+
+// CheckManifestCorrectness validates that the changes in manifest files between
+// two versions aligns to the corresponding RPM changes. Any mismatched files
+// between the manifests and RPMs will be printed as errors. When there are no
+// errors, package and file statistics for each modified bundle will be displayed.
+func (b *Builder) CheckManifestCorrectness(fromVer, toVer, downloadRetries int) error {
+	/* TODO
+	- Add bundle delete support
+	- Add format bump support
```
jwakre: Format bumps add more file exceptions where files are modified in manifests without a corresponding package change. Also, the current MCA does not support cross format comparisons, so I am planning on disabling MCA when the formats are different.

phmccarty: You're right that the current MCA doesn't *support* cross-format comparisons, but they do occur in practice. During a format bump, MCA compares from +0 to +10 (same format), and from +0 to +20 (different formats). Then, the *reports* are compared to ensure that the expected changes appear:

- Because a new minversion is generally set during a format bump, the +0->+20 report indicates that a minversion bump occurred for every bundle in the distro.
- All other package/file adds, changes, and deletes should be identical for both +0->+10 and +0->+20 reports.
- Bundle deletes may occur, but are not reported about, as you have mentioned.

phmccarty: Note that the report comparison I just described is not performed by MCA -- it's a separate check that's currently not automated. I wanted to mention it here though, because the report comparisons *have* caught format bump process issues in the past.

jwakre: Ok, I can update the checks for manifest files without package equivalents to account for formats. Also, I will need to account for /usr/share/defaults/swupd/format which is changed in the +10.

jwakre: It is challenging to compare special case files (files changed by manifest without package equivalent) across different formats because there is no way to determine whether a version is a +20 or an arbitrary version in the same format. The +20 requires different treatment than other versions, but it cannot be determined without tracking the first version in the format.

Based on offline discussion, MCA will correctly check special case files when comparing manifests within the same format and the +10 -> +20 comparison. When comparing manifests in separate formats (excluding the +10 -> +20 comparison), special case file errors will generate warning messages instead of errors.

builder/build_validate.go
```diff
@@ -0,0 +1,911 @@
...
+			mandatoryFiles++
+		case "/usr/share/clear/versionstamp":
+			mandatoryFiles++
+		default:
+			tmp = append(tmp, f)
+		}
+	}
+	results.bundleDiff["os-core"].manFileDiffs.modList = tmp
+
+	if mandatoryFiles != 3 {
```
jwakre: These file exceptions will be refactored to handle +10, +20, and different formats.

builder/build_validate.go
```diff
@@ -0,0 +1,911 @@
...
+			allFiles: make(map[string]bool),
+			allPkgs:  make(map[string]*pkgInfo),
+		}
+
+		var pkgList = []string{}
+		for pkg := range m.BundleInfo.AllPackages {
+			pkgList = append(pkgList, pkg)
+		}
+
+		out, err := downloadRpms(packagerCmd, pkgList, buildVersionDir, downloadRetries)
```
jwakre: Packages in the local-rpms folder override the upstream repo, so I plan to disable the local repo for MCA. If the 'Mixer add-rpms' command was able to create a repo that supported release versions, MCA could add support for local RPMs.

jwakre:
With this change MCA is now able to handle previously stated shortcomings (still needs tests). This revision adds support for the following:
*  format bumps
*  bundle deletion
*  minversions
*  parallel package downloads/resolves

Additionally, this change adds warning messages when there is not enough information to guarantee special case file handling and to notify the user that the local RPM repo is disabled for MCA. The local RPM repo overrides the upstream repo by default which skews MCA results.

I'll continue work on tests, and address any provided feedback.

builder/build_validate.go
```diff
@@ -0,0 +1,927 @@
...
+			tmp = append(tmp, f)
+		}
+	}
+	results.bundleDiff["os-core"].manFileDiffs.modList = tmp
+
+	if mandatoryFiles != 3 {
+		errorList = append(errorList, "ERROR: os-core is missing mandatory modified file(s)\n")
+	}
+
+	// Compare manifest files against package files. Inconsistencies are added to the errorList.
```
rchiossi: What about other files added directly to the chroot between "build bundles" and "build update"? Won't they cause trouble here?

builder/build_validate.go
```diff
@@ -0,0 +1,927 @@
...
+
+// printMcaResults displays any MCA errors and prints bundle diff statistics when there are no errors.
+func printMcaResults(results *mcaDiffResults, fromInfo, toInfo map[string]*mcaBundleInfo, fromVer, toVer int, errorList []string) error {
+	var err error
+
+	// An overwhelming number of errors can be generated when this test
+	// identifies a manifest bug, so limit the error output to 50.
+	if len(errorList) > 0 {
+		for i, err := range errorList {
+			if i == 50 {
```
rchiossi: Would be nice to print a warning here to let the user know that there were errors that got omitted.

mixer/cmd/build.go
```diff
@@ -662,6 +683,9 @@ func init() {
 	_ = buildBundlesCmd.Flags().MarkHidden("new-chroots")
 	_ = buildBundlesCmd.Flags().MarkDeprecated("new-chroots", "new functionality is now the standard behavior, this flag is obsolete and no longer used")
 
+	buildValidateCmd.Flags().IntVar(&buildFlags.to, "to", 0, "To version")
+	buildValidateCmd.Flags().IntVar(&buildFlags.from, "from", 0, "From version")
```
rchiossi: It would be nice to have a more descriptive explanation of what "To"and "From" are since this is a help message.

builder/build_validate.go
```diff
@@ -0,0 +1,927 @@
...
+			fileMetadata[0] = strings.Replace(fileMetadata[0], "/bin/", "/usr/bin/", 1)
+		} else if strings.HasPrefix(fileMetadata[0], "/sbin/") {
+			fileMetadata[0] = strings.Replace(fileMetadata[0], "/sbin/", "/usr/bin/", 1)
+		} else if strings.HasPrefix(fileMetadata[0], "/lib64/") {
+			fileMetadata[0] = strings.Replace(fileMetadata[0], "/lib64/", "/usr/lib64/", 1)
+		} else if strings.HasPrefix(fileMetadata[0], "/lib/") {
+			fileMetadata[0] = strings.Replace(fileMetadata[0], "/lib/", "/usr/lib/", 1)
+		} else if strings.HasPrefix(fileMetadata[0], "/usr/sbin/") {
+			fileMetadata[0] = strings.Replace(fileMetadata[0], "/usr/sbin/", "/usr/bin/", 1)
+		}
```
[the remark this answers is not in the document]
phmccarty: These are "special" path->path mappings to accommodate the state of various Clear Linux packages as a stop-gap until those packages are fixed.

For example: the glibc-utils subpackage installs `/usr/sbin/zic`, but after the `/usr/sbin` symlink is followed during a dnf install, mixer will see `/usr/sbin/zic`. In the future, the Clear Linux team should fix up the glibc package %files sections and other packages that need similar treatment. In this case, we really want glibc-utils to install `/usr/bin/zic` instead.

builder/build_validate.go
jwakre: RPMs that only exist locally would be fine. Would you prefer that I re-enable the local repo and change my warning to identify possible errors with RPM naming conflicts?

builder/build_validate.go
jwakre: Inserting unique files wouldn't cause any problems for MCA because the bundle manifests and MoM would be unchanged (full manifest is not checked by MCA).

Altering files that exist in bundle manifests may be overlooked by MCA because it doesn't directly compare package content against manifest content. MCA confirms that every package file change has a corresponding manifest file change, but the content of the changes is left unchecked. So, if a specific file in a manifest is expected to change, MCA will succeed as long as that file is modified in any way.

jwakre:
@rchiossi Thanks for the feedback! I'll address it in the next revision.

jwakre:
Added tests to the PR. I will update the PR with review feedback after the 2nd wave of commits has a chance to be reviewed.

jwakre:
@rchiossi 
>What about RPMs that only exist locally? They wouldn't override an upstream RPM but they also won't be available for download.

This is true, but there is a workaround. Another local repo could be added to the .yum-mix.conf file.

Would you prefer the local repo to be re-enabled? In this case, MCA will fail when the local RPM names conflict with upstream (unless the user hacks at the repo to support release versions).

builder/build_validate.go
rchiossi: At what point during clear release is MCA intended to be run? If it was before the files are pushed to the cdn then we might need local repos enabled. If not, I believe it is ok to keep it disable for now if that is the current behavior of the script we are replacing.

jwakre: Yes you are right, I need to re-enable local repos. During the Clear release, it will be run before the files are pushed to the cdn.

builder/build_validate.go
```diff
@@ -255,46 +264,78 @@ func mcaPkgInfo(mInfo map[string]*swupd.Manifest, b *Builder, version, downloadR
...
-		var pkgList = []string{}
-		for pkg := range m.BundleInfo.AllPackages {
-			pkgList = append(pkgList, pkg)
-		}
-
-		out, err := downloadRpms(packagerCmd, pkgList, buildVersionDir, downloadRetries)
-		if err != nil {
+		select {
+		case mCh <- pInfo[m.Name]:
+		case err = <-errorCh:
```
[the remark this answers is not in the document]
jwakre: I could, but I'm not blocking on a close signal and typically examples from the Mixer code base just let the garbage collector handle closing the error channel.

builder/build_validate.go
```diff
@@ -255,46 +264,78 @@ func mcaPkgInfo(mInfo map[string]*swupd.Manifest, b *Builder, version, downloadR
...
+			bundlePkgInfo := pkgInfoFromNoopInstall(out.String())
+			for _, p := range bundlePkgInfo {
+				mux.Lock()
+				if pkgCache[p.name] == nil {
+					bundlePkgInfo[p.name].files, err = resolvePkgFiles(p, b, version)
+
+					// Sending the err over the channel outside of the for loop prevents
+					// blocking on the channel.
+					if err != nil {
+						resolveErr = err
```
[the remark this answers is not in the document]
jwakre: Ya, I will change this

builder/build_validate.go
jwakre: Thinking about this further, MCA would have issues getting the correct bundle definitions for the release process. Mixer hard codes the upstream bundle definition URL to the clr-bundles repo on Github which will not contain pre-release bundle definitions. This could be fixed by supporting a configurable repo for upstream bundle definitions.

Also, MCA's downstream support for RPM repos and bundle definitions is limited. Neither of these easily supports multiple mix versions.

jwakre: I will add a warning message when local RPMs are enabled.

builder/build_validate.go
```diff
@@ -254,46 +263,74 @@ func (b *Builder) mcaPkgInfo(mInfo map[string]*swupd.Manifest, version, download
...
+			if err != nil {
+				errorCh <- err
+				wg.Done()
+				return
+			}
+
+			// Collect metadata to resolve installed RPM file names
+			bundlePkgInfo := pkgInfoFromNoopInstall(out.String())
+			for _, p := range bundlePkgInfo {
+				mux.Lock()
```
rchiossi: Taking another look at this lock block, noticed that if one of the goroutines hit a package that is not yet part of the cache, all the other goroutines will be locked, even the ones trying to access packages that are already cached.
To prevent that, we could change to a RWLock, which would allow us to run `resolvePkgFiles` outside the lock context and let the other goroutines proceed. The downside of this approach is that while a goroutine is not done processing a package, another one might start the processing the same package in parallel. In that case, when we acquire the write lock we need to reevaluate `if pkgCache[p.name] == nil` before updating `pkgCache`.
All of this is under the assumption that  `resolvePkgFiles` is a costly operation given that it relies on `exec()`. It might be nice to compare both to have an idea of the impact in the overall time.

jwakre: I'll investigate this.

jwakre:
@rchiossi I updated the PR with the following to address your feedback:

* Re-enabled local RPMs and updated the warning message
* Stored parseNoopInstall package data inside a struct
* Removed '//removed' comment and TODO statement
* Changed 'mcaInfo' local variable name so it is not identical to function name
* Changed getUpstreamVersion to getLocalUpstreamVersion
* Check variable length instead of recalculating: len(strings.Split(out.String(), "\n"))-1
* Changed functions to *Builder receivers
* Removed extra newline
* Removed file name checks for isFileMod/isManMod
* Created isBundleMod
* Added a warning when error messages are truncated to 50
* Added more description to the help messages
* End Goroutine when an error occurs
* Close error channel

The local RPMs can still skew MCA results, but a warning message is displayed. This can be worked around by formatting the local the RPM repo directory structure to support release versions.

Also, I still need to take a look at RWLocks

rchiossi:
Thanks for the changes. Overall the patch is good to go. The RWLock is a suggestion for performance improvement, but the code as is is correct as well. It would be nice if you could give it a try. If not, I'll merge it as is.

jwakre:
@rchiossi I updated the PR with the RWLock change. After a few quick timing tests, it seems to reduce a significant amount of time.

builder/build_validate.go
```diff
@@ -254,46 +263,82 @@ func (b *Builder) mcaPkgInfo(mInfo map[string]*swupd.Manifest, version, download
...
+				errorCh <- err
+				wg.Done()
+				return
+			}
+
+			// Collect metadata to resolve installed RPM file names
+			bundlePkgInfo := pkgInfoFromNoopInstall(out.String())
+			for _, p := range bundlePkgInfo {
+				// Resolve files for package when it doesn't exist in the cache
+				rw.RLock()
```
rchiossi: instead of locking the entire if block, we could do:
```
rw.RLock()
name := pkgCache[p.name]
rw.RUnlock()

if name == nil {
...
} else {
    m.allPkgs[p.name] = name
}
```

This way `resolvePkgFiles()` will always run outside the lock context and won't prevent an write lock to be acquired if  this goroutine is stopped when exec() is called.

builder/build_validate.go
```diff
@@ -254,46 +263,82 @@ func (b *Builder) mcaPkgInfo(mInfo map[string]*swupd.Manifest, version, download
...
+				} else {
+					m.allPkgs[p.name] = pkgCache[p.name]
+				}
+				rw.RUnlock()
+
+				rw.Lock()
+				if pkgCache[p.name] == nil {
+					pkgCache[p.name] = p
+				}
+				rw.Unlock()
```
rchiossi: by implementing my suggestion above, this block can move inside the if block above since it wont be inside a lock anymore.

builder/build_validate.go
jwakre: Good idea! I pushed the change.

builder/build_validate.go
```diff
@@ -254,46 +263,81 @@ func (b *Builder) mcaPkgInfo(mInfo map[string]*swupd.Manifest, version, download
...
+
+				// Resolve files for package when it doesn't exist in the cache
+				if name == nil {
+					p.files, err = b.resolvePkgFiles(p, version)
+					if err != nil {
+						errorCh <- err
+						wg.Done()
+						return
+					}
+					rw.Lock()
```
rchiossi: you still need to check if `pkgCache[p.name] == nil` here before writing since another goroutine could have filled this entry before this one has finished. This is not a big deal since the information should be identical, but I would at least add a comment here explaining why the override is not a TOCTOU error.

builder/build_validate.go
```diff
@@ -254,46 +263,81 @@ func (b *Builder) mcaPkgInfo(mInfo map[string]*swupd.Manifest, version, download
...
+					pkgCache[p.name] = p
+					rw.Unlock()
```
rchiossi: it is better to do `m.allPkgs[p.name] = p` here and keep the code below inside the else block. This way you prevent releasing the lock just to reacquire it right after. Since `p` is a local variable, the assignment can be done out of a lock (and also outside the if block i suggested above if you decide to use it).

builder/build_validate.go
jwakre: Added this check

builder/build_validate.go
jwakre: Re-added the else statement. Also, I removed the name variable and reduced the number of rw.RLock() calls to one.

rchiossi:
thanks for the changes! =)

Issue #550, opened 2019-02-27
mixer init --clear-version may need a change in the wording


LibertadCruz:
So, I have read some issues like (#153 #123 ) merged about the --clear-version flag and the default that it should be assigned. I read that before it accepted integer and the default was 1 but now it says it accepts strings and the default is "latest". 

The thing is that now as is, I as a user am tempted to introduce something like:

mixer init --clear-version 28000s  

and that would fail with a message: 

Error: invalid argument "28000s" for "--clear-version" flag: strconv.Atoi: parsing "28000s": invalid syntax

As I user I understand that according to the "mixer init -h"  I can introduce a string other than latest and it should not fail.  Plus the man mixer.init does not specify if it can be string or int.

I think idealy a user should be restricted and limited in his input to any option. Ideally it should be by code.

LibertadCruz:
so, my point is documentation the wording to let the user know that it should be a valid clear version  and it should be integer

reaganlo:
Has been fixed.

PR #556, opened 2019-03-04
cmd: Improve error message for invalid upstream

rchiossi:
The `--clear-version` flag in mixer init accepts a string as parameter
to allow the user to set it to `latest`, which will auto fetch the
upstream version. For all other cases, this flag must be provided an
integer number. This patch clarifies the valid values for this flag both
in the error message and in the help description.

### DIFF
--- a/docs/mixer.init.1.rst
+++ b/docs/mixer.init.1.rst
@@ -38,7 +38,7 @@ more details), the following options are recognized.
 
 -  ``--clear-version {version}``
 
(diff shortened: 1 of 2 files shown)


rchiossi:
This PR addresses issue #550 

@LibertadCruz Can you take a look at this change?

jwakre:
@rchiossi I think there is a corresponding change required for the man page.

LibertadCruz:
nice, I like it.

Issue #560, opened 2019-03-20
Support OS identification modification for downstream build


sweeaun:
Suggest to create a new section in builder.conf to support OS identification data for downstream Clear Linux build. Allow Mixer be more friendly to downstream Clear Linux build by allowing custom modification on os-release.

Current default Clear Linux filesystem os-release is configured as below. Only VERSION_ID will be used during mixer version build which it store the mixversion value, the rest will be remain the same. 

NAME="Clear Linux OS"
VERSION=1
ID=clear-linux-os
ID_LIKE=clear-linux-os
VERSION_ID=1
PRETTY_NAME="Clear Linux OS"
ANSI_COLOR="1;35"
HOME_URL="https://clearlinux.org"
SUPPORT_URL="https://clearlinux.org"
BUG_REPORT_URL="mailto:dev@lists.clearlinux.org"
PRIVACY_POLICY_URL="http://www.intel.com/privacy"

Suggest a new section [Os] can be added into builder.conf configuration file.
New section allow user to configure NAME, VERSION, ID, ID_LIKE, PRETTY_NAME, ANSI_COLOR, HOME_URL, SUPPORT_URL, BUG_REPORT_URL and PRIVACY_POLICY_URL for downstream Clear Linux build if user provide the customized value in builder.conf. Take note that new section will not modify the VERSION_ID since it will be used by Mixer for mixverion. 

Example for [Os] section in builder.conf that change NAME, VERSION and ID:

[Os]
 NAME = “Clear Linux OS – Customized ”
 VERSION = “2”
 ID = “clear-linux-os-customized ”

reaganlo:
@rchiossi Is this still high priority?

rchiossi:
This has already been integrated. It is the `OS_RELEASE_PATH` option introduced in config version 1.1

PR #563, opened 2019-04-01
config: Allow OS identification customization #560

sweeaun:
New [Os] section is added in builder.conf file to support
OS identification modification such as NAME, VERSION, ID,
ID_LIKE, PRETTY_NAME, ANSI_COLOR, HOME_URL, SUPPORT_URL,
BUG_REPORT_URL and PRIVACY_POLICY_URL for downstream Clear Linux
build.

This section NOT supporting VERSION_ID since it has been used by
Mixer for mixversion.

Example to modify os-release file:
Change the NAME, PRETTY_NAME and PRIVACY_POLICY_URL

[Os]
  NAME = "Clear-Linux-A"
  PRETTY_NAME = "A Clear Linux OS"
  PRIVACY_POLICY_URL = ""

Example to keep value based on default os-release file:

[Os]
  ANSI_COLOR = "<NoChange>"
  HOME_URL = "<NoChange>"
  SUPPORT_URL = "<NoChange>"
  BUG_REPORT_URL = "<NoChange>"

Signed-off-by: sweeaun <swee.aun.khor@intel.com>

### DIFF
--- a/bat/tests/customize-os-release/run.bats
+++ b/bat/tests/customize-os-release/run.bats
@@ -11,8 +11,7 @@ setup() {
 
   mixer init > /dev/null
 
-  #Backup orig config and use modified config which has customized os-release file
-  mv builder.conf builder-orig.conf
+  #Use modified config which has customized os-release file
   cp files/my-builder.conf builder.conf
   sed -i "s:/home/clr/mix:$PWD:g" builder.conf
 
--- a/bat/tests/customize-os-release/run.bats
+++ b/bat/tests/customize-os-release/run.bats
@@ -1,0 +1,32 @@
+#!/usr/bin/env bats
+
+# shared test functions
+load ../../lib/mixerlib
+
+setup() {
+  global_setup
+}
+
+@test "Customize os-release file" {
+
+  mixer init > /dev/null
+
+  #Backup orig config and use modified config which has customized os-release file
+  mv builder.conf builder-orig.conf
+  cp files/my-builder.conf builder.conf
+  sed -i "s:/home/clr/mix:$PWD:g" builder.conf
+
+  #Mixer build bundle
+  mixer-build-bundles > $LOGDIR/build_bundles.log
+
+  #Compare the diff between customized os-release file and output of mixer build
+  #bundle. The only difference should be VERSION_ID 
+  run bash -c "diff files/my-os-release update/image/10/full/usr/lib/os-release | wc -l "
+  [ "$output" -eq  4 ]
+
+  run bash -c "diff files/my-os-release update/image/10/full/usr/lib/os-release"
+  [[ ${lines[1]} =~ "VERSION_ID" ]]
+  [[ ${lines[3]} =~ "VERSION_ID" ]]
+}
+
+# vi: ft=sh ts=8 sw=2 sts=2 et tw=80
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -380,7 +381,7 @@ func initRPMDB(chrootDir string) error {
 	)
 }
 
-func buildOsCore(packagerCmd []string, chrootDir, version string) error {
+func buildOsCore(b *Builder, packagerCmd []string, chrootDir, version string) error {
 	err := initRPMDB(chrootDir)
 	if err != nil {
 		return err
@@ -394,7 +395,7 @@ func buildOsCore(packagerCmd []string, chrootDir, version string) error {
 		return err
 	}
 
-	if err := fixOSRelease(filepath.Join(chrootDir, "usr/lib/os-release"), version); err != nil {
+	if err := fixOSRelease(b, filepath.Join(chrootDir, "usr/lib/os-release"), version); err != nil {
 		return errors.Wrap(err, "couldn't fix os-release file")
 	}
 
@@ -551,7 +552,7 @@ func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVers
 		// special handling for os-core
 		if bundle.Name == "os-core" {
 			fmt.Println("... building special os-core content")
-			if err := buildOsCore(packagerCmd, fullDir, version); err != nil {
+			if err := buildOsCore(b, packagerCmd, fullDir, version); err != nil {
 				return err
 			}
 		}
@@ -877,7 +878,40 @@ func createVersionsFile(baseDir string, packagerCmd []string) error {
 	return w.Flush()
 }
 
-func fixOSRelease(filename, version string) error {
+func fixOSRelease(b *Builder, filename, version string) error {
+
+	//Replace the default os-release file if customized os-release file found
+	path := b.Config.Mixer.OSReleasePath
+	if path != "" {
+
+		// Validate file path
+		_, err := os.Stat(path)
+		if err != nil {
+			return err
+		}
+
+		fSrc, err := os.OpenFile(path, os.O_RDONLY, 0666)
+		if err != nil {
+			return err
+		}
+
+		fDst, err := os.OpenFile(filename, os.O_RDWR, 0666)
+		if err != nil {
+			return err
+		}
+
+		//Overwrite the os-release file
+		_, err = io.Copy(fDst, fSrc)
+		if err != nil {
+			return err
+		}
+
+		//Close file descriptor to make sure the copy operation is completed.
+		_ = fSrc.Close()
+		_ = fDst.Close()
+
+	}
+
 	f, err := os.Open(filename)
 	if err != nil {
 		return err
@@ -885,13 +919,6 @@ func fixOSRelease(filename, version string) error {
 	defer func() {
 		_ = f.Close()
 	}()
-
-	// TODO: If this is a mix, NAME and ID should probably change too. Create a section in
-	// configuration that will be used as reference to fill this.
-	// TODO: If this is a mix, add extra field for keeping track of the Clear Linux version
-	// used. Maybe also put the UPSTREAM URL, so we are ready to support mixes of mixes.
-	//
-	// See also: https://github.com/clearlinux/mixer-tools/issues/113
 
 	var newBuf bytes.Buffer
 	scanner := bufio.NewScanner(f)
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -5,7 +5,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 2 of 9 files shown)


bat/tests/config-conversion/configs/1.0_builder.conf
```diff
@@ -21,3 +21,15 @@
...
+  NAME = "<NoChange>"
+  VERSION = "<NoChange>"
+  ID = "<NoChange>"
+  ID_LIKE = "<NoChange>"
+  PRETTY_NAME = "<NoChange>"
+  ANSI_COLOR = "<NoChange>"
+  HOME_URL = "<NoChange>"
+  SUPPORT_URL = "<NoChange>"
+  BUG_REPORT_URL = "<NoChange>"
+  PRIVACY_POLICY_URL = "<NoChange>"
```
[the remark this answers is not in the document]
sweeaun: Got it. I will make change accordingly

bat/tests/config-conversion/configs/legacy_builder.conf
```diff
@@ -21,3 +21,15 @@ LOCAL_BUNDLE_DIR=/home/clr/mix/local-bundles
...
+NAME=<NoChange>
+VERSION=<NoChange>
+ID=<NoChange>
+ID_LIKE=<NoChange>
+PRETTY_NAME=<NoChange>
+ANSI_COLOR=<NoChange>
+HOME_URL=<NoChange>
+SUPPORT_URL=<NoChange>
+BUG_REPORT_URL=<NoChange>
+PRIVACY_POLICY_URL=<NoChange>
```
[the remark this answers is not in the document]
sweeaun: Got it. I will make change accordingly

bat/tests/config-conversion/configs/nover_builder.conf
```diff
@@ -20,3 +20,15 @@
...
+  NAME = "<NoChange>"
+  VERSION = "<NoChange>"
+  ID = "<NoChange>"
+  ID_LIKE = "<NoChange>"
+  PRETTY_NAME = "<NoChange>"
+  ANSI_COLOR = "<NoChange>"
+  HOME_URL = "<NoChange>"
+  SUPPORT_URL = "<NoChange>"
+  BUG_REPORT_URL = "<NoChange>"
+  PRIVACY_POLICY_URL = "<NoChange>"
```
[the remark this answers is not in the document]
sweeaun: Got it. I will make change accordingly

config/config_convert.go
```diff
@@ -183,6 +183,17 @@ func (config *MixConfig) parseLegacy() error {
...
+		{`^NAME\s*=\s*`, &config.Os.Name, false},
+		{`^VERSION\s*=\s*`, &config.Os.Version, false},
+		{`^ID\s*=\s*`, &config.Os.ID, false},
+		{`^ID_LIKE\s*=\s*`, &config.Os.IDLike, false},
+		{`^PRETTY_NAME\s*=\s*`, &config.Os.PrettyName, false},
+		{`^ANSI_COLOR\s*=\s*`, &config.Os.AnsiColor, false},
+		{`^HOME_URL\s*=\s*`, &config.Os.HomeURL, false},
+		{`^SUPPORT_URL\s*=\s*`, &config.Os.SupportURL, false},
+		{`^BUG_REPORT_URL\s*=\s*`, &config.Os.BugReportURL, false},
+		{`^PRIVACY_POLICY_URL\s*=\s*`, &config.Os.PrivacyPolicyURL, false},
```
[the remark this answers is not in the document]
sweeaun: Got it. I will make change accordingly

config/config.go
```diff
@@ -108,6 +122,18 @@ func (config *MixConfig) LoadDefaultsForPath(path string) {
...
+	config.Os.Name = "<NoChange>"
+	config.Os.Version = "<NoChange>"
+	config.Os.ID = "<NoChange>"
+	config.Os.IDLike = "<NoChange>"
+	config.Os.PrettyName = "<NoChange>"
+	config.Os.AnsiColor = "<NoChange>"
+	config.Os.HomeURL = "<NoChange>"
+	config.Os.SupportURL = "<NoChange>"
+	config.Os.BugReportURL = "<NoChange>"
+	config.Os.PrivacyPolicyURL = "<NoChange>"
```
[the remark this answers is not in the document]
sweeaun: Initially I thought of put empty strings to indicate no change, but that might trigger another thought what if the user want to overwrite the value to empty instead of no change.

config/config.go
```diff
@@ -34,6 +34,7 @@ type MixConfig struct {
 	Swupd   swupdConf
 	Server  serverConf
 	Mixer   mixerConf
+	Os      osConf
```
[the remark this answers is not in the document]
sweeaun: Sure. I can change it to more specific name related to OsRelease. I will change the section name in the builder.conf as well. Use [OsRelease] instead of [Os].

builder/bundles.go
```diff
@@ -886,20 +892,59 @@ func fixOSRelease(filename, version string) error {
...
+			// Looking for matching
+			if strings.HasPrefix(text, key) {
+
+				// No change if <NoChange> string found
+				if value != "<NoChange>" {
+
+					text = key + "\"" + value + "\""
+				}
+				break
+			}
```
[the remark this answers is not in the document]
sweeaun: Thanks for identify the bug :)

builder/bundles.go
```diff
@@ -886,20 +892,59 @@ func fixOSRelease(filename, version string) error {
...
+		strSplit := strings.Split(strMatch, "=")
+		//Remove space in string
+		strSplit[0] = strings.Replace(strSplit[0], " ", "", -1)
+		strSplit[1] = strings.Replace(strSplit[1], " ", "", 1)
+		//Assign value into map table
+		osID[strSplit[0]] = strSplit[1]
+	}
+
+	// Add VERION_ID into map before os-release file modification is performed
+	osID["VERSION_ID"] = version
```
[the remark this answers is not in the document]
sweeaun: We thought of creating custom os-release file also, just concern that need to create additional new file and thought of utilizing the builder.conf might be a good idea. 

Of course, there are some downsides for using builder.conf such as modification is need if we want to support new OS param and the string manipulation error risk that you mentioned. 
I think the string manipulation error risk can be low if we do code and validate properly.  For new OS param issue, I can add all the supported OS param by https://www.freedesktop.org/software/systemd/man/os-release.html. But that will create a long list of the [OsRelease] list. What do you think? 

I can go alternative if you are not agree with my suggestion.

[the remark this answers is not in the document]
rchiossi: Just adding the values on that link doesn't solve that problem. os-release can contain any key-pair value provided that they respect shell formatting.

The idea is not to create a new file dedicated to set those values, but rather allow the user to provide the path to a file that will be replaced "as-is". For a normal mix, this path would be empty and no file would need to be provided.

Today it is already possible to change os-release file between "build bundles" and "build update". The problem is that unlike other custom files, mixer actually set the version value inside this file during "build bundles", which means that for this manual replacement to work the user must also manually set the version. The idea here is just to facilitate this process, letting mixer set the version on the custom file during "build bundles" step.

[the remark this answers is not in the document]
sweeaun: Okay. I would like to align with you regarding the modification needed for alternative way. 

1. Enable builder.conf to support new OSRelease file as following format. Default is empty mean use the default os-release file. 
[OSRelease]
CUSTOM_OS_RELEASE = ""

2. Change the builder.conf format version from 1.0 -> 1.1. 

3. Replace the os-release file if custom path exist
- Perform file replacement inside fixOSRelease function if custom path provided inside builder.conf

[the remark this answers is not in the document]
rchiossi: > 1. Enable builder.conf to support new OSRelease file as following format. Default is empty mean use the default os-release file.
>    [OSRelease]
>    CUSTOM_OS_RELEASE = ""

We don't need a new section for this. Add a new value to [Mixer] session and call it OS_RELEASE_PATH to keep the naming convention.

> 2. Change the builder.conf format version from 1.0 -> 1.1.

Correct. Conversion should work out of the box for this change, just make sure you add the new config to `bat/tests/config-conversion/configs/` as well. Name it `1.1_builder.conf`. The test automatically run on all configs on this folder.

> 3. Replace the os-release file if custom path exist
> 
> * Perform file replacement inside fixOSRelease function if custom path provided inside builder.conf

You can either do it inside fixOSRelease() or create a new function and call it inside buildOSCore(). Both would work just fine. You will need a reverence to the config in both cases. It is better to pass a reference to the builder object and access it with `b.Config` instead of using the config object directly. This will keep the changes in sync with the rest of the code and facilitate future refactoring.

There is also step 4: create a bat test to test the new feature.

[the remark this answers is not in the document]
sweeaun: Okay. Thanks.

sweeaun:
@rchiossi , I had push the new changes. Please review. Thanks.

bat/tests/customize-os-release/run.bats
```diff
@@ -0,0 +1,32 @@
...
+setup() {
+  global_setup
+}
+
+@test "Customize os-release file" {
+
+  mixer init > /dev/null
+
+  #Backup orig config and use modified config which has customized os-release file
+  mv builder.conf builder-orig.conf
```
rchiossi: No need to backup the original config here.

builder/bundles.go
```diff
@@ -877,7 +878,40 @@ func createVersionsFile(baseDir string, packagerCmd []string) error {
...
+		// Validate file path
+		_, err := os.Stat(path)
+		if err != nil {
+			return err
+		}
+
+		fSrc, err := os.OpenFile(path, os.O_RDONLY, 0666)
+		if err != nil {
+			return err
+		}
```
rchiossi: Use `defer fSrc.Close()` here to prevent file file descriptor leak.

builder/bundles.go
```diff
@@ -877,7 +878,40 @@ func createVersionsFile(baseDir string, packagerCmd []string) error {
...
+
+		fDst, err := os.OpenFile(filename, os.O_RDWR, 0666)
+		if err != nil {
+			return err
+		}
```
rchiossi: Same as above, use `defer fDst.Close()` to prevent leak. Without it, the file descriptors would remain open if the function exits through any error condition.

builder/bundles.go
```diff
@@ -877,7 +878,40 @@ func createVersionsFile(baseDir string, packagerCmd []string) error {
...
+
+		//Overwrite the os-release file
+		_, err = io.Copy(fDst, fSrc)
+		if err != nil {
+			return err
+		}
+
+		//Close file descriptor to make sure the copy operation is completed.
+		_ = fSrc.Close()
+		_ = fDst.Close()
```
rchiossi: Those are not needed if you use `defer` clause.

builder/bundles.go
```diff
@@ -877,7 +878,40 @@ func createVersionsFile(baseDir string, packagerCmd []string) error {
...
+
+	//Replace the default os-release file if customized os-release file found
+	path := b.Config.Mixer.OSReleasePath
+	if path != "" {
+
+		// Validate file path
+		_, err := os.Stat(path)
+		if err != nil {
+			return err
+		}
```
rchiossi: Using Stat here can cause TOCTOU problems. There is no need to stat the file if you are going to open it afterwards since Open will error out jsut like stat if the file doesnt exist. Open by itself can perform both jobs here.

builder/bundles.go
```diff
@@ -877,7 +878,40 @@ func createVersionsFile(baseDir string, packagerCmd []string) error {
...
+
+		//Close file descriptor to make sure the copy operation is completed.
+		_ = fSrc.Close()
+		_ = fDst.Close()
+
+	}
+
```
rchiossi: Even thought the code above looks good, we could improve it a bit. In the code bellow we open os-release file to replace the version number. To do so, we read all the content and rewrite the entire file at the end. As such, instead of replacing the file with the custom one, we can just change where we read the content from. 
If we do so, the whole code above can be removed by replacing the line below with:

```
var err error
var f *os.File
if b.Config.Mixer.OSReleasePath != "" {
    f, err = os.Open(b.Config.Mixer.OSReleasePath)
} else {
    f, err = os.Open(filename)
}
```

bat/tests/customize-os-release/run.bats
sweeaun: I will remove it.

builder/bundles.go
sweeaun: I tried to defer the file close() operation but i always get empty fill for /usr/lib/os-release file after mixer build bundle.
I suspect the copy operation has not been completed at the moment when file opened for VERSION_ID. Thus, I did a manual file close before same file being opened again. 
However, this will be be issue anymore with your suggestion to open different file instead of copy.

builder/bundles.go
sweeaun: No issue anymore since doesn't use copy operation.

builder/bundles.go
sweeaun: No issue anymore since doesn't use copy operation.

builder/bundles.go
sweeaun: No issue anymore since doesn't use copy operation.

builder/bundles.go
sweeaun: Good suggestion and I like this.

sweeaun: @rchiossi , Thanks for your review and I had push new change.

Issue #564, opened 2019-04-08
Delete entries are not generated for removed files in some cases


busykai:
When generating a downstream version, mixer fails to generate delete entries in the bundle manifest under the conditions described in details below.

First of all, the simplest case of updating an RPM with some files removed from it, the delete entries are properly generated:
```
Manifest.os-core:.d..   0000000000000000000000000000000000000000000000000000000000000000        20      /usr/lib/systemd/system/multi-user.target.wants/systemd-networkd.service
Manifest.os-core:.d..   0000000000000000000000000000000000000000000000000000000000000000        20      /usr/lib/systemd/system/sockets.target.wants/systemd-networkd.socket
```
However, when I add an unrelated bundle, edit it, and then re-generate a new version (with --clean), I don't get the delete entries in the manifests.

Specific steps are:
```
$ mixer versions update
$ mixer bundle add NetworkManager
$ mixer bundle edit NetworkManager
$ (edit the bundle to remove some deps, e.g. libX11client)
$ sudo mixer build all --native
```
After that, the Manifests are simply do not contain any mention of the removed files.

I also tested the following three scenarios:

1. (WORKS) If I add the unrelated existing upstream bundle to version 10 (as
   opposed to after upgrading to version 20), the entry does appear.

2. (WORKS) If I add a new bundle (I was using NetworkManager-basics, containing, for example, just NetworkManager package), the delete entry appears.

3. (DOES NOT WORK) If add a new bundle (NetworkManager-basics) and include it in
   NetworkManager (replacing the corresponding entries), the delete entry does
   not appear.

4. (DOES NOT WORK) While trying to go back to 2, I reverted the local copy of
   NetworkManager bundle to be like upstream, but local, the delete entry does
   not appear.

5. (DOES NOT WORK) So I removed local bundle and added pristine upstream again,
   rebuilt, the entry did not appear. Which is weird, because technically, the
   state is like 2.

I always used the same version 10 and was rebuilding the version 20 using "sudo mixer build all --native --clean".

jwakre:
This looks like an issue with the **'--clean'** flag. When rebuilding a version with the --clean flag, the 'previous version' field in the header is incorrect along with other manifest information, like file deletes.

These errors are caused because the --clean flag does not revert the previous version file (update/image/LAST_VER). After manually changing the LAST_VER file, I could successfully rebuild a mix with the --clean flag.

busykai:
[...]

PR #565, opened 2019-04-11
Fix bug with --clean flag

jwakre:
Issue #564 identified an issue with the --clean flag.  

* This change **impacts format bumps**, so we need to be careful that nothing breaks. This PR has been updated so that LAST_VER overrides PREVIOUS_MIX_VERSION, so format bumps should behave similarly.


bat/lib/mixerlib.bash
```diff
@@ -30,10 +30,10 @@ mixer-init-stripped-down() {
 }
 
 mixer-versions-update() {
-  mixer $MIXARGS versions update --mix-version $1 --upstream-version $2 --native=true
+  sudo -E mixer $MIXARGS versions update --mix-version $1 --upstream-version $2 --native=true
 }
 mixer-mixversion-update() {
-  mixer $MIXARGS versions update --mix-version $1 --native=true
+  sudo -E mixer $MIXARGS versions update --mix-version $1 --native=true
```
jwakre: The added dependence on sudo for this command could cause users problems.

rchiossi:
I quite dislike this approach because it changes the meaning on LAST_VER. This was supposed to be the last good bundles build, not the previous. The problem is that if you are rebuilding over a build that succeeded we currently have no track of what the previous one was. On top of that, the added sudo dependency just further emphasize that this is not the best way to track that information.

I believe a cleaner approach would be to use `mixer.state` to track previous version number. We would them update that value when `mixer versions update` is called and use that information with `--clean` instead of LAST_VER. This way we don't add the extra sudo dependency and minimize the impact on code that relies on the current LAST_VER.

Note that the contents of `update/image` are not part of the published content, so we can change it if we see fit. If we could move to use mixer.state only and ditch LAST_VER in the future, it would be even better.

jwakre:
@rchiossi To rework this PR so the previous version is tracked by mixer.state, I plan on making the following changes:

1. Create a ```PREVIOUS``` section inside mixer.state to track the previous version and **increment the minor mixer.state version**. The ```PREVIOUS```  version will be updated by the ```mixer versions update``` command, increment flag, and format bumps.

2. Add a config/state_convert.go file to update the mixer.state file for version changes similarly to config/config_convert.go. This will be useful to insert the ```PREVIOUS``` field into the mixer.state file for users that upgrade Mixer and have an existing mixer.state file.

3. When unset, the  ```PREVIOUS```  field will default to the ```LAST_VER``` and fallback to 0 when  ```LAST_VER``` does not exist.

4. The value of ```PREVIOUS``` will be used to set the previous version for manifests instead of ```LAST_VER```.

5. Format bumps will override the ```PREVIOUS``` field with ```LAST_VER``` so that format bump behavior remains unchanged.

The values of ```PREVIOUS``` and ```LAST_VER``` will be identical except for the following case:
* ```mixer build update``` or ```mixer build all``` was ran without the increment flag
  * Since the ```mixversion``` has not been updated, the value of ```PREVIOUS``` cannot be updated. This resolves the bug with the clean flag.

rchiossi:
> 1. Create a `PREVIOUS` section inside mixer.state to track the previous version and **increment the minor mixer.state version**. The `PREVIOUS`  version will be updated by the `mixer versions update` command, increment flag, and format bumps.

Use a more descriptive name here since it is likely that we will have versions for multiple different things here in the future (e.g. upstreamver and mixver), so it is good to make it explicit that this refers to the mix version.

> The values of `PREVIOUS` and `LAST_VER` will be identical except for the following case:
> 
> * `mixer build update` or `mixer build all` was ran without the increment flag
>   
>   * Since the `mixversion` has not been updated, the value of `PREVIOUS` cannot be updated. This resolves the bug with the clean flag.

They will also be different after `build bundles` and before `build update`. This might be relevant for format bumps, keep this in mind when checking that code.

Other than that, the plan looks good!

jwakre:
I updated the PR with my previously outlined changes. I will continue testing them to confirm everything works as expected.

jwakre:
I've performed additional testing on upstream/downstream format bumps, config conversions, builds, and clean builds. The PREVIOUS_MIX_VERSION field in mixer.state has been adjusted as expected and update content has been generated correctly.

jwakre:
After testing Mixin, the state conversion saves the updated state file in the wrong directory. I will update this PR to address this issue with Mixin.

config/state.go
```diff
@@ -74,6 +77,16 @@ func (state *MixState) loadDefaultFormat() {
 	state.formatSource = "Mixer internal value"
 }
 
+func (state *MixState) loadDefaultPreviousMixVer() {
+	/* The LAST_VER is the default for PREVIOUS_MIX_VERSION */
+	lastVer, err := ioutil.ReadFile("update/image/LAST_VER")
```
jwakre: Typically, mixes are built in the current working directory, so relative paths will usually work. Since Mixin uses an alternative mix path (/usr/share/mix), a path relative to the current directory will likely be incorrect.

Also, there are existing cases on the master branch where state files use relative paths which will cause issues for Mixin.

jwakre:
I updated the PR to make the following for changes to Mixin:

* The value of PREVIOUS_MIX_VERSION is overridden by LAST_VER for Mixin. This keeps Mixin behavior consistent. 
* The working directory for Mixin is changed to the mix directory (/usr/share/mix) before interacting with the mixer.state file. This solves an issue where the mixer.state file was saved to the current working directory instead of the mix directory.

config/state_convert.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/clearlinux/mixer-tools/helpers"
+)
+
+// CurrentStateVersion is the current revision for the state file structure
+const CurrentStateVersion = "1.0"
+
+func (state *MixState) parseVersionAndConvert() error {
+	// Reset version for files without versioning
+	state.version = "0.0"
```
rchiossi: This is not needed for mixer.state because unlike builder.conf, it's file always had the `#VERSION` header, so there is no valid mixer.state file without versioning.

config/state_convert.go
```diff
@@ -0,0 +1,73 @@
...
+	// Already on latest version
+	if found && state.version == CurrentStateVersion {
+		return nil
+	}
+
+	fmt.Printf("Converting state to version %s\n", CurrentStateVersion)
+
+	if err := helpers.CopyFile(state.filename+".bkp", state.filename); err != nil {
+		return err
+	}
```
rchiossi: This is done for builder.conf because it is a file that holds user config. Since mixer.state should only be modified by mixer commands and is considered and "internal storage", there is no need to do a backup.

bat/tests/state-conversion/run.bats
```diff
@@ -0,0 +1,57 @@
...
+  sed -i "s/\(FORMAT\).*/\1 = \"$format\"/" mixer.state
+}
+
+@test "Convert all state versions to latest with version update" {
+   # Verify that all state version conversions work
+  for f in "configs"/*
+  do
+    copy_test_state $f
+
+    mixer versions update --mix-version 20
```
rchiossi: For simple conversion, it would be better to use a command that does not change mixer.state, like `mixer versions --offline` for instance. Since `mixer versions update` will overwrite that value, it doesn't guarantee that the value is properly converted. It doesn't make any difference if you convert from v1.0 to v1.1 since the value is not there, but this wont be true for future versions.

bat/tests/clean-rebuild/run.bats
```diff
@@ -0,0 +1,31 @@
...
+
+  mom="$BATS_TEST_DIRNAME"/update/www/20/Manifest.MoM
+
+  test $(sed -n 's/^version:\t//p' "$mom") -eq 20
+  test $(sed -n 's/^previous:\t//p' "$mom") -eq 10
+  test $(< "$BATS_TEST_DIRNAME"/update/image/LAST_VER) -eq 20
+
+  # Check that PREVIOUS_MIX_VERSION in mixer.state is correct
+  test $(sed -n 's/[ ]*PREVIOUS_MIX_VERSION[ ="]*\([0-9]\+\)[ "]*/\1/p' mixer.state) -eq 10
+}
```
rchiossi: Could we reproduce the bug with the missing entry as part of this test as well?

jwakre:
> I'm still a bit reluctant on the way we rely on `LAST_VER` to set the value for `PREVIOUS_MIX_VER`. The correct value to use would be the version of the last successful `mixer build update` rather than the last successful `mixer build bundles`, but that is a value that we don't keep track.

`LAST_VER` is initialized to 0 by `mixer build bundles`, but otherwise it is updated by `mixer build update` (and format bumps). Although, `LAST_VER` will incorrectly initialize `PREVIOUS_MIX_VER` when `LAST_VER` has been updated, but mixversion has not ( `mixer build update` w/o the increment flag).

> This information is somewhat available in the MoM of the most recent version, but but there is no way to tell if the highest version number in the www folder is actually a good build.
> 
> For now, the patch is good as it is. I'd like to have a follow-up on it though, where we improve our overall version tracking. We should sit down and come up with a plan for it.

Sounds good!

config/state_convert.go
```diff
@@ -0,0 +1,73 @@
...
+
+	fmt.Printf("Old state saved as %s\n", state.filename+".bkp")
+
+	return state.convertCurrent()
+}
+
+func (state *MixState) convertCurrent() error {
```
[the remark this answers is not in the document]
jwakre: Ok, I think consistency makes it easier to understand both configs, so I will keep this the same.

bat/tests/clean-rebuild/run.bats
jwakre: sure!

jwakre:
Thanks for the feedback. I made the following changes:

*  Removed state file version reset (setting to 0.0)
*  Removed backup statefile
*  Replaced `mixer versions update` with `mixer versions --offline`
* Added a negative test case for the clean build bat test

PR #570, opened 2019-05-03
Mca bugfixes

jwakre:
I found additional special case errors for MCA:

1. When a bundle has no packages, MCA should not call dnf without any packages in its argument list.

2. /usr/lib/os-release can be modified by both Mixer and the filesystem package. When Mixer and the filesystem package modified this file, MCA would generate an invalid special case error. This PR reworks how MCA handles /usr/lib/os-release so that an error is not generated when both Mixer and the filesystem package modify the file.

### DIFF
--- a/bat/tests/build-validate/run.bats
+++ b/bat/tests/build-validate/run.bats
@@ -18,7 +18,8 @@ setup() {
 
   # Strip required bundles to minimum package set.
   mixer init --no-default-bundles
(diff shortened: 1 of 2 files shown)


jwakre:
I added another commit so that Mixer will exit with a nonzero value when MCA detects a failure. This is useful for scripts to easily determine that a failure occurred.

PR #571, opened 2019-05-07
update.go: Limit number of compression goroutines

jwakre:
A large number of compression goroutines can cause the system to reach
the open file limit. This change sets an upper bound on the number of
goroutines that can be generated.

### DIFF
--- a/builder/update.go
+++ b/builder/update.go
@@ -106,12 +106,10 @@ func (b *Builder) buildUpdateContent(params UpdateParameters, timer *stopWatch) 
 	}
 
 	var wg sync.WaitGroup
-	// TODO: handle "too many open files" when mom.UpdatedBundles gets too
-	// large. May have to limit the number of workers here.
-	workers := len(mom.UpdatedBundles)
-	wg.Add(workers)
+
+	wg.Add(b.NumBundleWorkers)
 	bundleChan := make(chan *swupd.Manifest)
-	errorChan := make(chan error, workers)
+	errorChan := make(chan error, b.NumBundleWorkers)
 	fmt.Println("Compressing bundle manifests")
 	compWorker := func() {
 		defer wg.Done()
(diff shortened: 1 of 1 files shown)


builder/update.go
```diff
@@ -106,9 +106,14 @@ func (b *Builder) buildUpdateContent(params UpdateParameters, timer *stopWatch)
 	}
 
 	var wg sync.WaitGroup
-	// TODO: handle "too many open files" when mom.UpdatedBundles gets too
-	// large. May have to limit the number of workers here.
+
+	// Large bundle sets can cause the system to have "too many open files",
+	// so the number of workers is limited.
 	workers := len(mom.UpdatedBundles)
+	if workers > 20 {
```
jwakre: The number of updated bundles can be upwards of 800 when building every bundle. I chose 20 because it seemed relatively safe, but I don't have any performance numbers.

jmb82: It seems to me like we should just make sure we're less than the per process number of open files allowed.  Throwing in an arbitrary number doesn't seem right either.  Just my $0.02....

jwakre: Ok, after taking a closer look I can do the following:

* Get max file descriptor limit with Getrlimit
* Get current file descriptor limit with either lsof or looking in /proc
* Calculate number of available descriptors

It doesn't seem like Go has an atomic way to do this, but if no other goroutines run, then additional files shouldn't be opened.

rchiossi: Do we really want to spawn that many go routines? I always thought that this part of the code should use the number of workers set with `--bundle-workers`, the only reason it was never changed is because we didn't have data on performance impact if we followed the change.
If we are limiting it here in any way, we need those numbers. This can heavily impact devops' build time if we are not careful.

jwakre: Using a configurable worker seems like the best way and it is easy to implement. I can collect some performance numbers on this to give a setting recommendation for devops.

mbelluzzo:
So, I did to runs with the exactly same input, starting from scratch. First run is master mixer with NOFILE workaround in the server.

Second run is without NOFILE workaround, mixer + this PR (already using the number of workers instead of 20). The impact is noticeable but minimal.

Screenshot from 2019-05-09 08-42-53

rchiossi:
I believe it is small enough to allow us to merge as is right now and look for improvements afterwards. At least we get a version out that is free from the too many files bug.

PR #572, opened 2019-05-09
swupd: Remove directories from contentsize calculation

### LINKED ISSUE #568: contentsize calculation issue
For two identical mixes, the contentsize manifest header values should always match. However, I am seeing some unexpected mismatches. See steps below.

Steps to reproduce:

1. mkdir mix1; cd mix1
2. mixer init --native --upstream-version=29040
3. mixer bundle add os-clr-on-clr
4. sudo mixer build bundles --native
5. sudo mixer build update --native --skip-fullfiles --skip-packs --no-publish
6. cd ..
7. mkdir mix2; cd mix2
8. sudo cp ../mix1/*.pem .
9. repeat steps 2-5
10. cd ..
11. Check for contentsize differences. I used the script below for this purpose, which ignores several expected differences:. much of the `os-core-update-index` manifest, the `timestamp` header values in all manifests, and the `/usr/share/clear/versionstamp` file from `os-core`.

```
for m in $(awk '/^M./ { if ($4 == "os-core-update-index") next; print $4 }' mix1/update/www/10/Manifest.MoM); do
  out=$(diff -U0 mix{1,2}/update/www/10/Manifest.$m | tail -n +4 | grep -Ev '(^@|[+-]timestamp|/usr/share/clear/versionstamp$)') && echo -e "$m diff:\n$out"
done
```

jwakre:
Directories can be installed with inconsistent sizes. To make mixes
reproducible, the size of directories is omitted from the contentsize
calculation.

Fixes #568 

### DIFF
--- a/bat/tests/contentsize-check/Makefile
+++ b/bat/tests/contentsize-check/Makefile
@@ -1,0 +1,9 @@
+.PHONY: check clean
+
+check:
+	bats ./run.bats
+
+CLEANDIRS = ./update ./test-chroot ./logs ./.repos ./bundles ./update ./mix-bundles ./clr-bundles ./local-yum ./results ./repodata ./local-rpms ./upstream-bundles ./local-bundles
+CLEANFILES = ./*.log ./run.bats.trs ./yum.conf.in ./builder.conf ./mixer.state ./.{c,m}* *.pem .yum-mix.conf mixversion upstreamurl upstreamversion mixbundles
+clean:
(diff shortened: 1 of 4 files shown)


jwakre:
After discussion with @jmb82 and @phmccarty it was decided that it is better to omit directories from the contentsize calculation than to set it to an incorrect value. I will update this PR to skip directories when calculating the contentsize

PR #573, opened 2019-05-09
cmd: skip format bump check

rchiossi:
When doing manual format bumps, the upstream version for +20 needs to be
updated to the upstream version in the new format, but since the format
differs, the build will fail.
This PR adds a new flag to `mixer versions update` and to `mixer build`
subcommands that allows the check this check to be skipped.

### DIFF
--- a/afb.sh
+++ b/afb.sh
@@ -97,7 +97,7 @@ mixer build update --native
 ###############################################################################
 
 # update mixer to build version 30, which in our case is the +20
-mixer versions update --mix-version 30
+mixer versions update --mix-version 30 --skip-format-check
 # update mixer.state to new format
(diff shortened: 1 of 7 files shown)


jwakre:
nit: In the commit message you should pick either `the check` or `this check`

"that allows **the check** **this check** to be skipped."

rchiossi:
It shouldn't affect devops, It is only an issue if you are trying to the update upstream version. Since devops is the mainstream, if never performs the bump check, even without this flag.

Good catch on the commit message, I'll update.

jwakre:
It was an issue for the devops downstream scripts: [clr-distro-factory](https://github.com/clearlinux/clr-distro-factory)

mbelluzzo:
:+1:  :-)

PR #580, opened 2019-05-24
Support "optional" bundles in manifests

castulo:
An optional bundle is similar to the required bundles (included
bundles) but have a couple of differences.

 - Optional bundles will be installed by default when the main
   bundle that includes them is installed.
 - Users will have the option to skip the installation of Optional
   bundles by using the flag --skip-optional/-o with bundle-add.
 - Optional bundles can be removed from the system even if the
   main bundle that installed them is still in the system.

From the perspective of Mixer a bundle will be considered as
optional whenever it is marked as optional in the bundle spec
file. E.g. optional(lib-openssl). And it should be added as such
in the bundle's manifest.

Signed-off-by: Castulo Martinez <castulo.martinez@intel.com>

### DIFF
--- a/builder/bundleset.go
+++ b/builder/bundleset.go
@@ -20,13 +20,13 @@ var (
 )
 
 type bundle struct {
-	Name     string
-	Filename string
-	Header   swupd.BundleHeader
-
-	DirectIncludes []string
-	DirectPackages map[string]bool
-	AllPackages    map[string]bool
+	Name             string
+	Filename         string
+	Header           swupd.BundleHeader
+	DirectIncludes   []string
+	OptionalIncludes []string
+	DirectPackages   map[string]bool
+	AllPackages      map[string]bool
 
 	Files map[string]bool
 }
@@ -40,9 +40,9 @@ type bundleSet map[string]*bundle
 //    bundle is also in the set.
 // 2) Cycle-Free. The set contains no bundle include cycles.
 func validateAndFillBundleSet(bundles bundleSet) error {
-	// Sort the bundles so that all includes appear before a bundle, then
-	// calculate AllPackages for each bundle. Cycles and missing bundles are
-	// identified as part of sorting the bundles.
+	// Sort the bundles so that all includes and optional (also-add) includes appear
+	// before a bundle, then calculate AllPackages for each bundle.
+	// Cycles and missing bundles are identified as part of sorting the bundles.
 	sortedBundles, err := sortBundles(bundles)
 	if err != nil {
 		return err
@@ -97,6 +97,15 @@ func sortBundles(bundles bundleSet) ([]*bundle, error) {
 				if err != nil {
 					return err
 				}
+			}
+			for _, opt := range b.OptionalIncludes {
+				_, exists := bundles[opt]
+				if !exists {
+					return fmt.Errorf("bundle %q includes optional (also-add) bundle %q which is not available", b.Name, opt)
+				}
+				// we don't need to visit optional bundles, they will not be used to determine
+				// the contents of the manifest that includes them and since they are an
+				// independent bundles, they will be validated on their own in a separate step
 			}
 			visiting = visiting[:len(visiting)-1]
 			mark[b] = Visited
@@ -293,7 +302,8 @@ func parseBundle(contents []byte) (*bundle, error) {
 	scanner := bufio.NewScanner(bytes.NewReader(contents))
 
 	var b bundle
-	var includes, packages []string
+	var duplicate bool
+	var includes, packages, optional []string
 
 	line := 0
 	for scanner.Scan() {
@@ -333,6 +343,15 @@ func parseBundle(contents []byte) (*bundle, error) {
 				return nil, fmt.Errorf("Invalid bundle name %q in line %d", text, line)
 			}
 			includes = append(includes, text)
+		} else if strings.HasPrefix(text, "also-add(") {
+			if !strings.HasSuffix(text, ")") {
+				return nil, fmt.Errorf("Missing end parenthesis in line %d: %q", line, text)
+			}
+			text = text[9 : len(text)-1]
+			if !validBundleNameRegex.MatchString(text) {
+				return nil, fmt.Errorf("Invalid bundle name %q in line %d", text, line)
+			}
+			optional = append(optional, text)
 		} else {
 			if !validPackageNameRegex.MatchString(text) {
 				return nil, fmt.Errorf("Invalid package name %q in line %d", text, line)
@@ -345,6 +364,20 @@ func parseBundle(contents []byte) (*bundle, error) {
 		return nil, scanner.Err()
 	}
 
+	// if a bundle was added to both includes and optional
+	// remove it from the optional
+	for _, optionalBundle := range optional {
+		duplicate = false
+		for _, includedBundle := range includes {
+			if optionalBundle == includedBundle {
+				duplicate = true
+				break
+			}
+		}
+		if duplicate == false {
+			b.OptionalIncludes = append(b.OptionalIncludes, optionalBundle)
+		}
+	}
 	b.DirectIncludes = includes
 	b.DirectPackages = make(map[string]bool)
 	for _, p := range packages {
--- a/swupd/bundleinfo.go
+++ b/swupd/bundleinfo.go
@@ -35,13 +35,14 @@ type BundleHeader struct {
 
 // BundleInfo describes the JSON object to be read from the *-info files
 type BundleInfo struct {
-	Name           string
-	Filename       string
-	Header         BundleHeader
-	DirectIncludes []string
-	DirectPackages map[string]bool
-	AllPackages    map[string]bool
-	Files          map[string]bool
+	Name             string
+	Filename         string
+	Header           BundleHeader
+	DirectIncludes   []string
+	OptionalIncludes []string
+	DirectPackages   map[string]bool
+	AllPackages      map[string]bool
+	Files            map[string]bool
 }
 
 // GetBundleInfo loads the BundleInfo member of m from the bundle-info file at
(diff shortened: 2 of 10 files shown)


swupd/bundleinfo.go
```diff
@@ -66,6 +67,9 @@ func (m *Manifest) GetBundleInfo(stateDir, path string) error {
 			return err
 		}
 		m.BundleInfo.DirectIncludes = includes
+		// TODO: this part of the code seems to be a legacy thing. Keep
+		// an eye on this to make sure m.BundleInfo.OptionalIncludes don't
+		// need to be added here too
```
jwakre: MCA is broken by optional includes. To fix it, I think the following must be done:

1. Set `m.BundleInfo.OptionalIncludes`
2. Update MCA pkg/file subtraction to iterate over the optional includes

swupd/manifest.go
```diff
@@ -264,7 +267,7 @@ func ParseManifest(r io.Reader) (*Manifest, error) {
 
 		fields := strings.Split(text, manifestFieldDelim)
 		entry := fields[0]
-		if entry != "includes:" && parsedEntries[entry] > 0 {
+		if !(entry == "includes:" || entry == "optional:") && parsedEntries[entry] > 0 {
```
jwakre: I think the following is easier to read:

if (entry != "includes:" && entry != "optional:") &&........

swupd/format_switch.go
```diff
@@ -34,6 +34,9 @@ timestamp:	{{(.TimeStamp.Unix)}}
 contentsize:	{{.ContentSize -}}
 {{range .Includes}}
 includes:	{{.Name}}
+{{- end -}}
+{{range .Optional}}
+optional:	{{.Name}}
```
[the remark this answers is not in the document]
castulo: I thought the format had to be changed only if the change introduced was a breaking change

swupd/manifest.go
castulo: ok, I'll make the change

swupd/manifest.go
```diff
@@ -658,6 +671,10 @@ func (m *Manifest) subtractManifests(m2 *Manifest) {
 	for _, mi := range m2.Header.Includes {
 		m.subtractManifests(mi)
 	}
+
+	for _, mi := range m2.Header.Optional {
+		m.subtractManifests(mi)
```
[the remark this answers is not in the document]
castulo: so the process of subtracting is only to remove files from a manifest  if they are contained in an included manifest to reduce the size of a manifest file?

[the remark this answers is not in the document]
bryteise: General update speed and size improvements since it also impacts pack creation and pack size. But yes in this case you can not do the subtraction because they are optional and content in the bundle must exist in order to do things like satisfy runtime dependencies.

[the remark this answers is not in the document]
castulo: ok, thanks for the explanation

swupd/bundleinfo.go
castulo: **`m.BundleInfo.OptionalIncludes`** is already being set up in line 81:
`err = json.Unmarshal(biBytes, &m.BundleInfo)`

when doing the unmarshal of the JSON content the OptionalIncludes will be automatically read from the JSON object and added to the BundleInfo struct.

jwakre: Actually, I was wrong above. We don't want the subtraction.

builder/bundleset.go
```diff
@@ -57,6 +57,11 @@ func validateAndFillBundleSet(bundles bundleSet) error {
 				b.AllPackages[k] = v
 			}
 		}
+		for _, optional := range b.OptionalIncludes {
+			for k, v := range bundles[optional].AllPackages {
+				b.AllPackages[k] = v
```
jwakre: Adding optional packages to the all package list for a bundle will insert all of the optional manifest file entries into the original manifest

swupd/bundleinfo.go
jwakre: MCA should be happy if the optionally included packages are removed from the AllPackges list.

builder/bundleset.go
jwakre: We need to make sure that optionally included manifests do not impact the original manifest while also guaranteeing that mixer will build them.

swupd/format_switch.go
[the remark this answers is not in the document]
jwakre: This doesn't change the format, but it does change how the manifest header will be generated based on the value of the current format. It would make sense to me that changes to the templates should only occur in the latest template or a new one should be made. @rchiossi will probably have a preference for this.

castulo:
I think I addressed all your comments @jwakre except for the one about the manifest templates, as we discussed I'll wait to hear @rchiossi 's opinion about it.

jwakre:
While testing this PR, I noticed that removing an optionally included bundle that was not explicitly added to the mix will delete the bundle from the mix which breaks the bundle deletion workflow. This could be resolved by persistently adding the optionally included bundles to the mix.

Also, this is an existing issue for included bundles.

jwakre:
@castulo With the exception of the format_switch, my feedback has been addressed. I'll wait for @rchiossi's feedback on this before approving.

I can also open a separate issue for removed included/optional bundles causing a bundle deletion, but maybe it is a simple fix that can be added to this PR.

swupd/format_switch.go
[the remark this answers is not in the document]
rchiossi: One of the things the `format` tracks is the file format used by manifests. Right now, the correct thing is to bump the format whenever we change how the file looks like. Other than that, we never propagate those changes to previous formats. Here you would need a new template for a new format.

jwakre:
> It seems to be that from mixer's perspective, there is no difference in handling optional bundles and include bundles, other than the fact that they are separated in the manifest.

One important difference is that optional bundles should not be subtracted from the bundles that include them. Since an optional bundle may not be used, it should not alter the contents of other manifests.

> I see a lot of duplication in code, which is not a good sign. I believe a better approach would be to have a flag in the bundle that marks it as an optional include and have direct and optional includes under the same list. This way we could eliminate all the code duplication on bundle handling.

jwakre:
> While testing this PR, I noticed that removing an optionally included bundle that was not explicitly added to the mix will delete the bundle from the mix which breaks the bundle deletion workflow. This could be resolved by persistently adding the optionally included bundles to the mix.
> 
> Also, this is an existing issue for included bundles.

I created issue #592 to document the issue described above. This is an existing issue that can be addressed in a separate PR.

rchiossi:
> One important difference is that optional bundles should not be subtracted from the bundles that include them. Since an optional bundle may not be used, it should not alter the contents of other manifests.

Indeed. But adding to the solution of #592, if an include is permanently added to the mix (and if we do the same for optional includes), there is no needed to iterate the contents of that optional bundle, we just need to add an entry to the header.

castulo:
The update I just pushed doesn't include @rchiossi comment about reducing code duplication, it just changes the header "optional:" with "also-add:" and uses a new template for Manifests with format 29 or higher.

> I see a lot of duplication in code, which is not a good sign. I believe a better approach would be to have a flag in the bundle that marks it as an optional include and have direct and optional includes under the same list. This way we could eliminate all the code duplication on bundle handling.

This comment ^^^ will be addressed in a separate update,

castulo:
@jwakre I think the latest update takes care of  #592, if you have a chance please confirm that is the case. If it is, it now comes down to removing code duplication.

castulo:
@rchiossi I think the PR is working fine now. I addressed all the concerns you and @jwakre had. I didn't have enough time to refactor the code to get rid of duplication. Feel free to work on top of this PR if you want to remove that duplication.

builder/bundleset.go
```diff
@@ -345,7 +364,20 @@ func parseBundle(contents []byte) (*bundle, error) {
...
 	}
 
+	// if a bundle was added to both includes and optional
+	// remove it from the optional
+	for i, optionalBundle := range optional {
+		for _, includedBundle := range includes {
+			if optionalBundle == includedBundle {
+				optional[i] = optional[len(optional)-1]
+				optional[len(optional)-1] = ""
+				optional = optional[:len(optional)-1]
```
jwakre: I don't think that this is safe.

* `optional[i] = optional[len(optional)-1]`
  * The last bundle in the slice is accepted without confirming that it is not in the includedBundle slice.
* `optional[len(optional)-1] = ""`
  * This line can be removed. In the following line, a new slice is created that does not include the underlying array element at index len(optional)-1.

Also, we need to be careful about modifying the variable that we range over. I think `range` creates a copy of the values in the slice and iterates over them, so it will not notice changes to that occur while iterating.

builder/bundleset.go
```diff
@@ -98,6 +98,16 @@ func sortBundles(bundles bundleSet) ([]*bundle, error) {
 					return err
 				}
 			}
+			for _, opt := range b.OptionalIncludes {
+				bundle, exists := bundles[opt]
+				if !exists {
+					return fmt.Errorf("bundle %q includes optional (also-add) bundle %q which is not available", b.Name, opt)
+				}
+				err := visit(bundle)
```
rchiossi: I don't believe we need to visit the optional bundles. They will not be used to determine the contents of the manifest that includes them and since they are an independent bundles, they will be validated on their own in a separate step.

builder/bundleset.go
rchiossi: It is probably better to set the values that needs to be removed to `""` and them trim them out of the slice afterwards. This will also preserve the order.

builder/bundleset.go
castulo: if we remove this part we won't be able to detect cycle errors in also-add, because of this, this test fail TestParseBundleSet (https://github.com/clearlinux/mixer-tools/pull/580/files#diff-a579009ef8fb588be3dec39355f8b2f5R451)

rchiossi: As we discussed on IRC, loops are OK with also-add since there is no dependency among them.

castulo: I removed the part where we were visiting the optional bundles but left the part where it checks for existence of the optional bundle

PR #581, opened 2019-05-28
Add `mixer repo set` command

jwakre:
A generic command for setting values in the .yum-mix.conf file would be useful. Currently, `mixer repo` only has ad-hoc commands to set specific values in the config file.


jwakre:
This PR was made to be used for DevOps scripts, but it is no longer necessary.

PR #583, opened 2019-06-05
MCA: Support repo/baseurl overrides for from and to versions

jwakre:
Problem:

In some cases (downstream mix with custom content) the `from` and `to` mix versions require different repo settings in the dnf conf file to download packages correctly. This makes it difficult for MCA to consume a single config file when downloading packages for both mix versions.

Approach:

Add arguments that override repo/baseurl settings in a temporary dnf config file for that will be used only when downloading packages for the `from` or `to` mixes.

If this approach is acceptable, I will update with tests

### DIFF
--- a/builder/build_validate.go
+++ b/builder/build_validate.go
@@ -2,6 +2,7 @@ package builder
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sort"
@@ -102,7 +103,7 @@ type fileInfo struct {
 // two versions aligns to the corresponding RPM changes. Any mismatched files
 // between the manifests and RPMs will be printed as errors. When there are no
 // errors, package and file statistics for each modified bundle will be displayed.
-func (b *Builder) CheckManifestCorrectness(fromVer, toVer, downloadRetries int) error {
+func (b *Builder) CheckManifestCorrectness(fromVer, toVer, downloadRetries int, fromRepoURLOverrides, toRepoURLOverrides map[string]string) error {
 	if fromVer < 0 || toVer < 0 {
 		return fmt.Errorf("Negative version not supported")
 	}
@@ -114,11 +115,11 @@ func (b *Builder) CheckManifestCorrectness(fromVer, toVer, downloadRetries int) 
 	fmt.Printf("WARNING: Local RPMs will override upstream RPMs for both the from and to versions.\n")
 
 	// Get manifest file lists and subtracted RPM pkg/file lists
-	fromInfo, err := b.mcaInfo(fromVer, downloadRetries)
+	fromInfo, err := b.mcaInfo(fromVer, downloadRetries, fromRepoURLOverrides)
 	if err != nil {
 		return err
 	}
-	toInfo, err := b.mcaInfo(toVer, downloadRetries)
+	toInfo, err := b.mcaInfo(toVer, downloadRetries, toRepoURLOverrides)
 	if err != nil {
 		return err
 	}
@@ -156,7 +157,7 @@ func (b *Builder) CheckManifestCorrectness(fromVer, toVer, downloadRetries int) 
 // mcaInfo collects manifest/RPM metadata and uses it to create a list
 // of manifest files, subtracted packages, and subtracted package files for each
 // bundle in the provided version.
-func (b *Builder) mcaInfo(version, downloadRetries int) (map[string]*mcaBundleInfo, error) {
+func (b *Builder) mcaInfo(version, downloadRetries int, repoURLOverrides map[string]string) (map[string]*mcaBundleInfo, error) {
 	allBundleInfo := make(map[string]*mcaBundleInfo)
 
 	// Get manifest info for valid bundle entries in the MoM.
@@ -166,7 +167,7 @@ func (b *Builder) mcaInfo(version, downloadRetries int) (map[string]*mcaBundleIn
 	}
 
 	// Download and collect metadata for all packages
-	pInfo, err := b.mcaPkgInfo(mInfo, version, downloadRetries)
+	pInfo, err := b.mcaPkgInfo(mInfo, version, downloadRetries, repoURLOverrides)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


jwakre:
Maybe a better solution would be to use a single dnf config file and allow mixer to enable and disable repos on the command line. If there is a way to disable/enable repos separately for the from/to mixes, this would also work. I am open to other suggestions.

builder/repo_control.go
```diff
@@ -224,6 +224,34 @@ func (b *Builder) SetExcludesRepo(reponame, pkgs string) error {
...
+// and overrides repo/URL settings.
+func (b *Builder) OverrideDnfConf(repoURLs map[string]string) (*os.File, error) {
+	// Create tmp dnf config based on original
+	tmpConf, err := ioutil.TempFile(os.TempDir(), "mixerConf")
+	if err != nil {
+		return tmpConf, err
+	}
+	origConf, err := ioutil.ReadFile(b.Config.Builder.DNFConf)
+	if err != nil {
+		return tmpConf, err
```
rchiossi: i don't like the fact that the caller is responsible for closing tmpConf if the function fails here. This can lead to potential leak if this function is ever reused by someone that is not aware of this requirement.

builder/repo_control.go
```diff
@@ -224,6 +224,34 @@ func (b *Builder) SetExcludesRepo(reponame, pkgs string) error {
...
+	}
+	err = ioutil.WriteFile(tmpConf.Name(), origConf, 0644)
```
rchiossi: You already have a file descriptor open for writing, you don't need to open it again here.

mixer/cmd/build.go
```diff
@@ -490,13 +492,62 @@ var buildValidateCmd = &cobra.Command{
...
-		setWorkers(b)
+		setWorkers(fromBuilder)
+
+		toBuilder, err := builder.NewFromConfig(configFile)
+		if err != nil {
+			fail(err)
+		}
+		setWorkers(toBuilder)
+
+		if buildFlags.fromRepoURLs != nil {
```
rchiossi: the code below should be on a function to avoid duplication.

builder/repo_control.go
jwakre: I'll move this to another function.

jwakre:
> Not in the scope of this PR, but to you have an idea of which parts of builder are needed for MCA? I want to measure the effort to isolating that as an initial work on refactoring `builder`.

Variables:
* b.Config.Builder.DNFConf - This can be different for from/to mix versions
* b.Config.Builder.ServerStateDir
* b.NumBundleWorkers
* b.State.Mix.Format

Functions:
* b.getLocalUpstreamVersion()
* b.isPlus10Version()
* b.checkFormatsMatch()
* b.getFormatForVersion()
* b.getLatestForFormat()

jwakre:
> I don't like the fact that we need two build objects here. It is just a good indication that `builder` has accumulated too many functionality and would be better split into dedicated and independent components.

I agree that 2 build objects is not ideal. Instead, I could pass the DNFConf path override directly and use a single build object. Separated components would also be nice.

jwakre:
@rchiossi I refactored the PR so that it no longer relies on 2 builder objects. The other feedback has been addressed too.

builder/bundles.go
```diff
@@ -471,7 +472,7 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
 	args = append(args, rpmList...)
 
 	for attempts := 0; attempts <= maxRetries; attempts++ {
-		out, downloadErr := helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
+		out, downloadErr = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
```
rchiossi: nice catch!

builder/build_validate.go
```diff
@@ -255,14 +257,34 @@ func (b *Builder) mcaPkgInfo(mInfo map[string]*swupd.Manifest, version, download
...
 	}
 
+	if repoURLOverrides != nil {
+		// Create tmp DNF conf with overridden baseurl values for specified repos
+		tmpConf, err := ioutil.TempFile(os.TempDir(), "mixerTmpConf")
+		if err != nil {
+			return nil, err
+		}
+		defer func() {
+			_ = os.Remove(tmpConf.Name())
```
rchiossi: You should also close the file descriptor before removing the file.

mixer/cmd/build.go
```diff
@@ -788,3 +802,23 @@ func checkFormatBump(b *builder.Builder) error {
 
 	return nil
 }
+
+// parseRepoURLOverrides parses the --from-repo-url and --to-repo-url arguments
+// to create a map of repo keys with URL values.
+func parseRepoURLOverrides(repoURLs []string) (map[string]string, error) {
```
rchiossi: Isn't there are facility provided by Cobra to parse a list of args like this?

rchiossi:
One more thing: Remember to update the docs as well on those two new flags.

builder/build_validate.go
jwakre: Is it not safe to rely on the garbage collector for this?

rchiossi: the GC runs concurrent to the main program, so we cannot ensure that the file descriptor will be closed before it is reused in some other function.

mixer/cmd/build.go
jwakre: You are right, the `StringToString` Pflags type can store flag arguments into a map[string]string. The flag format will change to the following:

`--from-repo-url <repo>=<url>`

Our current version of Pflags does not support this feature, so we need to bump the version.

jwakre: Upgrading Cobra from 0.0.1 to 0.0.5 enables the required flags.

jwakre:
I updated the PR with the following changes:

* Changed the underlying Pflag type for the `--from-repo-url/--to-repo-url` args from `StringArrayVar` with `StringToString`. This allows Cobra to populate a map[string]string type. As a result, I no longer need to parse input for these flags.
* Closed the tmp config file descriptor before erasing the file.
* Added documentation.

This PR is currently blocked because a newer version of Pflag is required to use the `StringToString` flag type.

Issue #582, opened 2019-05-30
Performance feature: Parallelize manifest subtraction


gtkramer:
During the minversion for the bump to format 28 (the +20 build), on a machine with 80 vCPUs from 2 Intel Xeon processors, manifest subtraction took 48 minutes with the machine largely idling for most of the time.

If possible, it would be great if this processing could be parallelized to make better use of machine resources and reduce the time it takes for this step of building the update.  This would help DevOps to scale as more bundle are added.

jwakre:
Also, the file subtraction uses a slice data structure, so its deletion operations take O(n). Using another data structure would allow us to achieve O(1). Maybe we could perform file subtraction during `mixer build bundles` and create the bundle info files with a subtracted file list. `mixer build bundles` stores the files in a map which can perform deletions in O(1).

jwakre:
#586 Resolves this issue

By more carefully creating file records, there is no longer a need to perform file subtraction for bundle manifests. Additionally, manifest file records were generated for all subtracted files which is an expensive operation because it involves file hash calculations. A significant amount of time is saved by not calculating file hashes for files that would have been subtracted.

jwakre:
#586 is merged and working well for DevOps. It seems to improve the speed of typical `mixer build update` commands by ~2.5x. Closing this issue.

PR #586, opened 2019-06-16
swupd: Skip included files when creating file records

jwakre:
When creating file records for a manifest, skipping files from the
bundle info file list that exist in one of the manifest's nested
includes removes the need for manifest file subtraction. This improves
performance because it reduces the number of unnecessary file hash
calculations and removes the need for bundle manifest file subtraction.

### DIFF
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -20,11 +20,14 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


jwakre:
Before this change, the "Create Manifests" step took ~8 hours on my 4 core, hyperthreaded machine. With the patch applied, the "Create Manifests" step takes ~3 hours.

jwakre:
This PR closes #582

jwakre:
@rchiossi I found an issue with this PR and MCA. I will investigate further and update the PR.

jwakre:
> @rchiossi I found an issue with this PR and MCA. I will investigate further and update the PR.

I used MCA to test this PR and during my testing I encountered several MCA errors. The root cause of these errors was an MCA bug and not related to this PR.

MCA does not properly subtract files when 2 packages install the same file (this can happen when the same srpm is used to create 2 rpms).

swupd/manifest.go
```diff
@@ -603,6 +606,139 @@ func (m *Manifest) hasUnsupportedTypeChanges() bool {
...
+func (m *Manifest) addManifestFiles(ui UpdateInfo, c config) error {
+	if m.Name == "full" {
+		chroot := filepath.Join(c.imageBase, fmt.Sprint(ui.version), "full")
+		if err := m.addFilesFromChroot(chroot, ""); err != nil {
+			return err
+		}
+	} else {
+		// Add files to manifest that do not exist in included bundles.
+		for f := range m.BundleInfo.Files {
+			if m.isFileIncluded(f) == false {
```
jwakre: The isFileIncluded re-calculates the recursive includes list for each file in the manifest. I am going to replace this function with a function that gets all recursive bundle includes. This can be calculated once and re-used.

jwakre:
The PR was updated to replace the `isFileIncluded` function with the `GetRecursiveIncludes` function. Now, the recursive includes are calculated once for each bundle. The recursive includes list is used to check for files that exist in an included bundle.

Also, the `GetRecursiveIncludes` function can more easily be re-used by MCA to perform file/pkg subtraction.

jwakre:
rebased

PR #590, opened 2019-06-28
MCA Bug Fixes

jwakre:
This PR makes the following changes:

* Serialize MCA package downloads:
  * There was a race condition where packages could be queried by `rpm` before the download finished. Serializing this section does not significantly impact performance.
* Use recursive includes for file/pkg subtraction:
  * There were cases when MCA would miss files/pkgs that should have been subtracted. This issue is resolved by checking each of the bundles from the recursive includes list for files/pkgs to subtract. 
* Use `resolveFileName` function to manage symlinks used by Clear Linux packages.
  * Re-use existing function to remove duplicated code.
* Skip files with blacklisted characters:
  * Mixer does not create entries for files that have names with blacklisted characters (;&|*`/<>\"'). Now, MCA ignores blacklisted files that it resolves from packages. Previously, these were reported as errors by MCA.

### DIFF
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -603,6 +603,29 @@ func (m *Manifest) hasUnsupportedTypeChanges() bool {
 	return false
 }
 
+// GetRecursiveIncludes returns a list of all recursively included bundles
+// for the given manifest.
+func (m *Manifest) GetRecursiveIncludes() []*Manifest {
+	manifests := []*Manifest{}
+	visited := make(map[string]bool)
+
+	for _, inc := range m.Header.Includes {
+		manifests = append(manifests, inc)
+		visited[inc.Name] = true
+	}
+
+	for i := 0; i < len(manifests); i++ {
+		for _, inc := range manifests[i].Header.Includes {
+			if visited[inc.Name] != true {
+				manifests = append(manifests, inc)
+				visited[inc.Name] = true
+			}
+		}
+	}
+
+	return manifests
+}
+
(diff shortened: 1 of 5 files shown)


swupd/manifest.go
```diff
@@ -603,6 +603,29 @@ func (m *Manifest) hasUnsupportedTypeChanges() bool {
 	return false
 }
 
+// GetRecursiveIncludes returns a list of all recursively included bundles
+// for the given manifest.
+func (m *Manifest) GetRecursiveIncludes() []*Manifest {
```
jwakre: This function is also part of #586

jwakre:
> I'm fine with serializing the downloads for now, but it should be possible to download the packages in parallel and process them afterwards once all the downloads are finished, instead of processing them on the go like it is done on the first patch.

I didn't gather a lot of data on this, but DNF should support parallel downloads. From the man page:

```
max_parallel_downloads
Maximum number of simultaneous package downloads. Defaults to 3.
```
Initially, I thought about running DNF in one goroutine that populates a channel with a queue of packages to resolve. Another goroutine could read from the channel and resolve the packages as they are ready. I decided not to do this yet so that I could get a fully functional example.

> 
> Also, what exactly is the problem with symlinks you are trying to solve here?

Clear Linux packages map packages in /bin -> /usr/bin, /sbin  -> /usr/sbin, etc. Previously, I was unaware of an existing Mixer function that performed this mapping, so I wrote my own. This commit replaced my implementation with the Mixer one.

PR #593, opened 2019-07-03
Delete index manifest

jwakre:
This PR deletes the index manifest when transitioning from format 28 -> 29. Since this manifest is directly injected into the mix, it cannot follow the typical bundle deletion workflow. As a result, there are some cases where the deletion will not be smooth:

1. Upstream mixes must create an empty bundle chroot during a format bump from 28 -> 29 for the os-core-update-index manifest to delete it.
2. Downstream mixes ahead of upstream Clear's format (format 28) will miss the index manifest deletion.
3. Downstream mixes that do not upgrade Mixer to include this change when advancing their mix to format 29 will miss the index manifest deletion.

If a mix does not successfully delete the index manifest, the following will happen:
1. The os-core-update-index manifest will be removed from the MoM.
2. The os-core-update-index manifest will be removed without deleting any of its files. Any systems that subscribed to this bundle will need to manually remove its files from their system or perform a `swupd verify --fix --picky`

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -515,18 +515,7 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
 		return err
 	}
 
-	err = ioutil.WriteFile(filepath.Join(bundleDir, bundle.Name), nil, 0644)
-	if err != nil {
-		return nil
-	}
-
-	metaPath := filepath.Join(baseDir, "usr/share/clear/allbundles")
-	err = os.MkdirAll(metaPath, 0755)
-	if err != nil {
-		return err
-	}
-
-	return writeBundleInfoPretty(bundle, filepath.Join(metaPath, bundle.Name))
+	return ioutil.WriteFile(filepath.Join(bundleDir, bundle.Name), nil, 0644)
 }
 
 func clearDNFCache(packagerCmd []string) error {
@@ -592,15 +581,6 @@ func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVers
 	}
 
 	return nil
-}
-
-func writeBundleInfoPretty(bundle *bundle, path string) error {
-	b, err := json.MarshalIndent(*bundle, "", "\t")
-	if err != nil {
-		return err
-	}
-
-	return ioutil.WriteFile(path, b, 0644)
 }
 
 func writeBundleInfo(bundle *bundle, path string) error {
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -280,6 +280,23 @@ var buildFormatOldCmd = &cobra.Command{
 		oldFormat := b.State.Mix.Format
 		b.State.Mix.Format = buildFlags.newFormat
 
+		oldFormatInt, err := strconv.Atoi(oldFormat)
+		if err != nil {
+			fail(err)
+		}
+		newFormatInt, err := strconv.Atoi(b.State.Mix.Format)
+		if err != nil {
+			fail(err)
+		}
+
+		// Delete index manifest when passing format 28
+		if oldFormatInt <= 28 && newFormatInt > 28 {
+			bdir := filepath.Join(b.Config.Builder.ServerStateDir, "image", b.MixVer, "os-core-update-index")
+			if err := os.MkdirAll(bdir, 0755); err != nil {
+				fail(err)
+			}
+		}
+
 		// Build bundles normally. At this point the bundles to be deleted should still
 		// be part of the mixbundles list and the groups.ini
 		if err = buildBundles(b, buildFlags.noSigning, buildFlags.clean, buildFlags.downloadRetries); err != nil {
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -239,6 +239,19 @@ func ParseManifestFile(path string) (*Manifest, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if strings.Compare(m.Name, "MoM") == 0 {
+		m.Type = ManifestMoM
+	} else if strings.Compare(m.Name, "full") == 0 {
+		m.Type = ManifestUnset
+	} else if strings.Contains(m.Name, ".I.") {
+		m.Type = ManifestIterative
+	} else if strings.Contains(m.Name, ".D.") {
+		m.Type = ManifestDelta
+	} else {
+		m.Type = ManifestBundle
+	}
+
 	return m, nil
 }
 
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -828,11 +828,6 @@ func writeIndexManifest(c *config, ui *UpdateInfo, bundles []*Manifest) (*Manife
 		return nil, errors.New("no full manifest found")
 	}
 
-	// construct the tracking files in the bundle and full chroots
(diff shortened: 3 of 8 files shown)


jwakre:
@rchiossi This PR is still WIP and largely untested, but I am open to early feedback to close this PR more quickly. So far I have done the following to smoke test the PR:

```
mixer init
mixer build all --native --increment
mixer build format-bump --native --new-format 29
```

In this example, the index manifest is generated in the first mix, deleted in the +10, and does not exist in the +20.

builder/bundles.go
```diff
@@ -520,13 +520,22 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
...
 
-	metaPath := filepath.Join(baseDir, "usr/share/clear/allbundles")
-	err = os.MkdirAll(metaPath, 0755)
-	if err != nil {
-		return err
-	}
+	// Index manifest content is only generated for formats < 29 and when there
+	// is no bundle chroot.
+	_, err = os.Stat(filepath.Join(buildVersionDir, "os-core-update-index"))
+	if format <= "28" && os.IsNotExist(err) {
```
jwakre: The content for the os-core-update-index manifest is created during `mixer build bundles`. Should format switching in the builder package be centralized like swupd?

mixer/cmd/build.go
```diff
@@ -280,6 +280,14 @@ var buildFormatOldCmd = &cobra.Command{
 		oldFormat := b.State.Mix.Format
 		b.State.Mix.Format = buildFlags.newFormat
 
+		// Delete index manifest when transitioning from format 28 -> 29
+		if oldFormat == "28" {
```
jwakre: This can be more robust: `oldFormat <= 28 && b.State.Mix.Format > 28`

swupd/manifest.go
```diff
@@ -853,6 +856,9 @@ func writeIndexManifest(c *config, ui *UpdateInfo, bundles []*Manifest) (*Manife
 	oldM.sortFilesName()
 	// linkPeersAndChange will update file versions correctly
 	_, _, _ = idxMan.linkPeersAndChange(oldM, ui.minVersion)
+
+	idxMan.Header.FileCount = uint32(len(idxMan.Files))
```
rchiossi: Why do you need to set this again here? Does it get overwritten after line 819?

jwakre: I intended to set the file count on 860 instead of 819, so I will remove the reference on 819.

When a file is deleted `linkPeersAndChange` will add a deleted file entry, but it doesn't update `idxMan.Header.FileCount`, so this value would be incorrect when deleting bundles (it leads to a failure when deleting the index manifest). Also, in the typical, bundle workflow, `idxMan.Header.FileCount` is set after `linkPeersAndChange`

builder/bundles.go
jwakre: At a glance, I think I should be able to consolidate this in one place.

The index manifest entries are generated from the same object that populates the bundle-info files. During `mixer build update`, the bundle-info files are loaded into the `BundleInfo` object. This could be used to generate the index manifest entries.

swupd/manifest.go
```diff
@@ -826,23 +821,31 @@ func writeIndexManifest(c *config, ui *UpdateInfo, bundles []*Manifest) (*Manife
...
-	// if the allbundles directory was created add all those bundle files
-	// to the index as well
-	metaRoot := filepath.Join(bundleDir, "full", indexAllBundleDir)
-	if _, err = os.Stat(metaRoot); err == nil {
-		err = idxMan.addFilesFromChroot(metaRoot, filepath.Join(bundleDir, "full"))
-		if err != nil {
+	// When there is no bundle chroot for the index manifest, create a tracking file and add
+	// the allbundles directory to the index manifest. The index manifest will be deleted when
+	// a bundle chroot directory exists.
+	if _, err := os.Stat(filepath.Join(bundleDir, IndexBundle)); os.IsNotExist(err) {
```
jwakre: When a bundle chroot for the index manifest exists (even if there is content in the bundle chroot), the index manifest will be deleted. Currently, index manifests do not support adding files with bundle chroots, so no functionality is lost. I do not want to enable bundle chroots for the index manifest since it is deprecated in future formats.

jwakre:
The PR has been updated with the following changes:
* Consolidate index manifest generation to the swupd package.
  * Previously, index manifest content was generated and injected into the full chroot during `mixer build bundles`. Now this is performed in format_switch.go.
* Added unit test to validate proper index manifest generation/deletion behavior.
* Updated index manifest code so that it will be deleted when an index manifest bundle chroot exists.

jwakre:
@rchiossi This PR is ready for review again.

@gtkramer When this PR is merged, Mixer will require **upstream mixes to create an empty bundle chroot for "os-core-update-index"** when performing **format bump 28 -> 29** to properly delete the bundle.

jwakre:
@rchiossi I am double checking that DevOps can handle the manual format-bump intervention required for this change. Please hold on merging until they sign off

jwakre:
> @rchiossi I am double checking that DevOps can handle the manual format-bump intervention required for this change. Please hold on merging until they sign off

DevOps has accepted the requirements to delete the index manifest for an upstream mix, so this PR is unblocked.

swupd/manifest.go
```diff
@@ -236,6 +236,12 @@ func ParseManifestFile(path string) (*Manifest, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Only iterative manifests have names with .I.
+	if strings.Contains(m.Name, ".I.") {
+		m.Type = ManifestIterative
+	}
+
```
rchiossi: Unrelated to this patch, but we should probably set the types for everything that we parse. It seems like a bug that fact that this is left unset.

mixer/cmd/build.go
```diff
@@ -280,6 +280,14 @@ var buildFormatOldCmd = &cobra.Command{
 		oldFormat := b.State.Mix.Format
 		b.State.Mix.Format = buildFlags.newFormat
 
+		// Delete index manifest when passing format 28
+		if oldFormat <= "28" && b.State.Mix.Format > "28" {
```
rchiossi: This passed by me unnoticed at first, but you are comparing strings, not ints. You need to convert or this won't work as intended.

swupd/manifest.go
```diff
@@ -882,35 +885,39 @@ func writeIndexManifest(c *config, ui *UpdateInfo, bundles []*Manifest) (*Manife
...
 
-	// done processing, sort by version before writing
-	idxMan.sortFilesVersionName()
-	// check that there is actually a change to this manifest before writing
-	if idxMan.Files[len(idxMan.Files)-1].Version < idxMan.Header.Version {
-		// return the old version of the manifest since there was no change
-		// in this version
-		return oldM, nil
-	}
+	if idxMan.Header.FileCount > 0 {
```
rchiossi: Since you are at it, this will be cleaner if you just return if FileCount <= 0 instead of moving all the code inside a block.

jwakre:
@rchiossi Thanks for the feedback. I rebased this PR and made the following changes to address your review:
* ParseManifest function now sets all manifest types
* Format strings are converted to ints before comparison
* When FileCount == 0, writeIndexManifest will return early

PR #596, opened 2019-07-24
MCA: Parse rpm output with \a characters

jwakre:
Previously, MCA parsed rpm output with commas which is a valid file
character. Now, rpm output is parsed with \a which are
blacklisted file characters for Mixer.

### DIFF
--- a/builder/build_validate.go
+++ b/builder/build_validate.go
@@ -348,7 +348,7 @@ func (b *Builder) mcaPkgInfo(manifests []*swupd.Manifest, version, downloadRetri
 
 // resolvePkgFiles queries a package for a list of file metadata.
 func (b *Builder) resolvePkgFiles(pkg *pkgInfo, version int) ([]*fileInfo, error) {
-	queryCmd := "[%{filenames}, %{filesizes}, %{filedigests}, %{filemodes:perms}, %{filelinktos}, %{fileusername}, %{filegroupname}\n]"
+	queryCmd := "[%{filenames}\a%{filesizes}\a%{filedigests}\a%{filemodes:perms}\a%{filelinktos}\a%{fileusername}\a%{filegroupname}\n]"
 	rpmCmd := []string{"rpm", "-qp", "--qf=" + queryCmd}
 
 	validationDir := filepath.Join(b.Config.Builder.ServerStateDir, "validation")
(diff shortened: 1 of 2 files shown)


jwakre:
MCA was confused by `/usr/lib64/R/library/packrat/tests/testthat/Ugly, but legal, path for a project (long)/bread/DESCRIPTION`.

The other MCA implementation parses on tab characters, but tab characters could be used because they are not blacklisted.

builder/build_validate.go
```diff
@@ -348,7 +348,7 @@ func (b *Builder) mcaPkgInfo(manifests []*swupd.Manifest, version, downloadRetri
 
 // resolvePkgFiles queries a package for a list of file metadata.
 func (b *Builder) resolvePkgFiles(pkg *pkgInfo, version int) ([]*fileInfo, error) {
-	queryCmd := "[%{filenames}, %{filesizes}, %{filedigests}, %{filemodes:perms}, %{filelinktos}, %{fileusername}, %{filegroupname}\n]"
+	queryCmd := "[%{filenames}; %{filesizes}; %{filedigests}; %{filemodes:perms}; %{filelinktos}; %{fileusername}; %{filegroupname}\n]"
```
phmccarty: I think there should be a FIXME here regarding a future action to eventually replace `;` with a control character. The control characters accepted by rpm's queryformat are the following:

* \a
* \b
* \f
* \n
* \r
* \t
* \v

IMO, \a or \b would be most appropriate, since they are not classified as "space" characters. Ideally, \0 could be used, but queryformat interprets `\0` as `0`, so that's not helpful.

jwakre: I can play around a little bit with these. I chose not to use tabs because it would make Mixer's acceptable file names more restrictive.

phmccarty: Yes, that is understandable. However I think the likelihood of `;` appearing in a filename is higher than a literal tab character, so there is a choice of either:

1) using one of the control characters above as the field separator, and also ban that character for all filenames
2) using an already-banned character as the field separator

I would prefer the first approach, but it is debatable.

Another option to avoid the issue entirely would be:

3) reorder the queryformat string so that `%{filenames}` is referenced last. Then you could split each line into a fixed number of fields, and the leftover rightmost field will be `%{filenames}`. Or you could keep the current order, and split from right-to-left. I don't know what's easier with Go :-)

jwakre: >     1. using one of the control characters above as the field separator, and also ban that character for all filenames

The `\a` character is now used for parsing rpm output and has been added to the character blacklist.

PR #597, opened 2019-07-26
travis: Set travis timeout to 60min

jwakre:
The bat tests in travis fail due to timeout when there is no output for
10 minutes. This change adds the `travis_wait` command which will prevent
the timeout by writing to the build log every minute for 60 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,4 @@ before_install:
     - docker build -t testdock .
 
 script:
-    - docker run testdock
+    - travis_wait 60 docker run testdock


rchiossi:
Did we add any new test that takes more than 10 mins to run? I wonder why we are hitting this problem now.

jwakre:
> Did we add any new test that takes more than 10 mins to run? I wonder why we are hitting this problem now.

The timeout has been a problem for me in the past, but restarting the build usually worked around it. I opened this PR because #596 consistently triggers this timeout.

jwakre:
Also, I have noticed that `travis_wait` silences the log output. When the build is completed, suppressed log output is dumped.

jwakre:
After running a number of travis tests, it seems that this `travis_wait` solution is not very robust. I read online articles that `travis_wait` is not functional when the command prefixed by travis_wait creates additional processes.

PR #599, opened 2019-08-13
Return error instead of panic while generating delta.

ashleshaAtrey:

### DIFF
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -580,9 +580,7 @@ func linkDeltaPeersForPack(c *config, oldManifest, newManifest *Manifest) error 
 	}
 
 	// Run rename detection on old and new manifests
-	renameDetection(newManifest, added, removed, *c)
-
-	return nil
+	return renameDetection(newManifest, added, removed, *c)
 }
 
 func includesChanged(m1 *Manifest, m2 *Manifest) bool {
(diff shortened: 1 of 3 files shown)


swupd/manifest.go
```diff
@@ -580,8 +580,10 @@ func linkDeltaPeersForPack(c *config, oldManifest, newManifest *Manifest) error
 	}
 
 	// Run rename detection on old and new manifests
-	renameDetection(newManifest, added, removed, *c)
-
+	err := renameDetection(newManifest, added, removed, *c)
+	if err != nil {
+		return err
+	}
```
rchiossi: you can just call `return renameDetection(...)` instead.

swupd/rename_test.go
```diff
@@ -137,7 +137,7 @@ func TestRename(t *testing.T) {
 	for _, tc := range tests {
 		add := filelist(t, sn, tc.add)
 		remove := markdelete(filelist(t, sn, tc.remove))
-		renameDetection(&Manifest{}, add, remove, config{})
+		_ = renameDetection(&Manifest{}, add, remove, config{})
```
rchiossi: You should check the error and fail the test if it is not `nil`

swupd/manifest.go
ashleshaAtrey: Updated.

swupd/rename_test.go
ashleshaAtrey: Updated.

PR #600, opened 2019-08-15
WIP: Implemented --ignore-missing flag in mixer build

ashleshaAtrey:
if ignore-missing flag is set, deltas are generated for existing files,
ignoring the ones which are not present

### DIFF
--- a/builder/deltapacks.go
+++ b/builder/deltapacks.go
@@ -27,7 +27,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printReport bool, outputDir, bundleDir string, numWorkers int) error {
+func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printReport bool, outputDir, bundleDir string, numWorkers int, ignoreMissing bool) error {
 	timer := &stopWatch{w: os.Stdout}
 	defer timer.WriteSummary(os.Stdout)
 	timer.Start("CREATE DELTA PACKS")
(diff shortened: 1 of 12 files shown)


mixer/cmd/build.go
```diff
@@ -57,6 +57,7 @@ type buildCmdFlags struct {
 	numFullfileWorkers int
 	numDeltaWorkers    int
 	numBundleWorkers   int
+	ignoreMissing      bool
```
rchiossi: This flag should be in `buildDeltaPacksFlags`, not in `buildCmdFlags` as it is only part of the delta-packs command.

mixer/cmd/build.go
```diff
@@ -604,6 +609,8 @@ func runBuildDeltaPacks(cmd *cobra.Command, args []string) error {
 		fail(err)
 	}
 	setWorkers(b)
+	b.IgnoreMissing = buildFlags.ignoreMissing
+
```
rchiossi: This is the only place that this value needs to be set. Delta packs are not generated by the other build commands, including build all.

builder/deltapacks.go
```diff
@@ -75,7 +75,7 @@ func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printRepor
 			defer wg.Done()
 			for b := range bundleQueue {
 				fmt.Printf("  Creating delta pack for bundle %q from %d to %d\n", b.Name, b.FromVersion, b.ToVersion)
-				info, err := swupd.CreatePack(b.Name, b.FromVersion, b.ToVersion, outputDir, bundleDir, numWorkers)
+				info, err := swupd.CreatePack(b.Name, b.FromVersion, b.ToVersion, outputDir, bundleDir, numWorkers, ignoreMissing)
```
rchiossi: You are going too deep into the chain. we either create the whole package or no package at all. If anything fails beyond this point, the creation of that package should fail. As such, there is no need to propagate `ignoreMissing` here.

builder/deltapacks.go
```diff
@@ -75,7 +75,7 @@ func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printRepor
...
 				if err != nil {
 					log.Printf("ERROR: Pack %q from %d to %d FAILED to be created: %s\n", b.Name, b.FromVersion, b.ToVersion, err)
 					// Do not exit on errors, we have logging for all other failures and deltas are optional
```
rchiossi: Also, I just realize from this comment that ignoring the error is the default behavior. Apparently it wasn't working as such because of that bug you fixed last week. 
Since the errors are logged and ignored, it seems the flag may not even be needed. What we need are tests to check if the behavior is correct and there aren't other panic() calls that are preventing it from continuing on failure.

PR #602, opened 2019-08-21
Added test case for generating delta packs with missing files

ashleshaAtrey:
Added testcase for generating deltapacks with missing files for
different versions. If file is missing, delta pack for that
bundle should be ignored as delta pack generation is optional.

### DIFF
--- a/bat/tests/build-multiple-delta-packs/Makefile
+++ b/bat/tests/build-multiple-delta-packs/Makefile
@@ -1,0 +1,9 @@
+.PHONY: check clean
+
+check:
+	bats ./run.bats
+
+CLEANDIRS = ./update ./test-chroot ./logs ./.repos ./bundles ./update ./mix-bundles ./clr-bundles ./local-yum ./results ./repodata ./local-rpms ./upstream-bundles ./local-bundles
+CLEANFILES = ./*.log ./run.bats.trs ./yum.conf.in ./builder.conf ./mixer.state ./.{c,m}* *.pem .yum-mix.conf mixversion upstreamurl upstreamversion mixbundles
+clean:
+	sudo rm -rf $(CLEANDIRS) $(CLEANFILES)
--- a/bat/tests/build-multiple-delta-packs/description.txt
+++ b/bat/tests/build-multiple-delta-packs/description.txt
@@ -1,0 +1,4 @@
+build-multiple-delta-packs
+=====================
+This test case tests and runs 'mixer build delta-packs' command multiple times for different
+versions with some delta files missing when generating content with 'mixer build all'.
--- a/bat/tests/build-multiple-delta-packs/run.bats
+++ b/bat/tests/build-multiple-delta-packs/run.bats
@@ -1,0 +1,25 @@
+#!/usr/bin/env bats
+
+# shared test functions
+load ../../lib/mixerlib
+
+setup() {
+  global_setup
+}
+
+@test "build delta-packs after build all for different versions" {
+  current_ver=$(get-current-version)
+  mixer-init-stripped-down "$current_ver" 10
+  mixer-build-all
+  sudo mixer versions update
+  mixer-build-all
+  mixer-build-delta-packs 10
+  sudo rm -rf ./update/image/10/full/usr/lib/os-release
+  sudo mixer versions update
+  mixer-build-all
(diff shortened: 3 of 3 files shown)


bat/tests/build-multiple-delta-packs/run.bats
```diff
@@ -0,0 +1,38 @@
...
+load ../../lib/mixerlib
+
+setup() {
+  global_setup
+}
+
+@test "build delta-packs after build all for different versions" {
+  current_ver=$(get-current-version)
+  mixer-init-stripped-down "$current_ver" 10
+  sudo -E mixer build all --native
```
rchiossi: use `mixer-build-all` here instead.

bat/tests/build-multiple-delta-packs/run.bats
```diff
@@ -0,0 +1,38 @@
...
+  sudo mixer versions update
+  mixer-build-all --native
```
rchiossi: `mixer-build-all` already has the `--native` flag, so there is no need to add it again.

bat/tests/build-multiple-delta-packs/run.bats
```diff
@@ -0,0 +1,38 @@
...
+  sudo -E mixer build delta-packs --native --previous-versions=10
```
rchiossi: use `mixer-build-delta-packs 10` instead

bat/tests/build-multiple-delta-packs/run.bats
```diff
@@ -0,0 +1,38 @@
...
+  sudo rm -rf ./update/image/10/full/usr/lib/os-release
+  sudo mixer versions update
+  sudo -E mixer build all --native
```
rchiossi: `mixer-build-all`

bat/tests/build-multiple-delta-packs/run.bats
```diff
@@ -0,0 +1,38 @@
...
+  sudo -E mixer build delta-packs --native --previous-versions=10
```
rchiossi: `mixer-build-delta-packs 10`

bat/tests/build-multiple-delta-packs/run.bats
```diff
@@ -0,0 +1,38 @@
...
+  test $(< mixversion) -eq 30
```
rchiossi: Any particular reason why you are testing `mixversion` here?

bat/tests/build-multiple-delta-packs/run.bats
```diff
@@ -0,0 +1,38 @@
...
+
+  var1=`[ "$(ls -A ./update/www/10/delta)" ] && echo "Not Empty" || echo "Empty"`
+  test "$var1" = "Empty"
```
rchiossi: Not necessarily wrong, but since ls will already give you a binary value that will be checked by test, you can you that value directly instead of assigning strings to it. You can use something like `test $(ls -A1 | wc -l) -eq 0` for example.
It is fine as is though,  no need to change if you believe this is easier to read.

bat/tests/build-multiple-delta-packs/Makefile
```diff
@@ -0,0 +1,9 @@
+.PHONY: check clean
+
+check:
+	bats ./run.bats
+
+CLEANDIRS = ./update ./test-chroot ./logs ./.repos ./bundles ./update ./mix-bundles ./clr-bundles ./local-yum ./results ./repodata ./local-rpms ./upstream-bundles ./local-bundles
+CLEANFILES = ./*.log ./run.bats.trs ./yum.conf.in ./builder.conf ./mixer.state ./.{c,m}* *.pem .yum-mix.conf mixversion upstreamurl upstreamversion mixbundles
+clean:
+	sudo rm -rf $(CLEANDIRS) $(CLEANFILES)
```
rchiossi: add a new line here.

bat/tests/build-multiple-delta-packs/description.txt
```diff
@@ -0,0 +1,4 @@
+build-multiple-delta-packs
+=====================
+This test case tests and runs 'mixer build delta-packs' command multiple times for different versions with some delta files missing when generating
+content with 'mixer build all'.
```
rchiossi: line split is not in the right place here.

PR #603, opened 2019-08-27
Signing the latest file in version/formatXX dir

ashleshaAtrey:
The latest file in version/formatXX isn't signed by mixer.
Mixer signs the latest file so clients could check the signature of that file too.

Signed-off-by: Ashlesha Atrey ashlesha.atrey@intel.com

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -370,16 +370,35 @@ func (b *Builder) BuildUpdate(params UpdateParameters) error {
 	}
 
 	fmt.Printf("Setting latest version to %s\n", b.MixVer)
-
-	err = ioutil.WriteFile(filepath.Join(b.Config.Builder.ServerStateDir, "www", "version", "latest_version"), []byte(b.MixVer), 0644)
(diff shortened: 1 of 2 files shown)


builder/builder.go
```diff
@@ -380,6 +381,12 @@ func (b *Builder) BuildUpdate(params UpdateParameters) error {
 	if err != nil {
 		return errors.Wrapf(err, "couldn't update the latest version")
 	}
+
+	err = b.signLatestVersion(formatDir)
```
rchiossi: You need to check if `params.SkipSigning == false` before signing the file.

builder/builder.go
```diff
@@ -743,3 +750,23 @@ func (b *Builder) BuildDeltaManifestsPreviousVersions(prev, to uint32) error {
 
 	return nil
 }
+
+func (b *Builder) signLatestVersion(formatdir string) error {
```
rchiossi: other than the file name, this is an exact copy of ` SignManifestMoM()`. Instead of duplicating the code, create a function that can be used for both. Also, keep the code in `update.go`.

PR #604, opened 2019-08-27
Added zstd compression support for fullfile creation

ashleshaAtrey:
Currently, fullfile creation supports gzip, bzip2, and xz compression types.
Adding zstd support, since it tends to produce smaller files in general and decompression speed is excellent.


ashleshaAtrey:
Selective Compression #637 solves the issue.

PR #608, opened 2019-09-10
WIP: Mixer reset

ashleshaAtrey:
It reverts the mix to a good state in case of a build failure or in case the user wants to roll back to a previous version.
By default, the command will  bring back the mix to the same state it have when it last build a successful mix.
This value can be overridden if a `--to` flag is provided with a version number.The command will not be destructive unless a `--clean` flag is provided specified. If so, mixer will delete all files associated with versions that are bigger than the one provided. If not, only the values of the state files will change, but the files will be kept.


bat/tests/reset-command/run.bats
```diff
@@ -0,0 +1,34 @@
...
+  global_setup
+}
+
+@test "reset" {
+  mixer-init-stripped-down 27310 10
+  mixer build all --format 30 --native
+  mixer versions update --upstream-version 30000
+  mixer build upstream-format --new-format 28 --native
+  sudo mixer versions update --upstream-version 30800
+  mixer build upstream-format --new-format 20 --native
```
rchiossi: Any reason why you are doing format bumps in this test?

builder/builder.go
```diff
@@ -369,34 +369,9 @@ func (b *Builder) BuildUpdate(params UpdateParameters) error {
 		return nil
 	}
 
-	fmt.Printf("Setting latest version to %s\n", b.MixVer)
```
rchiossi: This is not part of this PR. Can you rebase and update?

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
+// Copyright © 2018 Intel Corporation
```
rchiossi: 2019

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+	"strings"
+
+	"github.com/clearlinux/mixer-tools/builder"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var resetCmd = &cobra.Command{
+	Use:   "reset",
+	Short: "Revert the mix to a good state",
```
rchiossi: I'd say `Revert mix to a previous version` since we are not actually validating it to ensure it is a good one.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+	Long: `Revert the mix to a good state in case of a build 
+	failure or in case the user wants to roll back to a previous
```
rchiossi: I believe it is better to tell what the command does rather than telling when it should be used.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/spf13/cobra"
+)
+
+var resetCmd = &cobra.Command{
+	Use:   "reset",
+	Short: "Revert the mix to a good state",
+	Long: `Revert the mix to a good state in case of a build 
+	failure or in case the user wants to roll back to a previous 
+	version`,
+	Run: runReset,
```
rchiossi: We should use `RunE` here instead. Commands that are currently using `Run` are outdated and should be updated.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+	RootCmd.AddCommand(resetCmd)
+
+	resetCmd.Flags().Int32Var(&resetCmdFlags.toVersion, "to", -1, "Reset to a specific mix version, default = PREVIOUS_MIX_VERSION")
+	resetCmd.Flags().BoolVar(&resetCmdFlags.clean, "clean", false, "Deletes all files with versions bigger than the one provided")
+}
+
+func runReset(cmd *cobra.Command, args []string) {
+	b, err := builder.NewFromConfig(configFile)
+	if err != nil {
+		fail(err)
```
rchiossi: if you use `RunE`, this becomes `return err`

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	// assuming mixer.state file has the correct info
+	stateMixPreviousMixVersionUint32, err := parseUint32(b.State.Mix.PreviousMixVer)
```
rchiossi: you should check `--to` flag first and only set this value if it was not provided.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+		b.State.Mix.PreviousMixVer = strconv.Itoa(int(resetCmdFlags.toVersion))
+		stateMixPreviousMixVersionUint32 = uint32(resetCmdFlags.toVersion)
+	}
+
+	// Make sure FORMAT in mixer.state has the same value as update/www/<previousMixVer>/format
+	var format []byte
+	filename := filepath.Join(b.Config.Builder.ServerStateDir, "www", b.State.Mix.PreviousMixVer, "format")
+	if format, err = ioutil.ReadFile(filename); err != nil {
+		fail(err)
+	} else {
```
rchiossi: No need for an `else` here. The flow will stop if the `if` is true.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+		if strings.TrimSpace(string(format)) != b.State.Mix.Format {
```
rchiossi: There is no need to check if they are the same here, just assign the value directly.  This will save you a call to `strings.TrimSpace`

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+			b.State.Mix.Format = strings.TrimSpace(string(format))
+		}
+	}
+
+	// Change upstreamVersion file content to the previous version
+	var lastStableMixUpstreamVersion []byte
+	filename = filepath.Join(b.Config.Builder.ServerStateDir, "www", b.State.Mix.PreviousMixVer, "upstreamver")
+	if lastStableMixUpstreamVersion, err = ioutil.ReadFile(filename); err != nil {
+		fail(err)
+	} else {
```
rchiossi: Same as above. There is no need for an `else` here.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+				fail(errors.Wrapf(err, "Couldn't parse upstream version"))
+			}
+			vFile := filepath.Join(b.Config.Builder.VersionPath, b.UpstreamVerFile)
+			if err = ioutil.WriteFile(vFile, []byte(b.UpstreamVer), 0644); err != nil {
+				fail(err)
+			}
+		}
+	}
+
+	// Change mixVersion to point to the last good build
```
rchiossi: I'd say `to PreviousMixVer` rather than `last good build`

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+	if err := ioutil.WriteFile(filepath.Join(b.Config.Builder.VersionPath, b.MixVerFile), []byte(b.State.Mix.PreviousMixVer), 0644); err != nil {
+		fail(err)
+	}
+	b.MixVer = b.State.Mix.PreviousMixVer
+	b.MixVerUint32, err = parseUint32(b.MixVer)
+	if err != nil {
+		fail(errors.Wrapf(err, "Couldn't parse mixer version"))
+	}
+
+	// Make sure update/image/LAST_VER points to the last good build
```
rchiossi: same as above.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	// update the state.mix file
+	err = b.State.Save()
+	if err != nil {
+		failf("Couldn't update mixer.state file: %s", err)
+	}
+
+	// update www/version/latest_version file and check if sig file exists, if it does, regenerate the signature
+	// update www/version/formatXX/latest file and update the content, if signature exits , regenerate the signature
```
rchiossi: Now, this is a tricky one. `latest_version` is a published file. If we touch it, we are actually altering the content, which should only happen if the `--clean` flag is used. Besides that, resigning this file is a bad idea. We cannot reuse the same method we normally use since we cannot guarantee that the key used to sign that file is the same is the current key we are using and using a different one can break the update for the client.
For this reason, I'm inclined to keep this out of the `reset` feature and don't touch that file. Once a new build is generated, this file will be overwritten, so it is safe to keep it unaltered. Also, mixer should not be using this file internally. If it is, it is a bug.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+	if _, err = ioutil.ReadFile(latestVerSigFilePath); err != nil {
+		skipSign = true
+	}
+	err = b.UpdateLatest(skipSign)
+	if err != nil {
+		fail(err)
+	}
+
+	// if clean flag set
+	if resetCmdFlags.clean {
```
rchiossi: I believe it will look a bit cleaner if you just return here if `!resetCmdFlags.clean`. This removes this huge indentation block.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+
+		for _, f := range files {
+			dirNameInt, err := parseUint32(f.Name())
+			if err != nil {
+				continue
+			}
+			if dirNameInt > stateMixPreviousMixVersionUint32 {
+				err := os.RemoveAll(b.Config.Builder.ServerStateDir + "/image/" + f.Name())
+				if err != nil {
+					continue
```
rchiossi: we should evaluate what was the error here and at least print a warning. I assume it will be common to forget to use sudo to run this command which should result in a permission error. It would be better to inform the user if something goes wrong.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,200 @@
...
+			}
+		}
+
+		// Remove any folder inside update/www/version/ for version above prevStableVersion
+		files, err = ioutil.ReadDir(b.Config.Builder.ServerStateDir + "/www/version")
+		if err != nil {
+			log.Fatal(err)
+		}
+		// iterate over all the formats folder
+		for _, f := range files {
```
rchiossi: this is also tricky for the same reason as stated for `latest_version`. I believe it is better to not change this folder.

ashleshaAtrey: I am just reading files in this folder, not changing any files. These folders will never be used if clean flag is set. Is  this still tricky and break smtg?

mixer/cmd/reset.go
ashleshaAtrey: yup, got it !

bat/tests/reset-command/run.bats
ashleshaAtrey: To validate if upstreamversion file content is updated. Upstreamversion file will only only change during format bump

rchiossi: That partially correct. `upstreamversion` will update on a build, and a since format bump will perform two builds, it ends up updating the value. Since you are not going through an upstream format barrier here, just a `mixer build all` will suffice.
With that being said, I believe it would be much better if `build upstream-format` was smart enough to run a single build if a format bump is not needed. Instead of defaulting to "always 1 format barrier will be crossed", the command should first check if a bump is really needed and, if not, if should behave as `mixer build all`. This will make the process of updating upstream version more streamlined. This is not for this PR though, it should be its own patch.

mixer/cmd/reset.go
rchiossi: I'll comment on this issue on your revised code.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Couldn't parse mixer version")
+	}
+
+	// Make sure update/image/LAST_VER points to PreviousMixVer
+	if err = ioutil.WriteFile(filepath.Join(b.Config.Builder.ServerStateDir, "image", "LAST_VER"), []byte(b.State.Mix.PreviousMixVer), 0644); err != nil {
+		return fmt.Errorf("couldn't update LAST_VER file: %s", err)
+	}
+
+	// update the state.mix file
```
rchiossi: `mixer.state`

mixer/cmd/reset.go
```diff
@@ -0,0 +1,205 @@
...
+				log.Println(err)
+				continue
+			}
+		}
+	}
+
+	// Remove any folder inside update/www for versions above prevMixVer
+	files, err = ioutil.ReadDir(b.Config.Builder.ServerStateDir + "/www")
+	if err != nil {
+		log.Fatal(err)
```
rchiossi: Any reason why you use `log.Fatal` for update/www while you use `log.Println` for update/image ?

mixer/cmd/reset.go
```diff
@@ -0,0 +1,205 @@
...
+				continue
+			}
+			versionInt, err := parseUint32(strings.TrimSpace(string(version)))
+			if err != nil {
+				log.Println(err)
+				continue
+			}
+			// if version > previousMixVersion
+			if versionInt > stateMixPreviousMixVersionUint32 {
+				err := os.RemoveAll(b.Config.Builder.ServerStateDir + "/www/version/" + f.Name())
```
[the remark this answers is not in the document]
ashleshaAtrey: But format number are user defined. if user doesn't assign format number in ascending order, this will break right?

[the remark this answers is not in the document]
rchiossi: Although format numbers are user assigned, they need to have specific values. First, to ensure 100% compatibility, they need to match upstream format. Second, if they are not in ascending order, manifest generation will eventually break. If the user does not have ascending formats, their mix is already broken.
Format needs to be user assigned to allow devops to enforce a given format and to change format during manual bumps.

mixer/cmd/reset.go
ashleshaAtrey: sorry forgot to change that one, should be log.Println

mixer/cmd/reset.go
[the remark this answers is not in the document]
ashleshaAtrey: Got it.

bat/tests/reset-command/run.bats
ashleshaAtrey: got it. Updated the PR.

ashleshaAtrey:
I updated the docs, but I am not sure auto-completion. How do we enable that ?

mixer/cmd/reset.go
```diff
@@ -0,0 +1,268 @@
...
+	} else {
+		// assuming mixer.state file has the correct info
+		b.MixVer = b.State.Mix.PreviousMixVer
+		b.MixVerUint32, err = parseUint32(b.State.Mix.PreviousMixVer)
+		if err != nil {
+			return err
+		}
+	}
+
+	if b.MixVer == "0"{
```
jwakre: mixversions are required to be > 0. Allowing a mix to be reset to 0 will lead to failures later.

```
mixer init
sudo mixer reset
sudo mixer build all
ERROR: Couldn't build update: failed to create update metadata: manifest has version zero, version must be positive
```

mixer/cmd/reset.go
```diff
@@ -0,0 +1,268 @@
...
+	if err = ioutil.WriteFile(filepath.Join(b.Config.Builder.ServerStateDir, "image", "LAST_VER"), []byte(b.MixVer), 0644); err != nil {
+		return fmt.Errorf("couldn't update LAST_VER file: %s", err)
+	}
+
+	// Make sure update/image/format#/latest points to new mixer version
+	newFormat := "format" + strconv.Itoa(int(currentMixFormatInt))
+	if err = ioutil.WriteFile(filepath.Join(b.Config.Builder.ServerStateDir, "/www/version/", newFormat, "latest"), []byte(b.MixVer), 0644); err != nil {
+		return fmt.Errorf("couldn't update latest file: %s", err)
+	}
+	//update the sig file for the latest
```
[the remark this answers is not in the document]
ashleshaAtrey: This actually needs to be decided, whether we should update the sig file or not

mixer/cmd/reset.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// Make sure update/image/latest_version points to new mixer version
+	if err = ioutil.WriteFile(filepath.Join(b.Config.Builder.ServerStateDir, "/www/version/", "latest_version"), []byte(b.MixVer), 0644); err != nil {
+		return fmt.Errorf("couldn't update latest_version file: %s", err)
+	}
+	//update the sig file for the latest_version
```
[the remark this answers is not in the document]
ashleshaAtrey: same for this, need to decide if updating the sig is a good idea or not

mixer/cmd/reset.go
ashleshaAtrey: yes, I will make it invalid to give reset 0 then

mixer/cmd/reset.go
```diff
@@ -0,0 +1,268 @@
...
+var resetCmdFlags resetFlags
+
+func init() {
+	RootCmd.AddCommand(resetCmd)
+
+	resetCmd.Flags().Int32Var(&resetCmdFlags.toVersion, "to", -1, "Reset to a specific mix version, default = PREVIOUS_MIX_VERSION")
+	resetCmd.Flags().BoolVar(&resetCmdFlags.clean, "clean", false, "Deletes all files with versions bigger than the one provided")
+}
+
+func runReset(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
ashleshaAtrey: yup will fix it

mixer/cmd/reset.go
```diff
@@ -0,0 +1,268 @@
...
+			return errors.Wrapf(err, "Couldn't parse upstream version")
+		}
+		vFile := filepath.Join(b.Config.Builder.VersionPath, b.UpstreamVerFile)
+		if err = ioutil.WriteFile(vFile, []byte(b.UpstreamVer), 0644); err != nil {
+			return err
+		}
+	}
+
+	// Change upstreamURL file content to the new current version upstreamurl
+	var lastStableMixUupstreamURL []byte
```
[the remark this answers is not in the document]
ashleshaAtrey: spelling mistake , will fix it

mixer/cmd/reset.go
ashleshaAtrey: what will be good approach to reset the mix to a state which is after mixer init?

jwakre: If the previous version is 0, I would use the current mixversion. If that is invalid (< 1), I would print a warning and default to 10 which is the `mixer init` mixversion default.

mixer/cmd/reset.go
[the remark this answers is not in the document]
jwakre: I don't think `mixer reset` should change these. I also don't think `mixer reset` should support the --clean flag because it is easy to corrupt manifests when rolling back and updating manifests at an old version.

I think that `mixer reset` should roll back the relevant versions (mixversion/last_ver/previous_ver). When running `mixer build update` or `mixer build all`, manifests ahead of the current mix should be erased otherwise they will become corrupted.

@reaganlo Thoughts?

mixer/cmd/reset.go
ashleshaAtrey: @jwakre  can you test the recent commit. I made the changes.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		return err
+	}
+	b.State.Mix.PreviousMixVer = strconv.Itoa(int(mom.Header.Previous))
+	currentMixFormatInt := mom.Header.Format
+
+	// Make sure FORMAT in mixer.state has the same value as update/www/<previousMixVer>/format
+	var format string
+	if mom.Header.Previous != 0 {
+		format, err = b.GetFormatForVersion(b.State.Mix.PreviousMixVer)
```
reaganlo: Handle the returned err. This can cause lint errors.

mixer/cmd/reset.go
```diff
@@ -0,0 +1,271 @@
...
+	} else {
+		format = strconv.Itoa(int(currentMixFormatInt))
+	}
+	b.State.Mix.Format = strings.TrimSpace(format)
+
+	// Change upstreamVersion file content to the new current version upstreamver
+	var lastStableMixUpstreamVersion string
+	lastStableMixUpstreamVersion, err = b.GetLocalUpstreamVersion(b.MixVer)
+	filename := filepath.Join(b.Config.Builder.ServerStateDir, "www", b.MixVer, "upstreamver")
```
reaganlo: This line should be after the error handling of GetLocalUpstreamVersion(). Can cause lint issues.

PR #609, opened 2019-09-17
Fixing bug for command "mixer build upstream-format"

ashleshaAtrey:
Mixer build upstream-format should run with optional flag --new-format. If
--new-format is not set, the default new format should be current
format+1. This default behavior was not getting set for the current mixer version.
Fixed the behavior to reflect the default behavior.

### DIFF
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -167,6 +167,21 @@ var buildUpstreamFormatCmd = &cobra.Command{
 			fail(err)
 		}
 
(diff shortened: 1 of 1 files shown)


ashleshaAtrey:
Got it. Will put in the fix.

rchiossi:
There were a bunch of new lines that were removed by mistake. Check your editor configuration to prevent that in the future. I manually fixed it and merged this commit on master.

PR #615, opened 2019-10-04
Extract Rpms using rpm2archive and tar instead of dnf

### LINKED ISSUE #631: Improve performance of rpm extraction
Currently Mixer uses dnf to download and extract rpms. Extracting rpms using dnf is quite slow. 

Research and implement a faster means of rpm extraction such as rpm2cpio or rpm2archive. Also, extract in parallel.

ashleshaAtrey:
Extracting rpm using rpm2archive and tar is faster than dnf and will
improve performance significantly.
Fixes #631 

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -36,6 +36,9 @@ var bannedPaths = [...]string{
 	"/tmp/",
 }
 
+const rpmDir = "/var/cache/yum/clear/"
+const extractRetries = 5
+
 func isBannedPath(path string) bool {
 	if path == "/" {
 		return true
@@ -139,6 +142,7 @@ func addUpdateBundleSpecialFiles(b *Builder, bundle *bundle) {
 
 // repoPkgMap is a map of repo names to the packages they provide
 type repoPkgMap map[string][]string
+type repoRpmMap map[string][]packageMetadata
 
 func resolveFilesForBundle(bundle *bundle, repoPkgs repoPkgMap, packagerCmd []string) error {
 	bundle.Files = make(map[string]bool)
@@ -294,18 +298,21 @@ func parseNoopInstall(installOut string) []packageMetadata {
 	return pkgs
 }
 
-func repoPkgFromNoopInstall(installOut string) repoPkgMap {
+func repoPkgFromNoopInstall(installOut string) (repoPkgMap, repoRpmMap) {
 	repoPkgs := make(repoPkgMap)
-
+	repoRpmMap := make(repoRpmMap)
 	// TODO: parseNoopInstall may fail, so consider a way to stop the processing
 	// once we find that failure. See how errorCh works in fullfiles.go.
 	pkgs := parseNoopInstall(installOut)
 
 	for _, p := range pkgs {
 		repoPkgs[p.repo] = append(repoPkgs[p.repo], p.name)
+		repoRpmMap[p.repo] = append(repoRpmMap[p.repo], p)
 	}
-	return repoPkgs
+	return repoPkgs, repoRpmMap
 }
+
+var fileSystemRpm string
 
 func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyDir string) *sync.Map {
 	var wg sync.WaitGroup
@@ -327,6 +334,7 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 			for p := range bundle.AllPackages {
 				queryString = append(queryString, p)
 			}
+			bundle.AllRpmPackages = make(map[string]bool)
 			// ignore error from the --assumeno install. It is an error every time because
 			// --assumeno forces the install command to "abort" and return a non-zero exit
 			// status. This exit status is 1, which is the same as any other dnf install
@@ -334,11 +342,19 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 			// fail in the actual install to the full chroot.
 			outBuf, _ := helpers.RunCommandOutputEnv(queryString[0], queryString[1:], []string{"LC_ALL=en_US.UTF-8"})
 
-			rpm := repoPkgFromNoopInstall(outBuf.String())
-			for _, pkgs := range rpm {
+			rpm, fullRpm := repoPkgFromNoopInstall(outBuf.String())
+			for _, pkgs := range fullRpm {
 				// Add packages to bundle's AllPackages
 				for _, pkg := range pkgs {
-					bundle.AllPackages[pkg] = true
+					name := pkg.name + "-" + pkg.version + "." + pkg.arch + ".rpm"
+					bundle.AllPackages[pkg.name] = true
+					bundle.AllRpmPackages[name] = true
+					// to find the fullName of filesystem rpm, as filesystem needs to be extracted first
+					if bundle.Name == "os-core" {
+						if pkg.name == "filesystem" {
+							fileSystemRpm = name
+						}
+					}
 				}
 			}
 
@@ -363,13 +379,36 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 	return &bundleRepoPkgs
 }
 
-func installFilesystem(packagerCmd []string, chrootDir string) error {
-	installArgs := merge(packagerCmd,
-		"--installroot="+chrootDir,
-		"install",
-		"filesystem",
-	)
-	_, err := helpers.RunCommandOutputEnv(installArgs[0], installArgs[1:], []string{"LC_ALL=en_US.UTF-8"})
+func installFilesystem(chrootDir string, localPath string, packagerCmd []string, downloadRetries int) error {
+	var rpmFull string
+	var err error
+
+	rpmFull = filepath.Join(localPath, fileSystemRpm)
+	rpmMap[fileSystemRpm] = true
+	if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
+		if !Offline {
+			rpmPath, err := downloadRpm(packagerCmd, []string{"filesystem"}, chrootDir, downloadRetries)
+			if err != nil {
+				return err
+			}
+			rpmFull = filepath.Join(rpmPath, fileSystemRpm)
+			if _, err = os.Stat(rpmFull); err != nil {
+				fmt.Println("rpm not found: ", rpmFull)
+				return err
+			}
+		} else {
+			fmt.Println("rpm not found: ", rpmFull)
+			return err
+		}
+	}
+
+	for i := 0; i < extractRetries; i++ {
+		err = extractRpm(chrootDir, rpmFull)
+		if err != nil {
+			continue
+		}
+		break
+	}
 	return err
 }
 
@@ -405,10 +444,6 @@ func initRPMDB(chrootDir string) error {
 func buildOsCore(b *Builder, packagerCmd []string, chrootDir, version string) error {
 	err := initRPMDB(chrootDir)
 	if err != nil {
-		return err
-	}
-
-	if err := installFilesystem(packagerCmd, chrootDir); err != nil {
 		return err
 	}
 
@@ -480,32 +515,142 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
 	return nil, downloadErr
 }
 
-func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *bundle, downloadRetries int) error {
+func downloadRpm(packagerCmd []string, rpmList []string, baseDir string, downloadRetries int) (string, error) {
+	// Retry RPM downloads to avoid timeout failures due to slow network
+	_, err := downloadRpms(packagerCmd, rpmList, baseDir, downloadRetries)
+	if err != nil {
+		return "", nil
+	}
+	path := baseDir + rpmDir
+
+	fullRpmPath, err := filepath.Glob(filepath.Join(path, "clear-*"))
+	if err != nil {
+		fmt.Print("cant find path to rpm")
+		return "", err
+	}
+	if len(fullRpmPath) < 0 {
+		return "", fmt.Errorf("cant find rpms packages")
+	}
+	fullRpmPath[0] = fullRpmPath[0] + "/packages"
+
+	return fullRpmPath[0], nil
+}
+
+func extractRpm(baseDir string, rpm string) error {
+	dir, file := filepath.Split(rpm)
+
+	rpm2Cmd := exec.Command("rpm2archive", file)
+	rpm2Cmd.Dir = dir
+	rpm2Cmd.Env = os.Environ()
+
+	rpmTar := file + ".tgz"
+	tarCmd := exec.Command("tar", "-xf", rpmTar, "-C", baseDir)
+	tarCmd.Env = os.Environ()
+	tarCmd.Dir = dir
+
 	var err error
-	baseDir := filepath.Join(buildVersionDir, "full")
-	rpmList := []string{}
 
-	if len(bundle.AllPackages) > 0 {
-		// There were packages directly included for this bundle so
-		// install to full chroot. This check is necessary so we don't
-		// call dnf install with no package listed.
-		for p := range bundle.AllPackages {
-			rpmList = append(rpmList, p)
+	err = rpm2Cmd.Run()
+	if err != nil {
+		return fmt.Errorf("rpm2archive cmd failed with %v", err.Error())
+	}
+
+	err = tarCmd.Run()
+	if err != nil {
+		return fmt.Errorf("tarCmd failed with %v", err)
+	}
+
+	err = os.Remove(dir + rpmTar)
+	if err != nil {
+		fmt.Println("failed to remove file")
+	}
+
+	return nil
+}
+
+func installBundleToFull(packagerCmd []string, baseDir string, bundle *bundle, downloadRetries int, numWorkers int, localPath string) error {
+	var err error
+	var wg sync.WaitGroup
+	rpmCh := make(chan string)
+
+	var errCh error
+	var rpmPath string
+	var missingRpms []string
+
+	if len(bundle.AllRpmPackages) < numWorkers {
+		numWorkers = len(bundle.AllRpmPackages)
+	}
+	errorCh := make(chan error, numWorkers)
+	wg.Add(numWorkers)
+
+	rpmWorker := func() {
+		var e error
+		for rpm := range rpmCh {
+			// running for extractRetries times as rpm2archive and tar can fail if two files are extracted at exact same path
+			for i := 0; i < extractRetries; i++ {
+				e = extractRpm(baseDir, rpm)
+				if e == nil {
+					break
+				}
+			}
+			if e != nil {
+				errorCh <- e
+				break
+			}
 		}
+		wg.Done()
+	}
 
-		// Retry RPM downloads to avoid timeout failures due to slow network
-		_, err = downloadRpms(packagerCmd, rpmList, baseDir, downloadRetries)
-		if err != nil {
-			return err
+	for rpm := range bundle.AllRpmPackages {
+		if rpmMap[rpm] {
+			continue
 		}
-
-		args := merge(packagerCmd, "--installroot="+baseDir, "install")
-		args = append(args, rpmList...)
-		_, err = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
-		if err != nil {
-			return err
+		rpmFull := filepath.Join(localPath, rpm)
+		if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
+			rpmName := strings.TrimSuffix(rpm, ".rpm")
+			missingRpms = append(missingRpms, rpmName)
 		}
 	}
+
+	if len(missingRpms) > 0 {
+		// downloadRpm if not in offline mode, if in offline mode return error
+		if !Offline {
+			rpmPath, err = downloadRpm(packagerCmd, missingRpms, baseDir, downloadRetries)
+			if err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("rpms not found: %s ", missingRpms)
+		}
+	}
+
+	for i := 0; i < numWorkers; i++ {
+		go rpmWorker()
+	}
+
+	// feed the channel
+	for rpm := range bundle.AllRpmPackages {
+		if rpmMap[rpm] {
+			continue
+		}
+		rpmFull := filepath.Join(localPath, rpm)
+		if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
+			rpmFull = filepath.Join(rpmPath, rpm)
+		}
+		rpmMap[rpm] = true
+		select {
+		case rpmCh <- rpmFull:
+		case errCh = <-errorCh:
+			fmt.Println("extraction failed for ", rpm)
+			break
+		}
+		if errCh != nil {
+			return errCh
+		}
+	}
+
+	close(rpmCh)
+	wg.Wait()
 
 	bundleDir := filepath.Join(baseDir, "usr/share/clear/bundles")
 	err = os.MkdirAll(filepath.Join(bundleDir), 0755)
@@ -556,7 +701,9 @@ func rmDNFStatePaths(fullDir string) {
 	}
 }
 
-func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVersionDir, version string, downloadRetries int) error {
+var rpmMap map[string]bool
+
+func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVersionDir, version string, downloadRetries int, numWorkers int) error {
 	fmt.Println("Cleaning DNF cache before full install")
 	if err := clearDNFCache(packagerCmd); err != nil {
 		return err
@@ -564,10 +711,24 @@ func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVers
 	fmt.Println("Installing all bundles to full chroot")
 	totalBundles := len(*set)
 	fullDir := filepath.Join(buildVersionDir, "full")
+	err := os.MkdirAll(fullDir, 0755)
+	if err != nil {
+		return err
+	}
 	i := 0
+	rpmMap = make(map[string]bool)
+
+	if err := installFilesystem(fullDir, b.Config.Mixer.LocalRepoDir, packagerCmd, downloadRetries); err != nil {
+		return err
+	}
+
 	for _, bundle := range *set {
 		i++
 		fmt.Printf("[%d/%d] %s\n", i, totalBundles, bundle.Name)
+
+		if err := installBundleToFull(packagerCmd, fullDir, bundle, downloadRetries, numWorkers, b.Config.Mixer.LocalRepoDir); err != nil {
+			return err
+		}
 		// special handling for os-core
 		if bundle.Name == "os-core" {
 			fmt.Println("... building special os-core content")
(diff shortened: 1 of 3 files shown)


builder/bundles.go
```diff
@@ -480,7 +480,50 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+	} else {
+		return fmt.Errorf("cant find rpms packages")
+	}
+
+	rpmFull := rpm + "*.rpm"
+	out, err := filepath.Glob(filepath.Join(newPath[0], rpmFull))
+	if err != nil {
+		return err
+	}
+	if len(out) > 0 {
```
jwakre: I think it's safe to remove this condition. If len(out) == 0, then the loop won't run.

bryteise:
@jwakre I like the idea of being able to stream rpms to rpm2cpio though I imagine folks may want to have the rpms cached somewhere. At least in the cases where they are rebasing their mix on top of Clear every release. That way since most of the rpms won't have changed between the two most recent releases they can limit their download since the rpms would still be required for creating chroots.

builder/bundles.go
```diff
@@ -480,7 +480,50 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+		for i := range out {
+			cmd := exec.Command("rpm2cpio", out[i])
```
bryteise: Would you consider adding a check that takes the filelist of the rpm and makes sure that all the paths in it exist already in the chroot before the rpm2cpio is run? This is likely a bit paranoid but I think I would like to have at least a warning available to look at in case something like that comes up that I can track down later in the release process.

builder/bundles.go
```diff
@@ -499,9 +542,28 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
...
-		_, err = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
+		// check later if download can be done concurrently
+		var wg sync.WaitGroup
+		fmt.Printf("Resolving packages using %d workers\n", numWorkers)
+		wg.Add(numWorkers)
+		rpmCh := make(chan string)
+
+		rpmWorker := func() {
+			for rpm := range rpmCh {
+				err = extractRpm(baseDir, rpm)
```
reaganlo: Is this err ignored?

builder/bundles.go
ashleshaAtrey: yup sure.

builder/bundles.go
```diff
@@ -499,9 +542,28 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
 			return err
 		}
 
-		args := merge(packagerCmd, "--installroot="+baseDir, "install")
-		args = append(args, rpmList...)
-		_, err = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
+		// check later if download can be done concurrently
```
reaganlo: If this comment is indicating future work, maybe prepend "TODO: " to it so that we don't overlook it. And I think it should be above the line
`_, err := downloadRpms(packagerCmd, []string{"filesystem"}, chrootDir, 1)`

jwakre:
> @jwakre I like the idea of being able to stream rpms to rpm2cpio though I imagine folks may want to have the rpms cached somewhere. At least in the cases where they are rebasing their mix on top of Clear every release. That way since most of the rpms won't have changed between the two most recent releases they can limit their download since the rpms would still be required for creating chroots.

@bryteise Caching the rpms is a good idea. We could even do both: stream the rpms to cpio and write them to disk. It would be better than writing rpms to disk and read them back into memory later.

jwakre:
> I like the idea of being able to stream rpms to rpm2cpio though I imagine folks may want to have the rpms cached somewhere

@bryteise @reaganlo After thinking about this further, I think we could have a faster, more disk efficient solution without directly caching rpms. Instead, we could leverage the bundle info files to determine the list of files for a given rpm that are installed in the n-1 chroot. When an rpm is re-used for incremental builds, we could hardlink the files instead of extracting from a cached rpm or downloading the rpm again. Hardlinking previously existing content would be fast and drastically reduce the disk impact of storing multiple chroots.

builder/bundles.go
```diff
@@ -364,13 +364,39 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
...
+	}
+	rpmFull := "filesystem" + "*.rpm"
+
+	fullPath, err := filepath.Glob(filepath.Join(newPath[0], rpmFull))
+	if err != nil {
+		fmt.Println("cant find rpm path")
+		return err
+	}
+	for _, v := range fullPath {
+		if rpmMap[v] == 1 {
```
bryteise: Is this because of the source rpm? I am struggling to understand what is being skipped. It seems very sensitive to the rpm name being consistently ordered.

builder/bundles.go
```diff
@@ -493,15 +547,79 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
...
+					break
+				}
+			}
+		}
+		close(rpmCh)
+		wg.Wait()
+		// an error could happen after all the workers are spawned so check again for an
+		// error after wg.Wait() completes.
+		if len(errorCh) > 0 {
+			err = <-errorCh
```
rchiossi: Nothing sets err above this line, so you can call return here directly and delete the `if err != nil` below.

builder/bundles.go
```diff
@@ -493,15 +547,79 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
...
-		_, err = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
+		// feed the channel
+		for _, rpm := range rpmList {
+			rpmFull := rpm + "*.rpm"
+			fullPath, err := filepath.Glob(filepath.Join(newPath[0], rpmFull))
+			if err != nil {
+				fmt.Println("cant find rpm path")
+				return err
+			}
+			for _, i := range fullPath {
```
rchiossi: `i` is a very consusing name here since it is not an index. Use `path` instead.

builder/bundles.go
```diff
@@ -556,7 +674,10 @@ func rmDNFStatePaths(fullDir string) {
 	}
 }
 
-func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVersionDir, version string, downloadRetries int) error {
+var rpmMap map[string]int
```
rchiossi: if the value will only be 0 and 1, use boolean instead of int for the map.

builder/bundles.go
```diff
@@ -480,7 +506,35 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
-func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *bundle, downloadRetries int) error {
+func extractRpm(baseDir string, rpm string) error {
+	cmd := exec.Command("rpm2cpio", rpm)
+	cmd.Env = os.Environ()
+	var outBuf bytes.Buffer
+	cmd.Stdout = &outBuf
+
+	err := cmd.Run()
+	if err != nil {
+		fmt.Println("cmd1 failed")
```
bryteise: This isn't a very actionable error. I assume that the err being returned has the command being run and the result in it so maybe just return instead. Otherwise rpm2cpio failed would be more useful.

builder/bundles.go
```diff
@@ -480,7 +506,35 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+		return err
+	}
+
+	cmd2 := exec.Command("sudo", "cpio", "-idm")
+	cmd2.Env = os.Environ()
+	cmd2.Dir = baseDir
+	cmd2.Stdin = &outBuf
+	err = cmd2.Run()
+	if err != nil {
+		fmt.Println("cmd2 failed")
```
bryteise: Same comment about this error message.

builder/bundles.go
```diff
@@ -364,13 +364,39 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
...
-		"install",
-		"filesystem",
-	)
-	_, err := helpers.RunCommandOutputEnv(installArgs[0], installArgs[1:], []string{"LC_ALL=en_US.UTF-8"})
-	return err
+	_, err := downloadRpms(packagerCmd, []string{"filesystem"}, chrootDir, 1)
+	if err != nil {
+		return err
+	}
+	path := chrootDir + "/var/cache/yum/clear/"
```
rchiossi: We should avoid hardcode paths, specially ones that repeat in other places. Put this into a variable or constant.

builder/bundles.go
```diff
@@ -480,7 +506,35 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
 	return nil, downloadErr
 }
 
-func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *bundle, downloadRetries int) error {
+func extractRpm(baseDir string, rpm string) error {
+	cmd := exec.Command("rpm2cpio", rpm)
+	cmd.Env = os.Environ()
```
rchiossi: Do you need to set env here?

builder/bundles.go
```diff
@@ -480,7 +506,35 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+	var outBuf bytes.Buffer
+	cmd.Stdout = &outBuf
```
rchiossi: You should probably use an io.Pipe here instead.

builder/bundles.go
rchiossi: pretty much she is adding the rpm name to a map and checking if it was already extracted to the chroot. This skip extraction if an rpm is part of multiple bundles (although in this particular case it should always be the first time filesystem is extracted so this check will be ignored).

builder/bundles.go
```diff
@@ -493,15 +547,79 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
...
+		fmt.Println("download time : ", time.Since(startTime).String())
+		elapsedTime = elapsedTime + time.Since(startTime)
+
+		path := baseDir + "/var/cache/yum/clear/"
+		newPath, err := filepath.Glob(filepath.Join(path, "clear-*"))
+		if err != nil {
+			fmt.Print("cant find path to rpm")
+			return err
+		}
+		if len(newPath) > 0 {
```
rchiossi: if you swap `if` and `else` conditions here, you wont need an `else` block anymore. It will make the code look cleaner.

builder/bundles.go
```diff
@@ -493,15 +547,79 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
...
+				return err
+			}
+			for _, i := range fullPath {
+				if rpmMap[i] == 1 {
+					continue
+				}
+				rpmMap[i] = 1
+				select {
+				case rpmCh <- i:
+				case <-errorCh:
```
jwakre: I think this throws away the error. Other mixer examples use:

`case err = <-errorCh:`

builder/bundles.go
ashleshaAtrey: yes as it checks if rpm2cpio is installed or not

builder/bundles.go
ashleshaAtrey: yes, I think this will increase performance. I will try it out. Thanks!

builder/bundles.go
```diff
@@ -326,19 +330,20 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 			)
 			for p := range bundle.AllPackages {
 				queryString = append(queryString, p)
+				delete(bundle.AllPackages, p)
```
rchiossi: Why do you need to delete the items here? Also, why delete items one by one instead of the whole list at once?

builder/bundles.go
```diff
@@ -480,30 +471,146 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
-func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *bundle, downloadRetries int) error {
+func downloadRpm(packagerCmd []string, rpmList []string, baseDir string, downloadRetries int) (error, string) {
+	// Retry RPM downloads to avoid timeout failures due to slow network
+	_, err := downloadRpms(packagerCmd, rpmList, baseDir, downloadRetries)
+	if err != nil {
+		return err, ""
+	}
+	path := baseDir + rpmDir
+
+	fullRpmPath, err := filepath.Glob(filepath.Join(path, "clear-*"))
```
rchiossi: Can't we use `upstreamver` here to form the path instead of using `Glob()` ?

builder/bundles.go
```diff
@@ -480,30 +471,146 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
-			return err
+	for i := 0; i < numWorkers; i++ {
+		go rpmWorker()
+	}
+
+	// feed the channel
+	var errCh error
+	var rpmList []string
+	for rpm, present := range bundle.AllPackages {
+		if present {
```
rchiossi:
```
if !present || rpmMap[rpm] {
    continue
}
````

This will move the core of the core out of the indentation block.

File: builder/bundles.go [hunk 19]
```diff
@@ -480,30 +471,146 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+			if rpmMap[rpm] {
```
rchiossi: If rpm is the map key, will there ever be a case where you will hit this condition?

builder/bundles.go
```diff
@@ -480,30 +471,146 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+				continue
+			}
+			rpmFull := filepath.Join(localPath, rpm)
+			if _, err = os.Stat(rpmFull); err != nil {
+				fmt.Println(rpm)
```
rchiossi: Should this print be here?

builder/bundles.go
```diff
@@ -480,30 +471,146 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+
+	if !Offline {
+		if len(rpmList) > 0 {
+			err, rpmPath := downloadRpm(packagerCmd, rpmList, baseDir, downloadRetries)
+			if err != nil {
+				return err
+			}
+			for _, rpm := range rpmList {
+				rpm = rpm + ".rpm"
+				if rpmMap[rpmPath] {
```
rchiossi: Same as above. Will you ever hit this condition?

builder/bundles.go
```diff
@@ -480,30 +471,146 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+					continue
+				}
+				rpmFull := filepath.Join(rpmPath, rpm)
+				if _, err = os.Stat(rpmFull); err != nil {
```
rchiossi: Wont this break the build if the package is missing at this point?

builder/bundles.go
```diff
@@ -524,7 +631,14 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
...
 
-	return writeBundleInfoPretty(bundle, filepath.Join(metaPath, bundle.Name))
+	err = writeBundleInfoPretty(bundle, filepath.Join(metaPath, bundle.Name))
+	if err != nil {
+		return err
+	}
+
+	close(rpmCh)
+	wg.Wait()
+	return nil
```
rchiossi: Move `close(rpmCh)` and `wg.Wait()` above `writeBundleInfoPretty` and keep the original return line.

builder/bundles.go
```diff
@@ -556,18 +670,30 @@ func rmDNFStatePaths(fullDir string) {
...
+
+func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVersionDir, version string, downloadRetries int, numWorkers int) error {
 	fmt.Println("Cleaning DNF cache before full install")
 	if err := clearDNFCache(packagerCmd); err != nil {
 		return err
 	}
 	fmt.Println("Installing all bundles to full chroot")
 	totalBundles := len(*set)
 	fullDir := filepath.Join(buildVersionDir, "full")
+	err := os.MkdirAll(filepath.Join(buildVersionDir, "full"), 0755)
```
rchiossi: use `fullDir` instead

builder/bundles.go
ashleshaAtrey: This was a debug print, will remove it.

builder/bundles.go
ashleshaAtrey: It is only storing the name of rpm , not the full name of the rpm. I am replacing it later with full name of the rpm so initial entries are not needed. yes, it will be to better to delete it as a whole list.

builder/bundles.go
reaganlo: I think its done because dnf downloads rpms to directories like
`update/image/10/full/var/cache/yum/clear/clear-1e95707add22ef34`

builder/bundles.go
ashleshaAtrey: Should I return error if rpm is not found or continue with rest of the bundles?

rchiossi: The build wont complete if one rpm is missing, so you can stop and return error.

builder/bundles.go
```diff
@@ -480,30 +506,147 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+	}
+	err = cpioCmd.Start()
+	if err != nil {
+		fmt.Printf("cpio failed with %v", err)
+		return err
+	}
+
+	err = rpm2Cmd.Wait()
+	if err != nil {
+		fmt.Printf("rpm2cpio failed with %v", err)
```
reaganlo: Testing this PR still generates error at this step.

builder/bundles.go
```diff
@@ -363,14 +371,32 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
...
+			}
+			rpmFull = filepath.Join(rpmPath, rpm)
+			if _, err = os.Stat(rpmFull); err != nil {
+				fmt.Println("rpm not found: ", rpmFull)
+				return err
+			}
+		}
+	}
+
+	return extractRpm(chrootDir, rpmFull, 5)
```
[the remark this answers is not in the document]
rchiossi: I agree. I would be better to retry the whole extract step and let the caller decide on the number of retries.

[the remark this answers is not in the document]
ashleshaAtrey: should I reuse the retires flag, or implement a new flag?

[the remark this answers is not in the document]
reaganlo: I would suggest using the existing retries flag. Its default value is 3. While testing if we feel that value is low, we could use a separate hard-coded value. 
I would like to avoid the user having to choose a value in this case, because this is a workaround we have as part of an implementation. The user might not be aware of why its used unless they understand the internal implementation details.

builder/bundles.go
```diff
@@ -480,30 +506,147 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+		fmt.Printf("cpio failed with %v", err)
+		return err
+	}
+
+	err = rpm2Cmd.Wait()
+	if err != nil {
+		fmt.Printf("rpm2cpio failed with %v", err)
+		return err
+	}
+	err = pw.Close()
```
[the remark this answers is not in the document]
rchiossi: You have to wait until cpio finishes before closing its stdin.

builder/bundles.go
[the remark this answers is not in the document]
rchiossi: Existing retry flags were intended to be used with download failures, so it makes sense to let the user decide since we don't know how good or bad their network is. What is the need for retries in this case in particular? Why the extraction fails? I don't think we should tie this to the retry flag since they don't seem to serve the same purpose.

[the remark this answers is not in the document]
reaganlo: I agree. I was trying to make a point that lets not have the caller decide the retry value in this case.

builder/bundles.go
```diff
@@ -363,13 +372,39 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
...
+			if _, err = os.Stat(rpmFull); err != nil {
+				fmt.Println("rpm not found: ", rpmFull)
+				return err
+			}
+		}
+	}
+
+	for i := 0; i < extractRetries; i++ {
+		err = extractRpm(chrootDir, rpmFull)
+		if err != nil {
```
reaganlo: You could do:
```
		if err == nil {
			break
		}
```
and avoid the extra `continue` statement

ashleshaAtrey:
cpio is not able to retain information regarding group the file belongs to. For all upstream bundles, these four files have different group names.
 /usr/bin/locate
/usr/bin/postdrop
/usr/bin/postqueue
 /usr/libexec/dbus-daemon-launch-helper

for example, after extraction, cpio defaults the group to root while dnf has group locate
 -rwx--s--x 1 root locate 39024 Jul 17 16:48 ./update-dnf2/image/10/full/usr/bin/locate     //dnf
-rwx--s--x 1 root root   39024 Jul 17 16:48 ./update/image/10/full/usr/bin/locate                // cpio

phmccarty:
@reaganlo @ashleshaAtrey You may need to use the `rpm2archive` command instead, which creates a tarball from the rpm's contents. Then you can extract it with `tar`.

For example:

```
$ rpm2archive mlocate-bin-0.26-2.x86_64.rpm
$ sudo tar -xf mlocate-bin-0.26-2.x86_64.rpm.tgz
$ ls -l ./usr/bin
total 84
-rwx--s--x 1 root locate 39024 Jul 17 16:48 locate
-rwxr-xr-x 1 root root   43120 Jul 17 16:48 updatedb
```

builder/bundles.go
```diff
@@ -480,30 +515,149 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+	return err
+}
+
+func installBundleToFull(packagerCmd []string, baseDir string, bundle *bundle, downloadRetries int, numWorkers int, localPath string) error {
+	var err error
+	var wg sync.WaitGroup
+	rpmCh := make(chan string)
+
+	var errCh error
+	var rpmList []string
```
rchiossi: define this variable within the scope that it is used. Also, maybe rename it to `missingRpms` or something else that indicates what it is being used for.

builder/bundles.go
```diff
@@ -480,30 +515,149 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
-			return err
+	for i := 0; i < numWorkers; i++ {
+		go rpmWorker()
+	}
+
+	if !Offline {
+		for rpm := range bundle.AllRpmPackages {
+			rpmFull := filepath.Join(localPath, rpm)
+			if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
+				rpm1 := strings.TrimSuffix(rpm, ".rpm")
```
rchiossi: i Know this is a short lived variable, but `rpm1` is a terrible name. Call it `rpmName` instead.

builder/bundles.go
```diff
@@ -480,30 +515,149 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
-		// Retry RPM downloads to avoid timeout failures due to slow network
-		_, err = downloadRpms(packagerCmd, rpmList, baseDir, downloadRetries)
-		if err != nil {
-			return err
+	for i := 0; i < numWorkers; i++ {
+		go rpmWorker()
+	}
+
+	if !Offline {
+		for rpm := range bundle.AllRpmPackages {
```
rchiossi: this block does not require connectivity and can be outside the `if !Offline` block.

builder/bundles.go
```diff
@@ -480,30 +515,149 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+			rpmFull := filepath.Join(localPath, rpm)
+			if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
+				rpm1 := strings.TrimSuffix(rpm, ".rpm")
+				rpmList = append(rpmList, rpm1)
+			}
+		}
+		if len(rpmList) > 0 {
+			rpmPath, err = downloadRpm(packagerCmd, rpmList, baseDir, downloadRetries)
```
rchiossi: this should be inside an `if !Offline` block and have an else that returns error. This way, if there is any missing rpm, we fail early.

builder/bundles.go
```diff
@@ -480,30 +515,149 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
-		_, err = helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
-		if err != nil {
-			return err
+	// feed the channel
+	for rpm := range bundle.AllRpmPackages {
+		if rpmMap[rpm] {
+			continue
+		}
+		rpmFull := filepath.Join(localPath, rpm)
+		if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
```
rchiossi: This block doesn't do anything. Plus, if you follow my suggestion above, this will never be true.

builder/bundles.go
```diff
@@ -480,30 +515,149 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+		wg.Done()
+	}
 
-		// Retry RPM downloads to avoid timeout failures due to slow network
-		_, err = downloadRpms(packagerCmd, rpmList, baseDir, downloadRetries)
-		if err != nil {
-			return err
+	for i := 0; i < numWorkers; i++ {
+		go rpmWorker()
+	}
```
rchiossi: Start the goroutines after you are done downloading all the packages.

builder/bundles.go
```diff
@@ -524,6 +678,9 @@ func installBundleToFull(packagerCmd []string, buildVersionDir string, bundle *b
 		return err
 	}
 
+	close(rpmCh)
+	wg.Wait()
```
rchiossi: move this below the feed channel loop.

builder/bundleset.go
```diff
@@ -27,6 +27,7 @@ type bundle struct {
 	OptionalIncludes []string
 	DirectPackages   map[string]bool
 	AllPackages      map[string]bool
+	AllRpmPackages   map[string]bool `json:"-"`
```
rchiossi: what is the json tag used for here?

ashleshaAtrey: we use this struct to create allbundles files. I dont want to include this field while writing that file

rchiossi: Then keep it separate at the end of the structure with a note explaining it is not part of the file.
As an example, se how we do it in the config file:
https://github.com/clearlinux/mixer-tools/blob/7a776cbb665986d7215e7c7b14dd32c401701bb6/config/config.go#L33-L47

builder/bundles.go
```diff
@@ -480,33 +515,141 @@ func downloadRpms(packagerCmd, rpmList []string, baseDir string, maxRetries int)
...
+		}
+	}
+
+	for i := 0; i < numWorkers; i++ {
+		go rpmWorker()
+	}
+
+	// feed the channel
+	for rpm := range bundle.AllRpmPackages {
+		if rpmMap[rpm] {
```
rchiossi: Is there ever a situation where this will be true other than for `filesystem`? From what I can see, since `rpm` is the map key, this loop will never loop twice on the same value, so other than for `filesystem` that is processed beforehand, this condition is never reached.

bryteise: Ideally not but it is possible to have subpackages where the content is in a folder of another subpackage. Would be something we'd want scanned prior to release generation I think.

Though actually we have another option, after the extraction we could do permissions checks/fixups based on what the rpm says the files it contains should have. That would let us skip the filesystem must be first hack too.

bryteise: That would make the process slower though so it isn't ideal.

rchiossi: I'm not sure I follow, but I don't think this applies here. We are iterating on rpm package names, their content should not interfere at all on this loop.

rchiossi: Oh nevermind, I realize my confusion now. This is indeed required.

PR #620, opened 2019-10-21
Iterative/Delta: Don't generate Iterative and Delta manifests anymore

otaviobp:
Stop generating delta/iterative manifests is format 29.

I wasn't sure on how to approach tests, so I didn't change anything yet:
 - Should I remove all tests that check for Iterative/Delta manifests?
or
 - Should I run them if format < 29?

And do I need any tests for the transition?


rchiossi:
Since iterative manifest is optional (and were never really used), is there any reason to keep it for older formats? I'd rather just remove it from the code base.

otaviobp:
> Since iterative manifest is optional (and were never really used), is there any reason to keep it for older formats? I'd rather just remove it from the code base.

I'm totally pro that. I'll do it

PR #623, opened 2019-10-22
Remove redundant getUpstreamBundles() call

reaganlo:
InitMix() calls AddBundles() and getUpstreamBundles().
AddBundles() already calls getUpstreamBundles().
So remove the redundant function call in InitMix().

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -200,11 +200,6 @@ func (b *Builder) InitMix(upstreamVer string, mixVer string, allLocal bool, allU
 		return err
 	}
 
-	// Get upstream bundles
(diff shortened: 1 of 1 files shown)


rchiossi:
`getUpstreamBundles()` should be a no-op if called a second time, since it checks if the directory already exists. In init, it marks one of the functionalities that need to take place during that step. The fact that `AddBundle` also calls it is a side effect.

I don't mind merging it though. It is unlikely that we will remove `AddBundle` from init anytime soon, even if we move to a minimalist version to work with 3rd-party bundle builds.

reaganlo:
I agree. But right now its a redundant call. So merge it.

PR #625, opened 2019-10-22
Manifests: Remove Iterative and Delta Manifests

otaviobp:
Reverting all patches that added Iterative and Delta manifests in mixer.
Only Iterative Manifest leftover is the ability to parse MoM with Iterative manifests. This is necessary because when we build an update we need to parse previous manifest, that may have an Iterative Manifest.
Also added code to prevent mixer to carry Iterative manifests to a new MoM when an update is built.


rchiossi:
Overall it looks good. Can you add a comment in the definition of `TypeIManifest` explaining that it is a legacy feature and why it is still there. Might also be good to mark it as a TODO for removal in the future.

otaviobp:
Done. I added a TODO saying we can remove that on format 30.

PR #633, opened 2019-10-30
3rd party: Build a minimum mix

### LINKED ISSUE #632: Make it possible to build a minimum mix to be used by swupd 3rd party
Right now it's not possible to build a mix without:
 - os-core
 - os-core-update
 - os-core-update-index

otaviobp:
In this PR I enable mixer to build a minimum mix including only an empty bundle.

Fixes  #632

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -773,11 +773,6 @@ func (b *Builder) buildBundles(set bundleSet, downloadRetries int) error {
 		return err
 	}
 
-	if _, ok := set[b.Config.Swupd.Bundle]; !ok {
-		return fmt.Errorf("couldn't find bundle %q specified in configuration as the update bundle",
-			b.Config.Swupd.Bundle)
-	}
-
 	// Write INI files. These are used to communicate to the next step of mixing (build update).
 	var serverINI bytes.Buffer
 	_, _ = fmt.Fprintf(&serverINI, `[Server]
@@ -868,7 +863,6 @@ src=%s
 		return err
 	}
 
-	updateBundle := set[b.Config.Swupd.Bundle]
 	var osCore *bundle
 	for _, bundle := range set {
 		if bundle.Name == "os-core" {
@@ -877,7 +871,10 @@ src=%s
 		}
 	}
 	addOsCoreSpecialFiles(osCore)
-	addUpdateBundleSpecialFiles(b, updateBundle)
+
+	if updateBundle, ok := set[b.Config.Swupd.Bundle]; ok {
+		addUpdateBundleSpecialFiles(b, updateBundle)
+	}
 
 	for _, bundle := range set {
 		err = writeBundleInfo(bundle, filepath.Join(buildVersionDir, bundle.Name+"-info"))
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -383,6 +383,10 @@ func installFilesystem(chrootDir string, localPath string, packagerCmd []string,
 	var rpmFull string
 	var err error
 
+	if fileSystemRpm == "" {
+		return nil
+	}
+
 	rpmFull = filepath.Join(localPath, fileSystemRpm)
 	rpmMap[fileSystemRpm] = true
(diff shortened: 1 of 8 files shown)


builder/bundles.go
```diff
@@ -638,7 +638,7 @@ func (b *Builder) buildBundles(set bundleSet, downloadRetries int) error {
 	}
 
 	if _, ok := set[b.Config.Swupd.Bundle]; !ok {
-		return fmt.Errorf("couldn't find bundle %q specified in configuration as the update bundle",
+		fmt.Printf("Warning: couldn't find bundle %q specified in configuration as the update bundle. Bundle won't be configured.\n",
```
rchiossi: Rather than having this warning, we still check if `b.Config.Swupd.Bundle` is `""`. This would indicate that there should be no update bundle. Then, whenever we use mixer for 3rd-party, we change config so that Swupd.BUNDLE is empty.

builder/bundles.go
```diff
@@ -624,7 +624,7 @@ func (b *Builder) buildBundles(set bundleSet, downloadRetries int) error {
 	outputDir := filepath.Join(b.Config.Builder.ServerStateDir, "www")
 
 	if _, ok := set["os-core"]; !ok {
-		return fmt.Errorf("os-core bundle not found")
+		fmt.Printf("Warning: os-core bundle not found. System may not be complete.\n")
```
rchiossi: I don't like having a warning for something that is expected behavior. I'd rather remove it or have it print some sort of "build mode" message, say "Building package mix"/"Building system mix". Those are not proper terms we use in mixer, so we might as well just remove the warning.

swupd/create_manifests.go
```diff
@@ -205,7 +205,7 @@ func processBundles(ui UpdateInfo, c config, numWorkers int) ([]*Manifest, error
 		oldM.sortFilesName()
 		changedFiles, added, deleted := bundle.linkPeersAndChange(oldM, ui.minVersion)
 		// if nothing changed, skip
-		if changedFiles == 0 && added == 0 && deleted == 0 && !changedIncludes {
+		if changedFiles == 0 && added == 0 && deleted == 0 && !changedIncludes && bundle.Name != "full" {
```
rchiossi: Do you know why mixer crashes without full?

builder/bundles.go
otaviobp: That was my first implementation, but I thought that this would be more strait forward. If you remove the os-core-update from the mix, we don't update that. If you add it, we update.
So it's about the bundles that are selected from the user, not the name of the bundle that should include swupd.
But I'm ok to change that too, if you think it's going to be less error prone

builder/bundles.go
otaviobp: ok. I'll remove.
Is there a problem on now enforcing os-core? Should I add a flag or config to make sure no one will create a regular mix without os-core?

swupd/create_manifests.go
otaviobp: It crashes if there's no manifests in a list of manifests in create_manifests. But anyway, we need to create the full manifest for this release even if it hasn't changed, right?

otaviobp:
> I'm not sure about the --skip-update-index option. It looks like a quick hack, It could actually be a viable option to get rid of that bundle in upstream.

First, yes, it's a hack. But my implementation isn't removing the update-index. if you already have that it will just move it forward. I can change it to make it empty and in this case it can be used to get rid of the bundle upstream. But if this is the case, can't we just enforce it to be empty? And then we don't even need a flag?

gtkramer:
If I'm understanding this PR correctly, would this `--skip-update-index` option be passed on `mixer build bundles` and `mixer build update`?  And would it work without supplying 3rd party content?

otaviobp:
V2 - Making some requested changes:
 - Remove error messages on building a mix without os-core or os-core-update
 - Removed the commit that added a os-core-update-index flag because this is going to be handled in #636.
 - Fixed the test to work with os-core-update-index generation

PR #636, opened 2019-10-31
index: Deprecate os-core-update-index

### LINKED ISSUE #541: Hash Mismatch for os-core-update-index Bundle in MoM
Sometimes, after an update where the os-core-update-index bundle has not been changed, the bundle hash entry in the MoM will not match the hash from the previous MoM.

This issue has occurred in the following versions of Clear:

27230->27240
27530->27540

Note: These errors started after the introduction of iterative manifests which did make changes to index manifests.

otaviobp:
Generate an empty bundle os-core-update-index to replace the old index,
    deleting all files included in previous manifests.
    
     - if the previous index exists and it's not empty, we generate an empty index
     - if the previous index doesn't exist we don't generate any index
     - if the previous index is an empty index, we just point to it in the new MoM
    
Fix #541


otaviobp:
As it was described in #541, the root cause of the problems reported by @gtkramer is not the index generation. It is when the index hasn't changed and it's moved to a new Manifest. I also changed the way we are doing that, so the hash won't be computed again and it will be identical to the hash in previous manifest.

jwakre:
> After a running a few builds and a format bump, I noticed the following:
> 
>     1. The os-core-update-index manifest is re-generated by format-bumps and will re-install the files that were deleted.
> 
>     2. The os-core-update-index files in `/usr/share/clear/allbundles/*` are directly injected into the full chroot during `mixer build bundles`. These are no longer inserted into the os-core-update-index manifest, but they are still added to the full manifest which makes it invalid. If we never write these files to the full chroot, their entries in the full manifest will disappear.

Nevermind about the format-bump problem. I think the internal calls to mixer were using my host system's binary.

otaviobp:
1. I was planning to solve format-bump problems by adding a `if (format-bump) remove_the_index_bundle();`

2. I'll stop generating that on a next commit. thanks

otaviobp:
V2:
 - Stop generating allbundles/* files
 - Tested on format bump and minversions releases and it's working as expected

@jwakre Please, let me know if you know any other possible leftovers. I tried to look at things in the code, but it's widely spread, so it's a bit hard.

swupd/create_manifests.go
```diff
@@ -382,29 +364,32 @@ func CreateManifests(version, previous, minVersion uint32, format uint, statedir
...
+		// track here as well so the manifest tar is made
+		// but only if we made it new for this version
+		if osIdx.Header.Version == newMoM.Header.Version {
+			newManifests = append(newManifests, osIdx)
+		}
+	} else {
+		// Carry old index if present
+		for _, f := range oldMoM.Files {
+			if f.Name == IndexBundle {
+				newMoM.Files = append(newMoM.Files, f)
```
reaganlo: Can there be multiple entries for IndexBundle in oldMoM? If not, we could `break` here.

otaviobp: Fixed.

otaviobp: Fixed

rchiossi:
Fixed and merged manually

PR #637, opened 2019-11-04
Selective Compression

rchiossi:
Allow user to configure compression methods for fullfiles.

### DIFF
--- a/swupd/helpers_test.go
+++ b/swupd/helpers_test.go
@@ -858,7 +858,7 @@ func (ts *testSwupd) createFullfiles(version uint32) {
 	}
 	chrootDir := ts.path(filepath.Join("image", fmt.Sprint(version), "full"))
 	outputDir := ts.path(filepath.Join("www", fmt.Sprint(version), "files"))
-	_, err = CreateFullfiles(m, chrootDir, outputDir, 0)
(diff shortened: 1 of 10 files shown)


swupd/helpers_test.go
```diff
@@ -858,7 +858,7 @@ func (ts *testSwupd) createFullfiles(version uint32) {
 	}
 	chrootDir := ts.path(filepath.Join("image", fmt.Sprint(version), "full"))
 	outputDir := ts.path(filepath.Join("www", fmt.Sprint(version), "files"))
-	_, err = CreateFullfiles(m, chrootDir, outputDir, 0)
+	_, err = CreateFullfiles(m, chrootDir, outputDir, 0, []string{"external-bzip2", "external-gzip", "external-xz"})
```
reaganlo: Any reason why we are not passing "external-zstd" also to this test?

rchiossi: Actually yes. The only reason I added the other compression methods here is because there is a test where we create a file and then modify it so that the delta will be bigger than the fullfile itself. This test checks if the right file is chosen. This is the only test where we need the other compression methods for the condition to trigger, but we don't need them all, just one that matches the condition.
This is not testing the compression methods themselves. If that was the case, we would indeed need zstd here as well.

reaganlo: Thank you for the clarification.

PR #640, opened 2019-11-07
Create zero packs in parallel

### LINKED ISSUE #629: Generate zero packs in parallel
On a RAIDed machine with 80 vCPUs, it appears that mixer is generating zero packs in serial.  This corresponds with how many xz processes, one or two, and how mixer is printing to stdout one zero pack at a time.  With the current performance focus on mixer, we could consider adding parallel zero pack generation to the docket.

reaganlo:
Create zero packs in parallel
Fixes #629

Decided not to reuse the parallel code of delta packs for zero packs because there were certain differences such as 1) For zero packs, stop processing in case of error. For delta packs continue processing even in case of error because delta packs are optional. 2) The report and print statements are different. etc.

Decided to reuse the delta-workers as the the numWorker for zero packs. This is because mixer is already overloaded with too many numWorker flags such as delta-workers, bundle-workers and fullfile-workers. Did not find the need to introduce another new one. Also, technically a zero pack is a delta pack.

### DIFF
--- a/builder/deltapacks.go
+++ b/builder/deltapacks.go
@@ -63,19 +63,18 @@ func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printRepor
 	if numWorkers < 1 {
 		numWorkers = runtime.NumCPU()
 	}
-	bundleWorkers := numWorkers
 
 	var bundleQueue = make(chan *swupd.BundleToPack)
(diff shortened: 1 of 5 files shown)


reaganlo:
@rchiossi Fixed the make check error. It was part of the first commit of cleaning up the unused parameters. Will make sure to run both make check and make batcheck.

PR #644, opened 2019-11-12
builder: use clr-installer

mdhorn:
Switch to using clr-installer instead of ister to generate image files.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -17,7 +17,6 @@ package builder
 import (
 	"crypto/rsa"
 	"crypto/x509"
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -26,7 +25,11 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"sort"
+	"strings"
 	"sync"
+
+	"gopkg.in/yaml.v2"
 
 	"github.com/clearlinux/mixer-tools/config"
 	"github.com/clearlinux/mixer-tools/helpers"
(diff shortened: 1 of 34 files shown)


builder/builder.go
```diff
@@ -480,12 +480,14 @@ func (b *Builder) copyImageTemplate(configFile, templateFile string) error {
 // updateImageBundles will update the image bundle list based on the mix bundles.
 func updateImageBundles(configValues *[]byte, bundles []string) error {
 	var data map[string]interface{}
-	err := json.Unmarshal(*configValues, &data)
+
```
reaganlo: If the new line is intentional, its fine. Else we could remove it.

builder/builder.go
```diff
@@ -480,12 +480,14 @@ func (b *Builder) copyImageTemplate(configFile, templateFile string) error {
...
+	err := yaml.Unmarshal(*configValues, &data)
 	if err != nil {
 		return err
 	}
-	data["Bundles"] = bundles
-	*configValues, err = json.MarshalIndent(data, "", " ")
+
```
reaganlo: If the new line is intentional, its fine. Else we could remove it.

builder/builder.go
```diff
@@ -465,10 +461,14 @@ func (b *Builder) copyImageTemplate(configFile, templateFile string) error {
 	if err == nil && len(mixBundles) > 0 {        // check if set is not empty
 		var bundles []string
 		for _, bundle := range mixBundles {
+			// Ignore additional kernel bundles
+			if strings.HasPrefix(bundle.Name, "kernel-") {
```
reaganlo: So the functional difference w.r.t to how ister worked is that,
If the user had built a mix with a different kernel like "kernel-lts", and tried to build an image with default "release-image-config.yaml", the image that gets created will have the default "kernel-native" and the "kernel-lts" would be ignored.

rchiossi:
@mdhorn do you use that yaml library in installer? Just checking if we will need approval to use that.

mdhorn:
> @mdhorn do you use that yaml library in installer? Just checking if we will need approval to use that.

Yes, this is the same version from the clr-installer project.

mdhorn:
Based on discussion this afternoon I'm going to refactor this code after a new releaser of clr-installer to decouple some of the dependencies.

PR #646, opened 2019-11-13
internal: Move internal folder to library

rchiossi:
The internal folder restricts imports to packages in its parent
directory only. Since internal/client is used both by swupd-extract and
swupd-inspector, dep ensure complains that the use is out of scope.

This PR changes internal to library so that the code can be shared.

@cmarcelo Can you take a look? You know this better than I do.


cmarcelo:
What is the error `ensure` is giving? The currently layout seems a correct use of the internal folder: the siblings of the internal can import from it.

rchiossi:
```
##[error]swupd-extract/main.go:25:2: use of internal package github.com/clearlinux/mixer-tools/internal/client not allowed
##[error]swupd-inspector/cat.go:11:2: use of internal package github.com/clearlinux/mixer-tools/internal/client not allowed
```

reaganlo:
Is this a recent error? `dep ensure `on current master seems to be working fine for me.

rchiossi:
I got this with dep v0.5.4 on github actions, but it seems it is a path problem rather than a dep problem. I'll put this PR on hold until I test it some more.

rchiossi:
Not required anymore

PR #647, opened 2019-11-17
Go mod

rchiossi:
Use go mod to manage dependencies and to allow make targets to work out of GOPATH


reaganlo:
Is this ready for review?

rchiossi:
Yes. I'm might tweak some values (like the binary paths), but the core of it shouldn't change.

rchiossi:
Nice catch.

The second one if not a typo though. I will reword it to make it clearer what I meant.

rchiossi:
> What is the significance of vendor/modules.txt file? Is it auto-generated?

Yes. It is a marker that say that this module (mixer) uses vendored dependencies. It is generated by `go mod vendor`

jwakre:
The tablewriter package was added as a dependency in #679 and will need to be included when this PR is rebased.

@reaganlo Was there something blocking this PR?

reaganlo:
@jwakre Yes this has to be added. Nothing is blocking. Was just waiting for the Mixer code to be come stable with all the dnf changes. I think we are good to go now. Will add the new package and merge this one.

reaganlo:
Closing this PR to open another PR #689 with the additional `tablewriter` package.

PR #649, opened 2019-11-20
Switch from ister.py to clr-installer for generating OS images

mdhorn:
Will migrate existing ister JSON file if it exists.
Generates a default YAML template if none exists.
Both ISO generation and Raw Image retention enabled by default.


builder/builder.go
```diff
@@ -404,89 +404,109 @@ func (b *Builder) BuildUpdate(params UpdateParameters) error {
...
-			configFile := filepath.Join(isterConfigDir, template)
-			fmt.Printf("Copying image template from %s\n", configFile)
-			if err = b.copyImageTemplate(configFile, templateFile); err != nil {
+	// If the user did not pass in a configFile, use the default configFile
+	if configFile == "" {
+		configFile = filepath.Join(b.Config.Builder.VersionPath, buildConfig)
+
+		// First check if the previous ister.py configuration file exist,
+		// if so, then we need to migrate it to clr-installer YAML
+		migrationFile := filepath.Join(b.Config.Builder.VersionPath, migrationConfig)
```
rchiossi: I believe it would be better to swap the order here and only checking for the ister config after we check for a clr-installer config. This way, if the user add a clr-installer config in an existing mix before running mixer, the config won't be overwritten by the conversion  of the old ister config.

builder/builder.go
```diff
@@ -404,89 +404,109 @@ func (b *Builder) BuildUpdate(params UpdateParameters) error {
...
 		} else if err != nil {
 			return err
+		} else {
+			fmt.Printf("Warning: Previous generation image config %s found\n", migrationFile)
+			if err = b.migrateConfigFile(migrationFile, configFile); err != nil {
+				return err
+			}
+		}
+	} else {
+		if !strings.HasSuffix(configFile, "yaml") {
```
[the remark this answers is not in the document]
mdhorn: > should we check for `.yml` as well?

clr-installer and other tools are only looking for YAML.

builder/builder.go
mdhorn: Pushed a new version which follows:
```
If YAML not exist:
    if JSON not exist:
        generate YAML
    else:
         migrate JSON to YAML

if JSON exists:
    rename JSON to JSON-EOL
```

mdhorn:
Note: Do not merge this until clr-installer 2.3.4 has been released into the OS

bat/tests/clr-installer-config/run.bats
```diff
@@ -7,28 +7,28 @@ setup() {
...
   # `build image` should fail because the prerequisite commands are not executed.
   # This is intentional in order to reduce test execution time.
-  # The goal is to unit test the creation of ister config file with relevant bundles.
+  # The goal is to unit test the creation of clr-installer config file with relevant bundles.
   run sudo mixer $MIXARGS build image
-  [[ "$status" -eq 1 ]]
-  [[ "$output" =~ "release-image-config.json not found" ]]
+  [[ "$status" -ne 0 ]]
+  [[ "$output" =~ "release-image-config.yaml not found" ]]
   [[ "$output" =~ "Copying image template" ]]
```
reaganlo: `make check` is failing because this `[[ "$output" =~ "Copying image template" ]]` is not required anymore.

mvincerx:
Hey @reaganlo. It's great to see this finally happening.  Please schedule a meeting for 12-02.

PR #651, opened 2019-11-21
Fix index out of bound error

reaganlo:
Fixes index out of bound error.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -519,18 +519,21 @@ func downloadRpm(packagerCmd []string, rpmList []string, baseDir string, downloa
 	// Retry RPM downloads to avoid timeout failures due to slow network
 	_, err := downloadRpms(packagerCmd, rpmList, baseDir, downloadRetries)
 	if err != nil {
-		return "", nil
+		return "", err
 	}
 	path := baseDir + rpmDir
 
-	fullRpmPath, err := filepath.Glob(filepath.Join(path, "clear-*"))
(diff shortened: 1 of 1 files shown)


ashleshaAtrey:
wanted to add this after line 641
	if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
				return fmt.Errorf("rpm not found: %s ", rpmFull)
			}

We are assuming, dnf will download rpm in clear-* repo, but if the package was defined in some other repo other than "clear", it will download in  that repo ,when multiple repos are available. Doing this check we can fail early. 
Not sure if it should be included in this PR though.

rchiossi:
> We are assuming, dnf will download rpm in clear-* repo, but if the package was defined in some other repo other than "clear", it will download in that repo ,when multiple repos are available. Doing this check we can fail early.
> Not sure if it should be included in this PR though.

I believe this would be better on a separate PR since it has a different purpose.

reaganlo:
Ok will add this in a separate PR.

> wanted to add this after line 641
> if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
> return fmt.Errorf("rpm not found: %s ", rpmFull)
> }
> 
> We are assuming, dnf will download rpm in clear-* repo, but if the package was defined in some other repo other than "clear", it will download in that repo ,when multiple repos are available. Doing this check we can fail early.
> Not sure if it should be included in this PR though.

PR #653, opened 2019-11-21
Fix to fail Early if rpm not found

ashleshaAtrey:
Fix to fail Early if rpm not found

if rpm is not found in the cache dir of clear-*, we can fail early
instead of failing during extraction.


reaganlo:
Editing commit message and merging.

PR #657, opened 2019-11-27
Setting multiple repos in mixer

### LINKED ISSUE #654: Setting multiple repos in Mixer
Currently mixer allows users to set multiple repos using 'mixer repo set' commands and name them anyhow. But the code within the mixer is tightly coupled to having only two repos - "clear" and "local". This is a contradiction and can lead to inconsistent results.

ashleshaAtrey:
Currently mixer allows users to set multiple repos using 'mixer repo set' commands and name them anyhow.
But the code within the mixer is tightly coupled to having only two repos - "clear" and "local".
This inconsistency is fixed.
fixes #654

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -36,7 +36,6 @@ var bannedPaths = [...]string{
 	"/tmp/",
 }
 
-const rpmDir = "/var/cache/yum/clear/"
 const extractRetries = 5
 
 func isBannedPath(path string) bool {
@@ -313,27 +312,27 @@ func repoPkgFromNoopInstall(installOut string) (repoPkgMap, repoRpmMap, error) {
 	return repoPkgs, repoRpmMap, nil
 }
 
-func queryRpmName(packageCmd []string, pkgName string) string {
+func queryRpmName(packageCmd []string, pkgName string, repo string) string {
 	queryStringRpm := merge(
 		packageCmd,
 		"repoquery",
 		"--location",
 		"--repo",
-		"local",
+		repo,
 	)
 	queryStringRpm = append(queryStringRpm, pkgName)
 	outBuf, _ := helpers.RunCommandOutputEnv(queryStringRpm[0], queryStringRpm[1:], []string{"LC_ALL=en_US.UTF-8"})
 	if outBuf.String() != "" {
 		rpm := strings.Split(outBuf.String(), "\n")
-		_, rpmName := filepath.Split(rpm[0])
-		return rpmName
+		_, file := filepath.Split(rpm[0])
+		return file
 	}
 	return ""
 }
 
-var fileSystemRpm string
+var fileSystemInfo packageMetadata
 
-func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyDir string, localPath string) (*sync.Map, error) {
+func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyDir string) (*sync.Map, error) {
 	var err error
 	var wg sync.WaitGroup
 	fmt.Printf("Resolving packages using %d workers\n", numWorkers)
@@ -355,14 +354,13 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 			for p := range bundle.AllPackages {
 				queryString = append(queryString, p)
 			}
-			bundle.AllRpmPackages = make(map[string]bool)
+			bundle.AllRpms = make(map[string]packageMetadata)
 			// ignore error from the --assumeno install. It is an error every time because
 			// --assumeno forces the install command to "abort" and return a non-zero exit
 			// status. This exit status is 1, which is the same as any other dnf install
 			// error. Fortunately if this is a different error than we expect, it should
 			// fail in the actual install to the full chroot.
 			outBuf, _ := helpers.RunCommandOutputEnv(queryString[0], queryString[1:], []string{"LC_ALL=en_US.UTF-8"})
-
 			rpm, fullRpm, e := repoPkgFromNoopInstall(outBuf.String())
 			if !isEmptyBundle(bundle) && e != nil {
 				e = errors.Wrapf(e, bundle.Name)
@@ -374,17 +372,13 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 				// Add packages to bundle's AllPackages
 				for _, pkg := range pkgs {
 					rpmName := pkg.name + "-" + pkg.version + "." + pkg.arch + ".rpm"
-					if pkg.repo == "local" {
-						if _, err := os.Stat(filepath.Join(localPath, rpmName)); os.IsNotExist(err) {
-							rpmName = queryRpmName(packagerCmd, pkg.name)
-						}
-					}
 					bundle.AllPackages[pkg.name] = true
-					bundle.AllRpmPackages[rpmName] = true
-					// to find the fullName of filesystem rpm, as filesystem needs to be extracted first
+					bundle.AllRpms[rpmName] = pkg
+
+					// to find the pkg Metadata of filesystem rpm, as filesystem needs to be extracted first
 					if bundle.Name == "os-core" {
 						if pkg.name == "filesystem" {
-							fileSystemRpm = rpmName
+							fileSystemInfo = pkg
 						}
 					}
 				}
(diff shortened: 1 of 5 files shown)


builder/bundles.go
```diff
@@ -352,30 +353,32 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
...
-			bundle.AllRpmPackages = make(map[string]bool)
+			bundle.AllRpmPackages = make(map[string][]string)
 			// ignore error from the --assumeno install. It is an error every time because
 			// --assumeno forces the install command to "abort" and return a non-zero exit
 			// status. This exit status is 1, which is the same as any other dnf install
 			// error. Fortunately if this is a different error than we expect, it should
 			// fail in the actual install to the full chroot.
 			outBuf, _ := helpers.RunCommandOutputEnv(queryString[0], queryString[1:], []string{"LC_ALL=en_US.UTF-8"})
-
 			rpm, fullRpm := repoPkgFromNoopInstall(outBuf.String())
```
reaganlo: Still reviewing. But to get started, can you see if we can use a single dnf query as suggested here https://github.com/clearlinux/mixer-tools/issues/658, instead of having multiple queries (install and repoquery). If yes, we could get rid of the function repoPkgFromNoopInstall().

builder/bundleset.go
```diff
@@ -31,7 +31,7 @@ type bundle struct {
 	Files map[string]bool
 
 	/* hidden property, not to be included in file usr/share/clear/allbundles */
-	AllRpmPackages map[string]bool `json:"-"`
+	AllRpmPackages map[string][]string `json:"-"`
```
reaganlo: Also, instead of `AllRpmPackages map[string][]string`, see if you can define it as `AllRpmPackages map[string]packageMetadata`. Then you can access the values as `pkg.name` and `pkg.repo` instead of `pkg[0]` and `pkg[2]`.

Maybe even consider renaming `AllRpmPackages` to `AllRpms` :)

jwakre: `var fileSystemRpm []string` uses the same string array structure. It would be good to use the `packageMetadata` struct there too.

builder/repo_control.go
```diff
@@ -316,6 +317,12 @@ func (b *Builder) ListRepos() error {
 		if url == "" {
 			continue
 		}
+		// if file uri scheme, the repo can be accessed offline
+		if strings.Contains(url, "file://") {
+			b.repoURL[name] = strings.TrimPrefix(url, "file://")
```
reaganlo: This check won't always give correct results.
Currently,  the `mixer repo` commands do not validate the urls. Users can even set offline repos without the "file://" prefix.

builder/repo_control.go
```diff
@@ -316,6 +317,12 @@ func (b *Builder) ListRepos() error {
 		if url == "" {
 			continue
 		}
+		// if file uri scheme, the repo can be accessed offline
+		if strings.Contains(url, "file://") {
```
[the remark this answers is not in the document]
ashleshaAtrey: Thanks, will look into it

reaganlo:
@ashleshaAtrey Can you please rebase this PR with latest upstream and resolve the conflicts? Also try to avoid using "PR" in the git message.

builder/bundles.go
```diff
@@ -420,24 +425,30 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
...
+	if !repoInfo[fileSystemMetadata.repo].repoOnline {
+		rpmFull = filepath.Join(repoInfo[fileSystemMetadata.repo].repoURL, rpmName)
+	} else {
+		packagerCmdNew := merge(packagerCmd, "--destdir", dnfDownloadDir)
+		_, err := downloadRpms(packagerCmdNew, []string{fileSystemMetadata.name}, chrootDir, downloadRetries)
+		if err != nil {
+			return err
+		}
+		rpmFull = filepath.Join(dnfDownloadDir, rpmName)
+	}
```
reaganlo: So as I mentioned in the above comment, we could determine the rpm download dir outside the queryRpmName().
E.g.
```
	rpm := pkgInfo.name + "-" + pkgInfo.version + "." + pkgInfo.arch + ".rpm"

	if repos[pkgInfo.repo].urlScheme == "file" {
		rpmDir = repos[pkgInfo.repo].url
	} else {
		rpmDir = dnfDownloadDir
		packagerCmdNew := merge(packagerCmd, "--destdir", rpmDir)
		_, err := downloadRpms(packagerCmdNew, []string{pkgInfo.name}, chrootDir, downloadRetries)
		if err != nil {
			return err
		}
	}

	rpmFull = filepath.Join(rpmDir, rpm)

		// If rpm is not found, it could be because rpm filename is not in autospec generated format.
		// In this case, query the dnf database to get the name of the rpm file containing the package.
		rpm = queryRpmName(packagerCmd, pkgInfo.name, pkgInfo.repo)
		if rpm == "" {
			return fmt.Errorf("rpm not found for pkg: %s", pkgInfo.name)
		}
		rpmFull = filepath.Join(rpmDir, rpm)
		if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
			return fmt.Errorf("rpm not found for pkg: %s", pkgInfo.name)
		}
	}

	rpmMap[rpm] = true
```

builder/repo_control.go
```diff
@@ -305,19 +306,36 @@ func (b *Builder) ListRepos() error {
...
+			repoInfo.repoURL = ur.String()
+			repoInfo.repoOnline = false
+			b.repoInfo[name] = repoInfo
+		} else {
+			repoInfo.repoURL = u
+			repoInfo.repoOnline = true
+			b.repoInfo[name] = repoInfo
+		}
+
+		fmt.Printf("%s\t%s\n", name, ur.String())
```
reaganlo: This logic should be in `AddRepo()` instead of in `ListRepos()`. Hence we can have it at just one place.
E.g.
```
	pURL, err := _url.Parse(url)
	if err != nil {
		return err
	}

	if pURL.Scheme == "" { // if url scheme is empty, explicitly set it to "file"
		pURL.Scheme = "file"
	}

	if b.Repos == nil {
		b.Repos = make(map[string]repoInfo)
	}
	var repos repoInfo
	repos.url = pURL.String()
	repos.urlScheme = pURL.Scheme
	b.Repos[name] = repos

	fmt.Printf("Adding %s at %s\n", name, repos.url)
```

mixer/cmd/repos.go
```diff
@@ -113,12 +114,18 @@ func runAddRepo(cmd *cobra.Command, args []string) {
...
-
-	err = b.AddRepo(args[0], args[1])
+	u, err := url.Parse(args[1])
+	if err != nil {
+		fail(err)
+	}
+	if u.Scheme == "" {
+		u.Scheme = "file"
+	}
+	err = b.AddRepo(args[0], u.String())
```
reaganlo: As I mentioned in the previous comment, this logic can be removed from here and can just be in `AddRepo()`

mixer/cmd/repos.go
```diff
@@ -164,9 +171,18 @@ func runSetURLRepo(cmd *cobra.Command, args []string) {
...
+	if u.Scheme == "" {
+		u.Scheme = "file"
+	}
+
+	err = b.SetURLRepo(args[0], u.String())
+	if err != nil {
+		fail(err)
+	}
+
+	fmt.Printf("Set %s baseurl to %s.\n", args[0], u.String())
```
reaganlo: Why do we need this here?

builder/bundles.go
```diff
@@ -671,15 +652,24 @@ func installBundleToFull(packagerCmd []string, baseDir string, bundle *bundle, d
...
 			continue
 		}
-		rpmFull := filepath.Join(localPath, rpm)
+
+		if repoInfo[pkg.repo].repoOnline {
+			rpmFull = filepath.Join(dnfDownloadDir, rpm)
+		} else {
+			rpmFull = filepath.Join(repoInfo[pkg.repo].repoURL, rpm)
+		}
+
```
reaganlo: Similar comments as in the installFilesystem()

builder/bundles.go
```diff
@@ -313,27 +313,37 @@ func repoPkgFromNoopInstall(installOut string) (repoPkgMap, repoRpmMap, error) {
...
+			return ""
+		}
+		if u.Scheme != "" && u.Scheme != "file" {
+			_, file := filepath.Split(rpm[0])
+			return filepath.Join(dnfDownloadDir, file)
+		} else if u.Scheme == "file" {
+			u.Scheme = ""
+			rpm[0] = u.String()
+			return rpm[0]
+		}
```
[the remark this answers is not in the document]
ashleshaAtrey: For the file scheme, location will give the full path. If I don't return full path for other schemes, I will have to check the scheme again in the calling function.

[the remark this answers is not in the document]
reaganlo: But we already know the repo, and we have the repo:scheme mapping in repoInfo. So why do we need to check the scheme from the outBuf again?

builder/repo_control.go
ashleshaAtrey: I didnt understand how we can add this logic to AddRepo, AddRepo is a separate cmd which will exit after adding the repo. We wont iterate over all the repos.

mixer/cmd/repos.go
ashleshaAtrey: This was already in the code, I just changed the return string of baseurl with the scheme. This is runSetURLRepo(), after updating the url of the repo, it is telling user what the url is updated to.

builder/repo_control.go
reaganlo: Actually you are correct. Only this should be here:
```
	pURL, err := _url.Parse(url)
	if err != nil {
		return err
	}

	if pURL.Scheme == "" { // if url scheme is empty, explicitly set it to "file"
		pURL.Scheme = "file"
	}
	fmt.Printf("Adding %s at %s\n", name, pURL.String())

Creation of the repoInfo map can be in ListRepo like you already have. Coming to think of it, it will also handle situations where the user has provided their own DNFConf which I wasn't thinking about.
```

mixer/cmd/repos.go
reaganlo: `SetURLRepo()` calls `AddRepo()`. So if you add the logic in `AddRepo()` including the print statement, you can remove this.

builder/bundles.go
```diff
@@ -420,24 +419,35 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
...
-			}
-			rpmFull = filepath.Join(rpmPath, fileSystemRpm)
-			if _, err = os.Stat(rpmFull); err != nil {
-				return fmt.Errorf("rpm not found: %s", rpmFull)
-			}
-		} else {
-			return fmt.Errorf("rpm not found: %s", rpmFull)
+		// if rpm is not found, name of rpm file is not in autospec generated format. In this case
+		// we query the database to get the name of the rpm file containing the package.
+		rpm = queryRpmName(packagerCmd, fileSystemInfo.name, fileSystemInfo.repo)
```
jwakre: Similarly to my other comment, we should use NVR instead of fileSystemInfo.name to be more specific.

builder/bundles.go
```diff
@@ -671,17 +650,33 @@ func installBundleToFull(packagerCmd []string, baseDir string, bundle *bundle, d
...
+			rpmFull = filepath.Join(dnfDownloadDir, rpm)
+		} else {
+			rpmFull = filepath.Join(repos[pkgInfo.repo].url, rpm)
+		}
+
 		if _, err = os.Stat(rpmFull); os.IsNotExist(err) {
-			rpmFull = filepath.Join(rpmPath, rpm)
+			// if rpm is not found, name of rpm file is not in autospec generated format. In this case
+			// we query the database to get the name of the rpm file containing the package.
+			rpm = queryRpmName(packagerCmd, pkgInfo.name, pkgInfo.repo)
```
jwakre: When the repo has multiple files with the same name, `dnf repoquery` will resolve the paths to multiple rpms. We can be more specific by using the NVRA (`pkg.name+"-"+pkg.version+"."+pkg.arch`) to get a single result.

mixer/cmd/repos.go
reaganlo: As you mentioned, since AddRepo() is called only if repo does not exist, this review comment can be ignored.

mixer/cmd/repos.go
reaganlo: As you mentioned, since AddRepo() is called only if repo does not exist, this review comment can be ignored.

PR #662, opened 2019-12-03
MCA: Skip hash comparisons between md5 and sha256

jwakre:
rpm 4.12 used md5 file hashes by default and rpm 4.14 switched the
default to sha256. As a result, in version 31680 of Clear, rpms started
to use sha256 hashes instead of md5. Since MCA assumes that hash changes
indicate content changes, hash comparisons between md5 and sha256 are
skipped to avoid false positive errors.

### DIFF
--- a/builder/build_validate.go
+++ b/builder/build_validate.go
@@ -95,7 +95,8 @@ type fileInfo struct {
 	user  string
 	group string
 
-	pkg string
+	hashLen int
+	pkg     string
 }
 
 // CheckManifestCorrectness validates that the changes in manifest files between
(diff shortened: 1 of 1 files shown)


builder/build_validate.go
```diff
@@ -592,6 +597,35 @@ func (bundleDiff *mcaBundleDiff) diffBundles(bundle string, fromInfo, toInfo map
...
+			continue
+		}
+
+		// Flag comparisons between md5 hashes (len == 32) and sha256 hashes (len == 64),
+		// so they can be skipped later.
+		fromFile := fromFiles[fName]
+		if fromFile != nil {
+			if (fromFile.hashLen == 32 || fromFile.hashLen == 64) &&
+				(toFile.hashLen == 32 || toFile.hashLen == 64) &&
+				fromFile.hashLen != toFile.hashLen {
```
reaganlo: Any reason why we have these 3 checks:
```
			if (fromFile.hashLen == 32 || fromFile.hashLen == 64) &&
				(toFile.hashLen == 32 || toFile.hashLen == 64) &&
				fromFile.hashLen != toFile.hashLen {
```

instead of just
`			if fromFile.hashLen != toFile.hashLen {`

jwakre: Symlinks have a hash length of 0, so the following hash length combinations are possible and should not be skipped: 0->32, 0->64, 32->0, 64->0

I can readjust the logic to better handle symlinks while also dealing with future hashing algorithm changes.

reaganlo: I think this is good for now. Maybe just add the comment about symlinks the code for readability.

jwakre:
@reaganlo I updated the comment regarding symlinks.

PR #664, opened 2019-12-06
Stop processing bundles if missing dependencies

### LINKED ISSUE #663: Mixer skips packages if dependencies are not found
**Describe the bug:**
If dependencies are not found,
**_Mixer v5.9.1_** catches it and errors out.
**_Mixer v6.1.1_** does not error out and just skips the rpms without notifying the user.

**To Reproduce:**
Download and run the attached `vim-test.txt` file from a clean mixer workspace. This script tries to build a vim bundle using local rpms that depend on libpython3.8.so.1.0 that is not available in upstream clear version 31800.

**_Mixer v5.9.1_** errors out as follows:
```
[3/19] vim
RPM download attempt 1 failed. Maximum of 4 attempts.
RPM download attempt 2 failed. Maximum of 4 attempts.
RPM download attempt 3 failed. Maximum of 4 attempts.
RPM download attempt 4 failed. Maximum of 4 attempts.

TIMINGS
  BUILD BUNDLES 0s
TOTAL: 0s
2019/12/05 16:12:18 ERROR: Error building bundles: 2019/12/05 16:12:18 failed to execute dnf --config=/home/reaganlo/mixer/.yum-mix.conf -y --releasever=31800 --installroot=/home/reaganlo/mixer/update/image/10/full install --downloadonly python3-dev python3-core perl-XML-NamespaceSupport perl-Test-Simple perl-IO-HTML setuptools mlocate swupd-search perl-Config-General findutils gcc-libs-math perl-LWP-Protocol-https perl-Try-Tiny perl-URI perl-libwww-perl vim-gnupg pycodestyle perl-HTTP-Negotiate perl-HTTP-Message perl-HTML-Parser openssl-lib vim perl-DateTime-TimeZone perl-HTML-Tagset perl-File-Listing libstdc++ perl-HTTP-Cookies perl-Error p11-kit perl virtualenv perl-Encode-Locale perl-XML-Parser perl-Config-Tiny perl-man vim-fugitive pip glibc-lib-avx2 perl-Params-Validate perl-DBI perl-LWP-MediaTypes perl-Crypt-SSLeay perl-HTTP-Date ctags vim-go cffi
2019/12/05 16:12:18
STDOUT:
Last metadata expiration check: 0:00:15 ago on Thu 05 Dec 2019 04:12:03 PM PST.
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
2019/12/05 16:12:18
STDERR:
Error:
 Problem: package vim-8.1.2384-1023.x86_64 requires vim-bin = 8.1.2384-1023, but none of the providers can be installed
  - cannot install the best candidate for the job
  - nothing provides libpython3.8.so.1.0()(64bit) needed by vim-bin-8.1.2384-1023.x86_64
2019/12/05 16:12:18 failed to execute
: exit status 1
```
**_Mixer v6.1.1_** does not error out and the vim binary is not created:
`ls update/image/10/full/usr/bin/ | grep vim`

[vim-test.txt](https://github.com/clearlinux/mixer-tools/files/3929852/vim-test.txt)

ashleshaAtrey:
For current version of mixer, if dependencies are not resolved, it
continues with installing other bundles.The correct behavior should be to fail at this point .
While resolving packages, if dependencies are not resolved then we stop the processing 
and return error.

fixes #663

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -259,14 +259,15 @@ func resolveFiles(numWorkers int, set bundleSet, bundleRepoPkgs *sync.Map, packa
 // Transaction Summary
 // =============================================================================
 // <more metadata>
-func parseNoopInstall(installOut string) []packageMetadata {
+func parseNoopInstall(installOut string) ([]packageMetadata, error) {
 	// Split out the section between the install list and the transaction Summary.
 	parts := strings.Split(installOut, "Installing:\n")
 	if len(parts) < 2 {
 		// If there is no such section, e.g. dnf fails because there's
 		// no matching package (so no "Installing:"), return nil. Real
 		// failure will happen later when doing the actual install.
-		return nil
+
+		return nil, fmt.Errorf("unable to resolve package")
 	}
 	pkgList := strings.Split(parts[1], "\nTransaction Summary")[0]
 
@@ -295,21 +296,21 @@ func parseNoopInstall(installOut string) []packageMetadata {
 		pkgs = append(pkgs, pkg)
 	}
 
-	return pkgs
+	return pkgs, nil
 }
 
-func repoPkgFromNoopInstall(installOut string) (repoPkgMap, repoRpmMap) {
+func repoPkgFromNoopInstall(installOut string) (repoPkgMap, repoRpmMap, error) {
 	repoPkgs := make(repoPkgMap)
 	repoRpmMap := make(repoRpmMap)
-	// TODO: parseNoopInstall may fail, so consider a way to stop the processing
-	// once we find that failure. See how errorCh works in fullfiles.go.
-	pkgs := parseNoopInstall(installOut)
-
+	pkgs, err := parseNoopInstall(installOut)
+	if err != nil {
+		return nil, nil, err
+	}
 	for _, p := range pkgs {
 		repoPkgs[p.repo] = append(repoPkgs[p.repo], p.name)
 		repoRpmMap[p.repo] = append(repoRpmMap[p.repo], p)
 	}
-	return repoPkgs, repoRpmMap
+	return repoPkgs, repoRpmMap, nil
 }
 
 func queryRpmName(packageCmd []string, pkgName string) string {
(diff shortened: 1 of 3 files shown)


builder/bundles.go
```diff
@@ -360,7 +365,12 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 			// fail in the actual install to the full chroot.
 			outBuf, _ := helpers.RunCommandOutputEnv(queryString[0], queryString[1:], []string{"LC_ALL=en_US.UTF-8"})
 
-			rpm, fullRpm := repoPkgFromNoopInstall(outBuf.String())
+			rpm, fullRpm, e := repoPkgFromNoopInstall(outBuf.String())
+			if e != nil {
+				errStr := fmt.Sprint("package dependencies not resolved for package ", bundle.Name)
```
jwakre: Was it intentional to create a new error instead of using the one from `repoPkgFromNoopInstall`? Also, I think the wording here is confusing:

"package dependencies not resolved **for package ", bundle.Name**

It sounds like the bundle is called a package.

builder/bundles.go
```diff
@@ -295,21 +296,23 @@ func parseNoopInstall(installOut string) []packageMetadata {
...
 
-	return pkgs
+	return pkgs, nil
 }
 
-func repoPkgFromNoopInstall(installOut string) (repoPkgMap, repoRpmMap) {
+func repoPkgFromNoopInstall(installOut string) (repoPkgMap, repoRpmMap, error) {
 	repoPkgs := make(repoPkgMap)
 	repoRpmMap := make(repoRpmMap)
 	// TODO: parseNoopInstall may fail, so consider a way to stop the processing
```
jwakre: This PR addresses the TODO, so we can remove it.

builder/bundles.go
reaganlo: Instead of creating a new error variable, just wrap.
`e = errors.Wrapf(e,`

builder/bundles.go
```diff
@@ -360,7 +363,11 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 			// fail in the actual install to the full chroot.
 			outBuf, _ := helpers.RunCommandOutputEnv(queryString[0], queryString[1:], []string{"LC_ALL=en_US.UTF-8"})
 
-			rpm, fullRpm := repoPkgFromNoopInstall(outBuf.String())
+			rpm, fullRpm, e := repoPkgFromNoopInstall(outBuf.String())
+			if e != nil {
+				errorCh <- errors.Wrapf(e, bundle.Name)
```
reaganlo: I would suggest also printing the error here for first hand info. Lets not wait for it to be trickled through the go channels. E.g. https://github.com/clearlinux/mixer-tools/blob/master/builder/update.go#L207

reaganlo:
@ashleshaAtrey The bat tests that create dummy bundles "no-pkgs", are failing.

reaganlo:
@ashleshaAtrey Sometimes users might remove make the bundle empty while deprecating it. I think lets continue to allow empty bundles like v5.9.1
We could add a function like this in bundleset.go

```
func isEmptyBundle(bundle *bundle) bool {
	if len(bundle.DirectIncludes) == 0 &&
		len(bundle.OptionalIncludes) == 0 &&
		len(bundle.DirectPackages) == 0 &&
		len(bundle.AllPackages) == 0 &&
		len(bundle.Files) == 0 {
		return true
	}
	return false
}
```


and have that check along with the error check:
```
			if !isEmptyBundle(bundle) && e != nil {
				e = errors.Wrapf(e, bundle.Name)
				fmt.Println(e)
				errorCh <- e
				break
			}
```

and see if all the tests pass.

reaganlo:
Updated commit message and merged.

PR #667, opened 2019-12-13
MCA refactor pkg queries

jwakre:

* MCA: Instead of querying packages downloaded by `dnf install`, the packages are resolved and queried with `rpm`. When the packages are in repos on the local system, the query operation is much faster.
* MCA: Added support to resolve packages that do not follow NVRA name format.
* MCA: Suppress verbose Mixer output
* Return more package metadata during `resolvePackages`

Based on local MCA testing, with all bundles and package repositories on the local system the execution time improved from 106min 11sec -> 15min 11sec.

### DIFF
--- a/builder/build_validate.go
+++ b/builder/build_validate.go
@@ -115,8 +115,13 @@ func (b *Builder) CheckManifestCorrectness(fromVer, toVer, downloadRetries int, 
 	if fromVer >= toVer {
 		return fmt.Errorf("From version must be less than to version")
 	}
-
-	fmt.Printf("WARNING: Local RPMs will override upstream RPMs for both the from and to versions.\n")
+	
+	// Suppress Stdout so that it doesn't clutter the results
+	stdOut := os.Stdout
+	os.Stdout,_ = os.Open(os.DevNull)
+	defer func() {
+		os.Stdout = stdOut
+	}()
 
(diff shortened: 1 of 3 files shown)


builder/build_validate.go
```diff
@@ -273,100 +293,178 @@ func (b *Builder) mcaPkgInfo(manifests []*swupd.Manifest, version, downloadRetri
...
-	pkgFileName := pkg.name + "-" + pkg.version + "." + pkg.arch + ".rpm"
-	pkgPath := filepath.Join(pkgDir, pkgFileName)
-
 	// Query RPM for file metadata lists
-	args := merge(rpmCmd, pkgPath)
-	out, err := helpers.RunCommandOutputEnv(args[0], args[1:], []string{"LC_ALL=en_US.UTF-8"})
-	if err != nil {
-		return nil, err
+	args := merge(rpmCmd, pkg.uri)
+	for attempts := 0; attempts <= downloadRetries; attempts++ {
```
reaganlo: won't this run 1 additional time?

jwakre: I thought of it as the number of times to try again. So, 3 retries would be 1 attempt + 3 retries.

reaganlo: aaah.. makes sense.

PR #676, opened 2019-12-18
repo: Handle rpms in repo sub directories

### LINKED ISSUE #674: Handle subdirs within repo baseurl
Currently, for repos with "file" baseurl scheme, mixer expects the rpms to be directly within the baseurl path. But it is possible that rpms can be in sub directories within the repo baseurl.

Mixer should handle subdirs within repo baseurl.

reaganlo:
Currently, for repos with "file" baseurl scheme, mixer expects the
rpms to be directly within the baseurl path.
But it is possible that rpms can be in sub directories within the repo baseurl.

In order to handle this, if an rpm is not found directly within the repo baseurl,
run repoquery to determine the actual location of the rpm.
This is expensive and it is recommended that rpms are stored directly
within the repo baseurl.

Includes a bat test.
Fixes #674

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -312,7 +313,12 @@ func repoPkgFromNoopInstall(installOut string) (repoPkgMap, repoRpmMap, error) {
 	return repoPkgs, repoRpmMap, nil
 }
 
-func queryRpmName(packageCmd []string, pkgName string, repo string) string {
+// queryRpmFullPath returns the expected rpm full path for a package using dnf repoquery.
+// It queries the dnf database for the location.
+// If the query result is empty, it returns an err.
+// If the query result has a "file" scheme, it returns the full path without the scheme.
+// If the query result has a non "file" scheme, it assumes the full path is within the corresponding repo cache dir.
+func queryRpmFullPath(packageCmd []string, pkgName string, repo string) (string, error) {
 	queryStringRpm := merge(
 		packageCmd,
 		"repoquery",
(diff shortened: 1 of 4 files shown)


builder/bundles.go
```diff
@@ -323,11 +326,14 @@ func queryRpmName(packageCmd []string, pkgName string, repo string) string {
...
-		_, file := filepath.Split(rpm[0])
-		return file
+		out := strings.Split(outBuf.String(), "\n")
+		pURL, err := url.Parse(out[0])
+		if err == nil { // obtain the path without the url scheme
+			pURL.Scheme = ""
+			rpmFullPath = pURL.String()
+		}
 	}
-	return ""
```
ashleshaAtrey: we should create an error string if outBuf.String() is empty. os.stat() won't catch the error as in this case as rpmFullPath will be empty string

builder/bundles.go
```diff
@@ -323,11 +326,14 @@ func queryRpmName(packageCmd []string, pkgName string, repo string) string {
 	queryStringRpm = append(queryStringRpm, pkgName)
 	outBuf, _ := helpers.RunCommandOutputEnv(queryStringRpm[0], queryStringRpm[1:], []string{"LC_ALL=en_US.UTF-8"})
 	if outBuf.String() != "" {
-		rpm := strings.Split(outBuf.String(), "\n")
-		_, file := filepath.Split(rpm[0])
-		return file
+		out := strings.Split(outBuf.String(), "\n")
+		pURL, err := url.Parse(out[0])
```
ashleshaAtrey: if url is not an file scheme, location will give url of the upstream repo but not the path to the rpms stored locally.
for example location will give https://cdn.download.clearlinux.org/releases/31930/clear/x86_64/os/Packages/filesystem-3.0.14-200.x86_64.rpm
we instead need the downloadDir as the full path

builder/bundles.go
reaganlo: os.stat() of empty string gives file not found err.

builder/bundles.go
reaganlo: good catch.. I was making the assumption that queryRpmFullPath() will be called only for file based scheme. Will fix.

builder/bundles.go
jwakre: Is there a reason we don't return an error here instead of during os.stat() when the rpm cannot be resolved?

reaganlo: It is mainly for fewer lines of code and the assumption that there would not be a performance impact as the chances of os.Stat("") during this stage is very less.

But I think since you both have brought up this concern, I will return an err if the rpm is not resolved.

builder/bundles.go
```diff
@@ -322,12 +328,26 @@ func queryRpmName(packageCmd []string, pkgName string, repo string) string {
...
+		return "", err
+	}
+
+	var rpmFullPath string
+	if pURL.Scheme == "file" { // obtain the full path without the url scheme
+		pURL.Scheme = ""
+		rpmFullPath = pURL.String()
+	} else { // obtain only the rpm name and append to repo cache dir
+		_, rpm := filepath.Split(out[0])
+		rpmFullPath = filepath.Join(repos[repo].cacheDir, rpm)
```
ashleshaAtrey: Instead of passing repos to the function, we can use dnfDownloadDir directly as it is global. This works too.

reaganlo: @ashleshaAtrey @jwakre I might even consider making `repos map[string]repoInfo` a global variable. But maybe in a different PR after @jwakre 's MCA PR is merged.

PR #680, opened 2019-12-20
repo: Improve performance for multiple repo subdir

reaganlo:
repo: Improve performance for multiple repo subdir

If an rpm is not found in the assumed cache dir of its repo, update the
repo cache dir list based on the location of the rpm.

This is an attempt to improve performance in case the rpms are not
stored directly within the repo baseurl.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -317,7 +317,7 @@ func repoPkgFromNoopInstall(installOut string) (repoPkgMap, error) {
 // If the query result is empty, it returns an err.
 // If the query result has a "file" scheme, it returns the full path without the scheme.
 // If the query result has a non "file" scheme, it assumes the full path is within the corresponding repo cache dir.
-func queryRpmFullPath(packageCmd []string, pkgName string, repo string) (string, error) {
+func queryRpmFullPath(packageCmd []string, pkgName string, repo string, repos map[string]repoInfo) (string, error) {
 	queryStringRpm := merge(
 		packageCmd,
 		"repoquery",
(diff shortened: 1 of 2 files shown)


builder/bundles.go
```diff
@@ -342,8 +342,16 @@ func queryRpmFullPath(packageCmd []string, pkgName string, repo string) (string,
 	if pURL.Scheme == "file" { // obtain the full path without the url scheme
 		pURL.Scheme = ""
 		rpmFullPath = pURL.String()
+		// If repo cache dir has not been guessed before, guess it just one time.
+		// This is an attempt to improve performance in case the rpms are not stored directly within the repo baseurl.
+		if !repos[repo].guessCache {
+			repoInfo := repos[repo]
+			repoInfo.guessCache = true
+			repoInfo.cacheDir = filepath.Dir(rpmFullPath)
+			repos[repo] = repoInfo
```
jwakre: Why is `repoInfo` made? Couldn't we directly update repos[repo] with less code?

```
repos[repo].guessCache = true
repos[repo].cacheDir = filepath.Dir(rpmFullPath)
```

builder/bundles.go
```diff
@@ -342,8 +342,16 @@ func queryRpmFullPath(packageCmd []string, pkgName string, repo string) (string,
 	if pURL.Scheme == "file" { // obtain the full path without the url scheme
 		pURL.Scheme = ""
 		rpmFullPath = pURL.String()
+		// If repo cache dir has not been guessed before, guess it just one time.
+		// This is an attempt to improve performance in case the rpms are not stored directly within the repo baseurl.
+		if !repos[repo].guessCache {
```
jwakre: I don't think `guessCache` is necessary right now. We probably will only call `queryRpmFullPath` when guessCache was a miss, so we could update it.

Alternatively, we could maintain a list of all the sub-directories with packages for a given repo. The `stat` commands could try all known sub-directories.

builder/bundles.go
reaganlo: go doesn't let you do that assignment. Its a compile error. I could maybe pass the repo map as a pointer.

builder/bundles.go
reaganlo: The reason guessCache is used is so that the guess is just done once in case there are multiple sub-dirs within the baseurl.

But if we make the change to make cacheDir as a list of directories, guessCache will not be required.

reaganlo: @jwakre Updated PR based on your suggestion
`Alternatively, we could maintain a list of all the sub-directories with packages for a given repo. The stat commands could try all known sub-directories.`

builder/repo_control.go
```diff
@@ -36,7 +36,7 @@ type dnfRepoConf struct {
 }
 
 type repoInfo struct {
-	cacheDir  string
+	cacheDirs map[string]string
```
jwakre: We are throwing away the value, so cacheDirs could be `map[string]bool` (typically done in Mixer) or `map[string]struct{}`

PR #685, opened 2019-12-30
swupd: Retain previously deleted file entries

### LINKED ISSUE #684: Deleted file entries sometimes dropped without format bump
**Steps to reproduce with Mixer v6.1.4**
1. mixer init --clear-version 31090
2. mixer bundle add evolution
3. mixer build all --skip-packs --skip-fullfiles
4. mixer versions update --clear-version 31110
5. mixer build all --skip-packs --skip-fullfiles
6. mixer versions update --clear-verison 31300
7. mixer build all --skip-packs --skip-fullfiles
8. The file, `usr/share/vala/vapi/gtkspell3-3.0.deps`, is deleted in the evolution bundle in mix 20, but the deletion entry was dropped in the evolution bundle in mix 30. The expected behavior is that deletion entries will be forwarded until a format bump.

**Expected Behavior:**
The file, `usr/share/vala/vapi/gtkspell3-3.0.deps` should exist in the evolution bundle in mix 30 as a deleted file entry.

**Cause**
When linking the old/new versions of a manifest, deleted manifests may be dropped when the length of file entries is longer in the old manifest.

jwakre:
Fixes #684 

* Prevent newly generated manifests from dropping previously deleted file entries when the new manifest has less file entries than the previous manifest.
* Add test case to verify that previously deleted file entries are retained.

### DIFF
--- a/swupd/manifest.go
+++ b/swupd/manifest.go
@@ -439,9 +439,12 @@ func (m *Manifest) linkPeersAndChange(oldManifest *Manifest, minVersion uint32) 
 		added = append(added, nf)
 	}
 
-	// anything remaining in oldManifest is newly deleted in the new manifest
+	// anything remaining in oldManifest is deleted in the new manifest
 	for _, of := range oldManifest.Files[ox:omFilesLen] {
-		if of.Status == StatusDeleted || of.Status == StatusGhosted {
+		if !of.Present() {
+			if of.Status == StatusDeleted && m.Header.Format == oldManifest.Header.Format {
+				m.Files = append(m.Files, of)
+			}
 			continue
 		}
 		m.newDeleted(of)
--- a/swupd/manifest_test.go
+++ b/swupd/manifest_test.go
@@ -395,6 +395,7 @@ func TestLinkPeersAndChange(t *testing.T) {
 			{Name: "3", Status: StatusGhosted, Info: sizer(0)},
 			{Name: "4", Status: StatusUnset, Info: sizer(0)},
 			{Name: "5", Status: StatusUnset, Hash: 1, Info: sizer(0)},
+			{Name: "7", Status: StatusDeleted, Info: sizer(0)},
 		},
 	}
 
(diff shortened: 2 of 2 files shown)


swupd/manifest.go
```diff
@@ -439,9 +439,12 @@ func (m *Manifest) linkPeersAndChange(oldManifest *Manifest, minVersion uint32)
 		added = append(added, nf)
 	}
 
-	// anything remaining in oldManifest is newly deleted in the new manifest
+	// anything remaining in oldManifest is deleted in the new manifest
 	for _, of := range oldManifest.Files[ox:omFilesLen] {
-		if of.Status == StatusDeleted || of.Status == StatusGhosted {
```
jwakre: @otaviobp @phmccarty Are either of you familiar with the expected behavior of ghosted files?

Currently, the **behavior of ghosted files is inconsistent**. Based on the name value of the ghosted file (different logic is used when the name is at the beginning or end of the sorted file list), the following scenarios can happen:
1. The ghosted file is dropped from the new manifest
2. The ghosted file is considered a newly deleted file in the new manifest

This change makes ghosted files follow `scenario 1` which is the most common case.

I am not familiar with ghosted files, but I don't feel like dropping them or marking them as newly deleted files is correct. Does it make sense for ghosted files to persist in future versions of the manifest within the same format?

jwakre:
@reaganlo The bat tests are passing for me locally and on Travis. What error do you see?

swupd/manifest.go
phmccarty: IIRC, ghosted files should only appear in exactly one version of a manifest. They are supposed to disappear from the manifests in the next build that the manifest changes.

phmccarty: For more background: ghosted files were implemented for the kernel-* bundles so that mixer could create (rename) deltas for kernel modules/images.

jwakre: Thanks for the feedback. In that case, this PR also fixes the behavior of ghosted files. When a manifest with ghosted files is updated, the ghosted file entries will be dropped in the new manifest.

phmccarty: Great! Sounds good to me.

PR #689, opened 2020-01-02
Go mod including tablewriter package

reaganlo:
Updating PR #647 to include the `tablewriter` package.


Makefile
```diff
@@ -1,49 +1,23 @@
...
-build: gopath
+build:
 	go install -ldflags="-X ${GO_PACKAGE_PREFIX}/builder.Version=${VERSION}" ${GO_PACKAGE_PREFIX}/mixer
 	go install ${GO_PACKAGE_PREFIX}/mixin
 	go install ${GO_PACKAGE_PREFIX}/swupd-extract
 	go install ${GO_PACKAGE_PREFIX}/swupd-inspector
 	go install ${GO_PACKAGE_PREFIX}/mixer-completion
 
-install: gopath
+install:
```
jwakre: `make install` succeeds when I set a GOPATH, but it fails when the GOPATH is unset.

By default, the go environment uses a GOPATH of `$HOME/go`, so `go install.....` will create the executable at `$HOME/go/bin/mixer`. The `install -m.....` command is not a go command, so it does not use the same default GOPATH. As a result, the executable is created at `$HOME/go/bin/mixer`, but the when installing `/bin/mixer` is used as the source path.

I suggest adding a default value for GOPATH: `GOPATH ?= ${HOME}/go`

reaganlo: @jwakre I've made the suggested change including doing a `build` while installing to take care of the sudo requirement. Let me know if this works..

Makefile
```diff
@@ -39,6 +38,10 @@ checkcoverage:
 
 .PHONY: lint
 lint:
+	@if ! $(shell go env GOPATH)/bin/golangci-lint --version &>/dev/null; then \
+		echo "Installing linters..."; \
+		curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.22.2; \
```
jwakre: This installs the linter successfully, but running the linter fails because `@golangci-lint run --deadline=10m --tests --disable-all \` cannot find the binary.

We should use `$(shell go env GOPATH)/bin/golangci-lint` when calling the linter so that the path can be consistently resolved. Also, I think it would make sense to have a variable for `$(shell go env GOPATH)/bin/`.

rchiossi: this commit really seems out of place. This PR was intended just to move mixer to go mod, not change the behavior of make targets.

reaganlo: @jwakre good catch.. I had my PATH set to my home dir... Hence was working for me.. Will make the change in another PR..

reaganlo: Removing the lint commit and merging..

PR #693, opened 2020-01-06
Add set-priority to repo command

### LINKED ISSUE #691: CLI: Set repo priority
Mixer should allow users to set the priority of the package repos that it manages. Below is the proposed CLI format for setting repo priority:

`mixer repo set-priority <repo> <num>`

This CLI follows a format similar to the existing `mixer repo set-url <repo> <url>` command.

jwakre:
Fixes #691 

* Add set-priority to repo command: `mixer repo set-priority <repo> <priority>`
* Increase code re-use for repo commands

### DIFF
--- a/builder/repo_control.go
+++ b/builder/repo_control.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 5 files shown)


ashleshaAtrey:
Should we also have optional priority parameter for add repo command ? This way user will not need to run two commands for setting the priority.

ashleshaAtrey:
As an enhancement it will be good to list the priority of repos in "repo list " command to verify the changes and to check priorities of existing repos.

jwakre:
@ashleshaAtrey 

To address your feedback, I added a `--priority` flag to `mixer repo add` and `mixer repo list` now includes the priority in the output.

builder/repo_control.go
```diff
@@ -351,7 +354,13 @@ func (b *Builder) ListRepos() error {
 			pURL.Scheme = "file"
 		}
 
-		fmt.Printf("%s\t%s\n", name, pURL.String())
+		priority := s.Key("priority").Value()
+		if priority == "" {
+			// When the priority field is omitted, the DNF default is 99.
+			priority = "99"
```
ashleshaAtrey: Should we keep dnf default and our default priority same ?

jwakre: From the perspective of a downstream mix, I prefer the current default behavior where the priority defaults to 1. This allows repos added by the user to override the upstream clear repo by default.

jwakre:
@reaganlo The `mixer repo add` command now prints the priority.

PR #695, opened 2020-01-08
Remove deleted upstream bundles when crossing upstream format

### LINKED ISSUE #694: Downstream format bumps fail when upstream deletes bundles
When crossing an upstream format bump with a bundle deletion as a downstream mix, the deleted bundle will not be deleted in the downstream mix. Additionally, future mixer commands will fail because the deleted upstream bundle definition will no longer exist.

**Steps to reproduce:**
```
mixer init --upstream-version 29690
mixer bundle add kernel-container
sudo mixer build all
mixer versions update --upstream-version 29800
sudo mixer build upstream-format
sudo mixer build all
2019/06/06 13:22:23 ERROR: Couldn't build bundles: Error building bundles: Cannot find bundle "kernel-container" in local or upstream bundles
```

**Expected Behavior:**
The kernel-container bundle was deleted in the upstream mix, so the downstream mix should delete the bundle and remove it from the bundle list when crossing the upstream format bump.

jwakre:
Fixes #694 

* When crossing an upstream format that deletes a bundle, the downstream mix will now also delete the bundle.
* Updated upstream-format-bump test to validate proper behavior for upstream/local bundle deletion.

### DIFF
--- a/bat/tests/upstream-format-bump/run.bats
+++ b/bat/tests/upstream-format-bump/run.bats
@@ -8,15 +8,26 @@ setup() {
 }
 
 @test "Perform format bump builds to cross upstream format bump" {
-  mixer-init-stripped-down 25740 10
+  # The kernel-container bundle is deleted in the format bump from 27 to 28
+  # and is used to test upstream bundle deletion. Version 29690 is in format
+  # 27 and version 29800 is in format 28.
+  mixer-init-stripped-down 29690 10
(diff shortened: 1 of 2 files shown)


ashleshaAtrey:
@jwakre the new bat test is failing in my environment

jwakre:
> @jwakre the new bat test is failing in my environment

The new bat test is passing for me locally and on Travis. Can you double check that the correct Mixer executable is used?

PR #696, opened 2020-01-09
Error Handling on channels

### LINKED ISSUE #673: Error handling using channels
Currently when we are catching errors using channels and select
statements, we are breaking from select loop but not the "for" loop the
"select" statement is in. The correct approach will be to return error
as soon as error channel is written.

ashleshaAtrey:
Currently while catching errors using channels and select
statements, when the error is caught in select statements, it
is not breaking from the outside loop.
The correct approach is to let the existing go routines finish
executing and return one of the errors caught by the error channel.
It should also close the channel in order to stop executing new go
routines.
fixes #673

### DIFF
--- a/swupd/fullfiles.go
+++ b/swupd/fullfiles.go
@@ -69,6 +69,7 @@ func CreateFullfiles(m *Manifest, chrootDir, outputDir string, numWorkers int, c
 	// Used by the task runners to indicate an error happened. The channel is buffered to ensure
 	// that all the goroutines can send their failure and finish.
 	errorCh := make(chan error, numWorkers)
+	defer close(errorCh)
 
 	infos := make([]FullfilesInfo, numWorkers)
 	for i := range infos {
@@ -77,6 +78,7 @@ func CreateFullfiles(m *Manifest, chrootDir, outputDir string, numWorkers int, c
 	}
 
 	taskRunner := func(info *FullfilesInfo) {
+		defer wg.Done()
 		for f := range taskCh {
 			var tErr error
 			input := filepath.Join(chrootDir, f.Name)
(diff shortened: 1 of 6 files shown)


swupd/manifest.go
```diff
@@ -598,6 +598,7 @@ func addAllManifestFiles(manifests []*Manifest, ui UpdateInfo, c config, numWork
 		defer wg.Done()
 		for m := range mCh {
 			if err := m.addManifestFiles(ui, c); err != nil {
+				fmt.Println(err)
```
jwakre: What was the reason for adding `fmt.Println(err)` in several locations in this PR? Since we are returning an error, I would expect the error message to eventually be printed again.

ashleshaAtrey: I added it for debugging purpose. We are only catching the first failure. As we are running multiple go routines, logging the errors in other go routines.

swupd/fullfiles.go
```diff
@@ -124,20 +125,18 @@ func CreateFullfiles(m *Manifest, chrootDir, outputDir string, numWorkers int, c
...
 
 	// Sending loop might finish before any goroutine could send an error back, so check for
 	// error again after they are all done.
-	if err == nil && len(errorCh) > 0 {
+	if len(errorCh) > 0 {
 		err = <-errorCh
-	}
-
-	if err != nil {
 		return nil, err
```
jwakre: We can make this one less line:
```
if len(errorCh) > 0 {
    err = <-errorCh
    return nil, err
}
```
```
if len(errorCh) > 0 {
    return nil, <-errorCh
}
```

ashleshaAtrey: fixed it !

ashleshaAtrey:
@reaganlo  updated the commit message

reaganlo:
@ashleshaAtrey Thank you.

PR #699, opened 2020-01-10
Improve Error Message for unresolve packages

### LINKED ISSUE #697: If a bundle has an unresolvable package, the package name isn't printed
When running mixer build bundles, if the bundle definition file contains a package name that dnf isn't able to resolve the build fails but without printing which package in a bundle that couldn't be resolved.

ashleshaAtrey:
while build bundles, if build fails to resolve packages, improve error
message by printing which packages were not found.
fixes #697

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -265,10 +265,20 @@ func parseNoopInstall(installOut string) ([]packageMetadata, error) {
 	parts := strings.Split(installOut, "Installing:\n")
 	if len(parts) < 2 {
 		// If there is no such section, e.g. dnf fails because there's
-		// no matching package (so no "Installing:"), return nil. Real
-		// failure will happen later when doing the actual install.
-
-		return nil, fmt.Errorf("unable to resolve package")
+		// no matching package (so no "Installing:"), return error with
+		// list of packages not found.
+		var pkgs string
+		if len(parts) == 1 {
+			parts = strings.Split(parts[0], "No match for argument:")
+			parts = parts[1:]
+			for i, v := range parts {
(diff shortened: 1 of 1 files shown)


builder/bundles.go
```diff
@@ -267,8 +267,11 @@ func parseNoopInstall(installOut string) ([]packageMetadata, error) {
 		// If there is no such section, e.g. dnf fails because there's
 		// no matching package (so no "Installing:"), return nil. Real
```
reaganlo: Can you please update the legacy comment to reflect the current behavior.

builder/bundles.go
```diff
@@ -267,8 +267,11 @@ func parseNoopInstall(installOut string) ([]packageMetadata, error) {
...
 		// failure will happen later when doing the actual install.
-
-		return nil, fmt.Errorf("unable to resolve package")
+		if len(parts) == 1 {
+			parts = strings.Split(parts[0], "No match for argument:")
+			parts = parts[1:]
+		}
+		return nil, fmt.Errorf("unable to resolve package %s", parts)
```
reaganlo: Can we format the output:
**Currently**:
```
unable to resolve package [ google-chrome-stable
  test-pkg
]
```
**Suggested**:
`unable to resolve package(s): google-chrome-stable, test-pkg`

builder/bundles.go
ashleshaAtrey: updated

builder/bundles.go
ashleshaAtrey: updated

reaganlo:
Updated commit message and merging.

PR #700, opened 2020-01-13
build: Add root permission check

### LINKED ISSUE #690: Improve error handling when lacking permissions
Some Mixer commands require sudo, but the failures when sudo is not used can be misleading. For instance, `mixer build bundles` will fail to resolve packages, but it does not mention that this issue was caused by a lack of permissions. When Mixer commands require sudo, they should fail earlier with a more informative message.

jwakre:
Fixes #690

* Add root permission check to `mixer build *` commands.

### DIFF
--- a/mixer/cmd/build.go
+++ b/mixer/cmd/build.go
@@ -140,6 +140,10 @@ var buildBundlesCmd = &cobra.Command{
 	Short:   "Build the bundles for your mix",
 	Long:    `Build the bundles for your mix`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if err := checkRoot(); err != nil {
+			fail(err)
+		}
+
 		b, err := builder.NewFromConfig(configFile)
 		if err != nil {
 			fail(err)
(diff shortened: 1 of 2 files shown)


rchiossi:
What about a `checkroot()` function instead?

reaganlo:
I thought the checkroot() would be:
```
func checkRoot() {
	if uid := os.Geteuid(); uid != 0 {
		fail(errors.Errorf("This command requires root (uid=%d)\n", uid))
	}
}
```
and usage would just be calling `checkRoot()`

jwakre:
> I thought the checkroot() would be:
> 
> ```
> func checkRoot() {
> 	if uid := os.Geteuid(); uid != 0 {
> 		fail(errors.Errorf("This command requires root (uid=%d)\n", uid))
> 	}
> }
> ```
> 
> and usage would just be calling `checkRoot()`

I thought about that, but I felt that it makes the code harder to read because it is not clear in the calling function that it can stop exection.

reaganlo:
cool.. fair enough..

PR #707, opened 2020-01-23
Making a minimal mix

### LINKED ISSUE #704: Create a minimal mix
Currently, the “os-core” bundle and its “filesystem” package is a hard requirement in Mixer. In order to enable a minimal mix for 3rd party, make the “filesystem” package optional. In case the “filesystem” package is found in any of the mix bundles, then Mixer should install the “filesystem” package first. If it’s not found in any of the mix bundles, then just don’t install it. 

The “os-core” bundle should still be a hard requirement in Mixer as it is required by swupd-client.
If the "os-release" file is not found, then Mixer should create and inject a default one.

ashleshaAtrey:
Currently, the “os-core” bundle and its “filesystem” package is a hard requirement in Mixer.
In order to enable a minimal mix for 3rd party, “filesystem” package is made optional.
In case the “filesystem” package is found in any of the mix bundles,
then Mixer should install the “filesystem” package first.
If it’s not found in any of the mix bundles, then filesystem is not
installed.

The “os-core” bundle is still a hard requirement in Mixer.
If the "os-release" file is not found, then Mixer creates and injects a default one.
fixes #704

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -420,12 +420,9 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 					rpmName := pkg.name + "-" + pkg.version + "." + pkg.arch + ".rpm"
 					bundle.AllPackages[pkg.name] = true
 					bundle.AllRpms[rpmName] = pkg
-
 					// to find the pkg Metadata of filesystem rpm, as filesystem needs to be extracted first
-					if bundle.Name == "os-core" {
-						if pkg.name == "filesystem" {
-							fileSystemInfo = pkg
-						}
+					if pkg.name == "filesystem" && fileSystemInfo == (packageMetadata{}) {
+						fileSystemInfo = pkg
 					}
 				}
 			}
@@ -535,8 +532,8 @@ func buildOsCore(b *Builder, packagerCmd []string, chrootDir, version string) er
 		return err
 	}
 
-	if err := fixOSRelease(b, filepath.Join(chrootDir, "usr/lib/os-release"), version); err != nil {
-		return errors.Wrap(err, "couldn't fix os-release file")
+	if err := updateOSReleaseFile(b, filepath.Join(chrootDir, "usr/lib/os-release"), version, b.Config.Swupd.ContentURL); err != nil {
+		return errors.Wrap(err, "couldn't update os-release file")
 	}
 
 	if err := createVersionsFile(filepath.Dir(chrootDir), packagerCmd); err != nil {
@@ -813,8 +810,10 @@ func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVers
 	i := 0
 	rpmMap = make(map[string]bool)
 
-	if err := installFilesystem(fullDir, packagerCmd, downloadRetries, b.repos); err != nil {
-		return err
+	if fileSystemInfo != (packageMetadata{}) {
+		if err := installFilesystem(fullDir, packagerCmd, downloadRetries, b.repos); err != nil {
+			return err
+		}
 	}
 
 	for _, bundle := range *set {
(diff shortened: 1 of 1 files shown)


reaganlo:
Instead of failing if the "os-release" file is not found, can Mixer create and inject a default one. I have updated the #704. Also please check if there are other hard requirements.

ashleshaAtrey:
@reaganlo made the changes. Ready for review.

builder/bundles.go
```diff
@@ -535,7 +532,7 @@ func buildOsCore(b *Builder, packagerCmd []string, chrootDir, version string) er
 		return err
 	}
 
-	if err := fixOSRelease(b, filepath.Join(chrootDir, "usr/lib/os-release"), version); err != nil {
+	if err := fixOSRelease(b, filepath.Join(chrootDir, "usr/lib/os-release"), version, chrootDir); err != nil {
```
reaganlo: Its not part of this change, but can we call this function `updateOSReleaseFile()`?
Lets also add a print statement inside the function, `fmt.Println("... updating os-release file")`

builder/bundles.go
```diff
@@ -1181,3 +1187,65 @@ func getClosestAncestorOwner(path string) (int, int, error) {
...
+	}
+	_, err = fmt.Fprintln(f, "ID="+"clear-linux-os")
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(f, "ID_LIKE="+"clear-linux-os")
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(f, "VERSION_ID="+"")
```
reaganlo: In the case the os-release file is not found, its contents are created and written. Its then opened again to update the VERSION_ID. Instead can we just update the VERSION_ID here itself.
So maybe instead of calling `createOSReleaseFile()` inside `updateOSReleaseFile()`, we can call it inside `buildOsCore()`. Something like this:

```
	var filename string
	if b.Config.Mixer.OSReleasePath != "" {
		filename = b.Config.Mixer.OSReleasePath
	} else {
		filename = filepath.Join(chrootDir, "usr/lib/os-release")
	}

	if _, err := os.Stat(filename); err != nil {
		if os.IsNotExist(err) {
			if err = createOSReleaseFile(filename, version, chrootDir); err != nil {
				return errors.Wrap(err, "couldn't create os-release file")
			}
		} else {
			return err
		}
	} else {
		if err := updateOSReleaseFile(b, filename, version, chrootDir); err != nil {
			return errors.Wrap(err, "couldn't update os-release file")
		}
	}
```

builder/bundles.go
```diff
@@ -1181,3 +1187,65 @@ func getClosestAncestorOwner(path string) (int, int, error) {
...
+	}
+	_, err = fmt.Fprintln(f, "PRETTY_NAME="+"\""+"Clear Linux OS"+"\"")
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(f, "ANSI_COLOR="+"\""+"1;35"+"\"")
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(f, "HOME_URL="+"\""+chrootDir+"\"")
```
reaganlo: I think @bryteise mentioned the default value for "HOME_URL" to be the `Swupd.CONTENTURL` in the builder.conf

builder/bundles.go
ashleshaAtrey: @reaganlo updated the createOSReleaseFile to update version_ID as well.
if createOSReleaseFile is successful, it return from function.
Still calling  createOSReleaseFile from UpdateOSReleaseFile.

builder/bundles.go
```diff
@@ -1123,14 +1122,19 @@ func createVersionsFile(baseDir string, packagerCmd []string) error {
...
 	var err error
 	var f *os.File
 	if b.Config.Mixer.OSReleasePath != "" {
 		f, err = os.Open(b.Config.Mixer.OSReleasePath)
 	} else {
+		_, err = os.Stat(filename)
+		if err != nil {
+			err = createOSReleaseFile(filename, homeURL, version)
+			return err
+		}
```
reaganlo: Shouldn't we be checking for `os.IsNotExist(err)`
Something like:
```
		if _, err = os.Stat(filename); os.IsNotExist(err) {
			return createOSReleaseFile(filename, homeURL, version)
		}
		if err != nil {
			return err
		}
```

ashleshaAtrey: yes, will fix it.

reaganlo:
Updating git commit message and merging

PR #709, opened 2020-01-27
Build bundles from content chroots

### LINKED ISSUE #705: Enable building bundles from its contents
“mixer build bundles” should be able to process a new "content()" keyword in the bundle definition file that specifies a relative/absolute path to the top level content directory. E.g. content(\<dir\>)

Once the full chroot is built, the content should be injected into the full chroot after conflict checks. If there are conflicts, then error out. Generate the bundle info files from the final chroot as usual.

jwakre:
Fixes #705 

* Add "content" field to the bundle definition file
* Resolve chroots specified by the "content" field with the full chroot and add their files to the corresponding bundle-info file.
* Add tests for the new features
* Remove bundle chroot tutorial.
* Increase versatility of Mixer's copy implementation

Workflow:
1. Create/edit a bundle definition to use the "content" keyword
   - content(/path/to/contentChroot)
2. `sudo mixer build all`
3. Verify that the full chroot, bundle-info file, and manifest were generated correctly


builder/bundles.go
```diff
@@ -840,6 +841,79 @@ func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVers
 		}
 	}
 
+	// Resolve bundle content chroots against the full chroot. New files are copied
+	// to the full chroot and the bundle file lists are updated to contain the files
+	// within the chroot.
+	return addBundleContentChroots(set, fullDir)
```
reaganlo: In case the user provides the os-core bundle also as content that contains the os-release file, should we call `buildOsCore()` and `genUpdateBundleSpecialFiles()` after `addBundleContentChroots()`. i.e take them outside the for loop and have something like this:

```
	if err := addBundleContentChroots(set, fullDir); err != nil {
		return err
	}

	// special handling for os-core
	fmt.Println("... building special os-core content")
	if err := buildOsCore(b, packagerCmd, fullDir, version); err != nil {
		return err
	}

	// special handling for update bundle
	fmt.Printf("... Adding swupd default values to %s bundle\n", b.Config.Swupd.Bundle)
	if err := genUpdateBundleSpecialFiles(fullDir, b); err != nil {
		return err
	}

	return err
```

jwakre: Are you saying that users should be allowed to override these special case files and prevent Mixer from auto-generating them?

reaganlo: What I meant is, in case the user provides the special case files, then it gets updated by Mixer after the content chroot is created.

reaganlo: As discussed, lets give the content more priority. So if the special case files exist as content, then do not overwrite them. The fields in os-release should still be updated.

jwakre:
@reaganlo As discussed, Mixer will not generate or modify special case files when they exist in content chroots with the exception of the os-release file. The os-release file will still be updated to have a BUILD_ID field and any existing VERSION_ID field will be updated.

builder/bundles.go
reaganlo: > In case the user provides the os-core bundle also as content that contains the os-release file, should we call `buildOsCore()` and `genUpdateBundleSpecialFiles()` after `addBundleContentChroots()`. i.e take them outside the for loop and have something like this:

Please ignore this comment and let Mixer fail if user provides conflicting special case files as contents. The reason being:
Swupd_Root.pem is a special case file that is either created by the Mixer or specified by the user using the Builder.CERT config. This certificate file is used by Mixer for signing. So if we let the user overwrite it as a content file, it will lead to an inconsistent mix as the certificate used and certificate in the mix will be different. One way to handle this is to treat this special case file as even more special. But lets not make it more complicated and hard to maintain. @jwakre Sorry for the additional work.

reaganlo: > As discussed, lets give the content more priority. So if the special case files exist as content, then do not overwrite them. The fields in os-release should still be updated.

Please ignore this comment and let Mixer fail if user provides conflicting special case files as contents. The reason being:
Swupd_Root.pem is a special case file that is either created by the Mixer or specified by the user using the Builder.CERT config. This certificate file is used by Mixer for signing. So if we let the user overwrite it as a content file, it will lead to an inconsistent mix as the certificate used and certificate in the mix will be different. One way to handle this is to treat this special case file as even more special. But lets not make it more complicated and hard to maintain. @jwakre Sorry for the additional work.

jwakre:
@reaganlo I dropped the previous changes. Now Mixer will fail when content chroots have special case file conflicts.

ashleshaAtrey:
is there a reason we are iterating through all bundles including content bundles while resolving packages and  installing bundle to full. While processing it gives wrong info.
for example:
processing chrome
Bundle chrome	5 files

jwakre:
> is there a reason we are iterating through all bundles including content bundles while resolving packages and installing bundle to full.

There was a decision to re-use the bundle definition file for content chroots instead of creating a separate content bundle concept. Since we decided to use the bundle definition file to specify content chroots, there is not a separation between content/package bundles. A single bundle definition can use both content chroots and packages.


> While processing it gives wrong info.
> for example:
> processing chrome
> Bundle chrome 5 files

I haven't looked closely at these numbers. I'll look into it.

builder/bundles.go
```diff
@@ -839,6 +840,79 @@ func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVers
...
+			if _, err := os.Stat(chrootPath); err != nil {
+				return err
+			}
+
+			err := filepath.Walk(chrootPath, func(path string, fi os.FileInfo, err error) error {
+				if err != nil {
+					return err
+				}
+
+				filePath := strings.TrimPrefix(path, chrootPath)
```
reaganlo: I think we should handle the scenario where the user provides the contents as `content(/dir/) `instead of the expected `content(/dir)`. Currently the mix will not fail but the Manifest will end up having files like `usr`, `usr/bin` instead of `/usr`, `/usr/bin`. This can cause errors in the swupd-client during installation.

builder/bundleset.go
```diff
@@ -355,6 +359,21 @@ func parseBundle(contents []byte) (*bundle, error) {
...
 			optional = append(optional, text)
+		} else if strings.HasPrefix(text, "content(") {
+			if !strings.HasSuffix(text, ")") {
+				return nil, fmt.Errorf("Missing end parenthesis in line %d: %q", line, text)
+			}
+			text = text[8 : len(text)-1]
+
+			// Remove unnecessary trailing slashes
+			for len(text) > 1 && strings.HasSuffix(text, "/") {
+				text = strings.TrimSuffix(text, "/")
```
jwakre: @reaganlo I added this condition to drop trailing slashes when parsing the bundle.

builder/bundleset_test.go
```diff
@@ -81,6 +99,28 @@ pkg1 # [TITLE]: wrongtitle
...
+		{
+			Contents: []byte(fmt.Sprintf(`# Duplicate content chroots and trailing /s
+# [TITLE]: fake
+# [DESCRIPTION]: a description
+# [STATUS]: a status
+# [CAPABILITIES]: the capabilities
+# [MAINTAINER]: the maintainer
+content(%s/)
+content(%s)
+content(%s///)
```
jwakre: @reaganlo I added this unit test to verify that content chroots can handle trailing slashes.

jwakre:
@reaganlo Now content chroots properly handle trailing slashes and I added a test case. Also, I fixed the typo in the commit.

PR #714, opened 2020-01-30
Always clean mix workspace before building bundles

### LINKED ISSUE #669: Ensure workspace is always clean before building content
Mixer should take reasonable precautions by default to ensure correctness of the manifests it produces.  Part of this should involve cleaning its workspace prior to doing work.  There is a `--clean` flag that defaults to false.  This prevents the `update/www/<ver>` and `update/image/<ver>` directories from being cleaned prior to use.  In the case a build is aborted and attempted again with the same version numbers, there is potential for not having correct data in these folders, leading to incorrect manifests.

jwakre:
Existing image and output directories for the current mix version will
be removed when building a mix. This prevents mixer from building on top
of directories contaminated by a prior mix.

Fixes #669

### DIFF
--- a/bat/tests/clean-rebuild/description.txt
+++ b/bat/tests/clean-rebuild/description.txt
@@ -1,4 +1,4 @@
 clean-rebuild
 ================
-This test rebuilds version 20 with the --clean flag. It confirms that the current
-and previous versions are correctly maintained when performing a clean build.
+This test rebuilds version 20 and confirms that the current/previous versions are correctly
+maintained when a build is cleaned.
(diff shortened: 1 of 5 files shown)


jwakre:
After offline discussion, I decided to only clean the current mix version with this PR. At this time, Mixer will not clean mix output ahead of the current mix version.

ashleshaAtrey:
I think clean-rebuild bat test uses the clean flag. We can remove the flag from that test

jwakre:
@ashleshaAtrey I removed the --clean flag from the tests. Also, I added a check to verify that `mixer build bundles` removes update/www/\<mixver\>.

PR #717, opened 2020-02-04
Print delta pack results together

### LINKED ISSUE #551: Output being interleaved during delta pack creation
When running with multiple workers, the output of creating delta packs (and likely other build commands) might get interleaved. E.g.

```
15:30:12   Creating delta pack for bundle "desktop-dev" from 28100 to 28120
15:30:12   Creating delta pack for bundle "v4l-utils" from 27770 to 28120
15:30:12   Creating delta pack for bundle "os-core" from 28100 to 28120
15:30:12     Fullfiles in pack: 0
15:30:12     Deltas in pack: 159
15:30:12     Fullfiles in pack: 2
15:30:12     Deltas in pack: 2
```

We likely should print the "Creating delta pack" message together with the summary at the end, or add a new output together with the summary at the end identifying the delta pack ("Finished delta pack for ...", then summary). The second option is for the case we care the work has started. IMHO both works fine.

As noted, this might also apply for other commands using workers.

jwakre:
Fixes #551

Delta packs are generated concurrently, so their results output must be
printed once to prevent interleaved results.

### DIFF
--- a/builder/deltapacks.go
+++ b/builder/deltapacks.go
@@ -83,10 +83,10 @@ func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printRepor
 
 				if len(info.Warnings) > 0 {
 					for _, w := range info.Warnings {
-						fmt.Printf("    WARNING: %s\n", w)
+						fmt.Fprintf(os.Stderr, "    WARNING: Bundle %s: %s\n", b.Name, w)
 					}
-					fmt.Println()
 				}
+				report := fmt.Sprintf("  Finished delta pack for bundle %q from %d to %d\n", b.Name, b.FromVersion, b.ToVersion)
 				if printReport {
 					max := 0
 					for _, e := range info.Entries {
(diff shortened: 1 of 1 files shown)


builder/deltapacks.go
```diff
@@ -81,11 +81,12 @@ func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printRepor
 					continue
 				}
 
+				report := fmt.Sprintf("  Finished delta pack for bundle %q from %d to %d\n", b.Name, b.FromVersion, b.ToVersion)
 				if len(info.Warnings) > 0 {
 					for _, w := range info.Warnings {
-						fmt.Printf("    WARNING: %s\n", w)
+						report += fmt.Sprintf("    WARNING: %s\n", w)
```
jwakre: I think it makes more sense to separate warnings from the report by writing them to stderr. @reaganlo Thoughts?

reaganlo: Makes sense.

jwakre: I updated the PR so that warning messages include the impacted bundle's name and are written to stderr.

PR #721, opened 2020-02-12
Clean dnf cache before using it

### LINKED ISSUE #719: Don't use host dnf cache when resolving files
The host dnf cache is used when resolving files which can sometimes lead to unexpected behavior. For instance, when maintaining a mix with local content, the stale cached content from the host system may be used which results in inconsistencies between the bundle info file and the full chroot.

An installroot can be used to avoid using the host's dnf cache.

jwakre:
Fixes #719

Existing dnf cache contents can cause dnf queries to retrieve stale
package metadata. Cleaning the dnf cache before using it prevents old
cache contents from altering query results.

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -779,11 +779,6 @@ var rpmMap map[string]bool
 var dnfDownloadDir string
 
 func buildFullChroot(b *Builder, set *bundleSet, packagerCmd []string, buildVersionDir, version string, downloadRetries int, numWorkers int) error {
-	fmt.Println("Cleaning DNF cache before full install")
-	if err := clearDNFCache(packagerCmd); err != nil {
-		return err
-	}
-
 	dnfDownloadDir = filepath.Join(buildVersionDir, "downloadedRpms")
 	err := os.MkdirAll(dnfDownloadDir, 0755)
 	if err != nil {
@@ -1029,6 +1024,12 @@ src=%s
 	}
 
 	fmt.Printf("Packager command-line: %s\n", strings.Join(packagerCmd, " "))
+
+	// Existing DNF cache content can cause incorrect queries with stale results
+	fmt.Println("Cleaning DNF cache")
+	if err := clearDNFCache(packagerCmd); err != nil {
+		return err
+	}
 
 	numWorkers := b.NumBundleWorkers
 	emptyDir, err := ioutil.TempDir("", "MixerEmptyDirForNoopInstall")


ashleshaAtrey:
Tested the changes with all upstream bundles. No discrepancies in the output. 
I was not able to recreate the issue #719.

jwakre:
My previous implementation that forced all dnf commands to execute within the full chroot's installroot was failing with a dnf lock contention error. Alternatively, I decided to clean the dnf cache earlier in the workflow. Since the dnf cache is cleaned before any dnf queries are run, cache content from prior mixes will not impact new queries.

ashleshaAtrey:
This will end up cleaning the system dnf cache files for that version right?

jwakre:
> This will end up cleaning the system dnf cache files for that version right?

I thought it was cleaning everything, but you are right this is only cleaning a specific version. The root cause of the problem may be some lingering temporary file that dnf clean removes.

I can update my comments and do more testing. Currently, I am able to reproduce #719 fairly consistently with clr-distro-factory and an internal repo. So far I have been unable to reproduce the failure with this PR, but I can run longer tests.

phmccarty:
Hmm, I doubt that `dnf clean all` will clean content only for a specific version. I think it cleans everything under the cache directory. If the cache directory was dependent on `$releasever`, then that might work.

jwakre:
> Hmm, I doubt that `dnf clean all` will clean content only for a specific version. I think it cleans everything under the cache directory. If the cache directory referenced the `$releasever` variable, then that might work.

Mixer is calling the following:
`dnf --config .yum-mix.conf -y --releasever <mixVer> clean all`

Looking on my system, it only cleans the specified version.

jwakre:
@ashleshaAtrey @phmccarty  After looking closer, I have more information on the problem:

Mixer does not version local content, so when incrementally building a mix with local content, dnf will continue to re-use the same local content cache. This is problematic when an rpm in the local content is updated to a new version. In this case, local content files will be resolved with old cached metadata, but the newer version of the rpm will be installed to the full chroot.

reaganlo:
> Mixer is calling the following:
> `dnf --config .yum-mix.conf -y --releasever <mixVer> clean all`
> 
> Looking on my system, it only cleans the specified version.

So is it possible to tweak the clean function to clean the entire cache by not specifying the --releasever?

jwakre:
> So is it possible to tweak the clean function to clean the entire cache by not specifying the --releasever?

The older versions of the clear repo are not problematic. The problem is re-using an existing version of a package repo, but with different content. This is easiest to reproduce by updating the versions of packages in the local repo because it does not have multiple versions.

PR #722, opened 2020-02-12
Export flag

### LINKED ISSUE #708: Introduce new export flag in Manifests
Mixer should introduce a new export flag in the Manifest files that specifies the binary files that should be exported. By default, Mixer should export the files with qualifiers "/bin","/usr/bin" and "/usr/local/bin". Users should be able to specify if certain files should not be exported.

In order to implement this requirement, the following changes could to be done:
- Build bundles can use the existing `Files map[string]bool` entry in the bundle info file to indicate exportable files. If exportable, the boolean value for the File will be set to `true`.
By default, the files with prefix "/bin","/usr/bin" and "/usr/local/bin" should be exportable. E.g.
`"Files":{"/bin/file1":true,"/usr/bin/file2":true","/usr/local/bin/file3":true}`

- Build bundles should be able to process a new "un-export()" keyword in the bundle definition file that specifies the binary files that should not be exported in a bundle. E.g.
`un-export(/bin/file1)`
`un-export(/usr/bin/file2)`

- Build bundles can set the un-exported file entries to `false` in its corresponding bundle Info. Based on the above examples, this would end up being:
`E.g. "Files":{"/bin/file1":false,"/usr/bin/file2":false","/usr/local/bin/file3":true}`

- Build update should add the exportable files to the Manifest file with an `x` in the 4th byte of the flag section of the Manifest. E.g.
`F..x         6a197bc89bacaa95ca55a9ed6cd6674f8512bba32fce61cf7ab1b5738bb41d7b       10           /usr/bin/file_3`

reaganlo:
- Change RenameFlag to MiscFlag
    Change RenameFlag to MiscFlag since "rename" has been deprecated and
    this flag will be used as a placeholder for various other flags.

- Introduce new export flag in manifests
    By default, files with the prefix "/bin/", "/usr/bin/" and "/usr/local/bin/"
    will have a new export flag "x" in the 4th byte of the flag section of the Manifest.
    The user can override this by using a new keyword "un-export()" in
    the bundle definition file.
    E.g. un-export(/usr/bin/file2)
    Fixes #708

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
 #    - TEST_DIR=build-validate
     - TEST_DIR=bundle-commands
     - TEST_DIR=clean-rebuild
+    - TEST_DIR=content-chroot
     - TEST_DIR=config-conversion
     - TEST_DIR=contentsize-check
 #    - TEST_DIR=create-mix-bump-version-add-remove-bundles
@@ -25,6 +26,7 @@ env:
     - TEST_DIR=create-mix-with-upstream-bundles
     - TEST_DIR=customize-os-release
     - TEST_DIR=clr-installer-config
+    - TEST_DIR=export-flag
     - TEST_DIR=manual-format-bump-flow
     - TEST_DIR=manual-upstream-format-bump-flow
     - TEST_DIR=mixin-repo-commands
--- a/bat/tests/export-flag/Makefile
+++ b/bat/tests/export-flag/Makefile
@@ -1,0 +1,9 @@
+.PHONY: check clean
+
+check:
+	bats ./run.bats
+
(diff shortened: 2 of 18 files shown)


ashleshaAtrey:
After a version update, if the exported flag changes, new Manifest is not being generated for that version.

reaganlo:
Thanks for testing.. will look into it..

reaganlo:
> After a version update, if the exported flag changes, new Manifest is not being generated for that version.

@ashleshaAtrey good catch. I've pushed a change and updated tests to handle this.

ashleshaAtrey:
bundle manifests looks good now. 
Manfiest.full are not getting updated though. Export flag for Manifest.full files are not getting generated properly.

reaganlo:
> Manfiest.full are not getting updated though. Export flag for Manifest.full files are not getting generated properly.

@ashleshaAtrey I figured out a way to populate the full manifest with the export flag, but then realized that the export flag is at a bundle-file level instead of at a file level. Hence it should not be updated in the full manifest. Just confirmed with @bryteise

ashleshaAtrey:
I tested the PR integrating with swupd.
 LGTM

PR #726, opened 2020-02-25
Write mixbundles file once for resolved bundle set

### LINKED ISSUE #724: Improve the scalability of getFullMixBundleSet
The `getFullMixBundleSet` function loads the bundle set many unnecessary times and writes many unnecessary intermediate results to disk. These unnecessary operations create a performance bottleneck that limits Mixer's ability to scale to a large number of bundles.

jwakre:
When resolving the full mix bundle set, the mixbundles file is now
updated once with the finalized set of bundles. Previously, the
mixbundles file was written many times with intermediate results which
scaled poorly when building with many bundles.

Fixes #724

### DIFF
--- a/builder/bundle_control.go
+++ b/builder/bundle_control.go
@@ -279,6 +279,12 @@ func (b *Builder) getFullBundleSet(bundles bundleSet) (bundleSet, error) {
 	recurseBundleSet = func(bundles []string) error {
 		for _, bName := range bundles {
 			if _, exists := set[bName]; !exists {
+				// validate names of included bundles
+				err := validateBundleName(bName)
+				if err != nil {
+					return err
+				}
+
 				bundle, err := b.getBundleFromName(bName)
 				if err != nil {
 					return err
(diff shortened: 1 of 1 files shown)


builder/bundle_control.go
```diff
@@ -341,16 +346,11 @@ func (b *Builder) getFullMixBundleSet() (bundleSet, error) {
...
-		if err != nil {
-			return nil, err
-		}
-		err = b.AddBundles(bundle.OptionalIncludes, false, false, false)
-		if err != nil {
-			return nil, err
-		}
+
+	// Write final mix bundle list back to file
+	err = b.writeMixBundleList(set)
```
jwakre: The set of bundles is already recursively resolved with includes and also-adds, so it is safe to directly write the result to disk.

With this implementation, bundles added to the mix by dependencies will be added silently without prints to stdout. I think the print dump to stdout can be overwhelming and hard to read as the number of bundles drastically increases. Maybe it makes more sense to improve Mixer's ability to log and write this information to a log file.

PR #728, opened 2020-02-25
Preserve file ownership when using content chroots

### LINKED ISSUE #727: Content chroots do not maintain file ownership
When using content chroots, files copied to the full chroot are owned by root which is not desirable. Files copied to the full chroot should maintain their original file ownership.

jwakre:
When using content chroots, file uid/gid values should be preserved when
copying them to the full chroot.

Fixes #727


bryteise:
LGTM

PR #732, opened 2020-02-28
Initialize local yum repo

### LINKED ISSUE #731: Compatibility with latest DNF
Newer versions of DNF require that all declared yum repos are *initialized*, even if they are empty. So, for mixer, the `local-yum` directory should be initialized with `createrepo_c` even if it is empty (i.e. contains no rpms).

jwakre:
Newer versions of DNF require that yum repos are initialized, so Mixer
must initialize the local repo, even when it's empty.

Fixes #731

### DIFF
--- a/builder/repo_control.go
+++ b/builder/repo_control.go
@@ -22,6 +22,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


builder/repo_control.go
```diff
@@ -130,6 +131,17 @@ func (b *Builder) NewDNFConfIfNeeded() error {
 			return errors.Wrapf(err, "Failed to write to dnf file: %s", b.Config.Builder.DNFConf)
 		}
 	}
+
+	if b.Config.Mixer.LocalRepoDir != "" {
+		localRepo := path.Join(b.Config.Mixer.LocalRepoDir, "repodata")
```
jwakre: This is fragile, but I didn't want to waste time re-creating a the yum repo for each build. Open to suggestions here.

On a side note, I noticed that `mixer build all` does re-create the yum repo each time.

PR #734, opened 2020-02-28
Set file perms after creating the dest file

jwakre:
When calling the copyFileWithFlags function with the preserveSrc flag
set, file permissions must be set after creating the destination file.
This is necessary because umask can prevent newly created files from
acquiring the source file's permissions.


jwakre:
@reaganlo The setuid/setgid permissions were lost when changing file ownership. This is resolved by changing file ownership before setting the permissions.

I made the following changes:

* Add support for setuid/setgid file permissions by setting file ownership before file permissions.
* Add permission support to directories that takes into account umask and setuid/setgid.
* Adjust bat test to add file permissions (default umask on Clear cannot create files with 777 permissions), check setuid/setgid permissions.
* Added for loop to bat test's permission checks to reduce duplicated code.

PR #742, opened 2020-03-05
Install special case files last

### LINKED ISSUE #738: Directory permission mismatch without filesystem package
When the os-core bundle does not include the filesystem package, it can create directories with potentially undesirable permissions.

1. Create mix that only uses content chroot bundles (no packages).
2. In one of the content chroots, create the "usr/lib" directory with 0755 which are the permissions used by the filesystem package.

Mixer will create the usr/lib directory in the full chroot with 0655 permissions which will conflict with the content chroot.

jwakre:
When adding special case files to the full chroot, they will create
missing directories in their path with permissions that may conflict
with content that will be added at a later point. By adding special case
files last, they will not create directories with potentially
conflicting permissions.

Fixes #738

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -759,13 +759,7 @@ func installBundleToFull(packagerCmd []string, baseDir string, bundle *bundle, d
 		return <-errorCh
 	}
 
-	bundleDir := filepath.Join(baseDir, "usr/share/clear/bundles")
-	err = os.MkdirAll(filepath.Join(bundleDir), 0755)
-	if err != nil {
-		return err
-	}
-
-	return ioutil.WriteFile(filepath.Join(bundleDir, bundle.Name), nil, 0644)
+	return nil
 }
 
 func clearDNFCache(packagerCmd []string) error {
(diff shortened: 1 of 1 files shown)


reaganlo:
So if I understand correctly, the special directories are:
- /usr/share/clear/bundles/
- /usr/share/clear/
- /usr/lib/
- /usr/share/defaults/swupd/
- /usr/share/clear/update-ca/

And with this change, the owners/perms of these directories will be created in this order:
1. packages
2. contents
3. mixer (root)

@jwakre correct?

jwakre:
> So if I understand correctly, the special directories are:
> 
>     * /usr/share/clear/bundles/
> 
>     * /usr/share/clear/
> 
>     * /usr/lib/
> 
>     * /usr/share/defaults/swupd/
> 
>     * /usr/share/clear/update-ca/
> 
> 
> And with this change, the owners/perms of these directories will be created in this order:
> 
>     1. packages
> 
>     2. contents
> 
>     3. mixer (root)
> 
> 
> @jwakre correct?

Yes. So with this change, content chroots would be able to alter ownership and permissions for the above listed directories.

Alternatively, if we do not want users to have this level of control, we can leave this code the same. In this case, it would make sense to avoid checking permissions and ownership for these files merging content chroot files into the full chroot.

bryteise:
The ordering makes sense to me. Is there a failure when a conflict happens between packages and chroots? Then the special case directories/files are created if they don't exist (so conflict is okay, just don't do anything).

jwakre:
> The ordering makes sense to me. Is there a failure when a conflict happens between packages and chroots?

Yes, Mixer will fail whenever there is a conflict between packages and content chroots.

> Then the special case directories/files are created if they don't exist (so conflict is okay, just don't do anything).

Yes, special case directories will be created with Mixer's default ownership/permissions only when they do not already exist. When packages or content chroots previously create these directories, Mixer special case files will re-use the existing directories without checking for conflicting ownership/permissions.

bryteise:
Sounds great to me.

reaganlo:
@bryteise @jwakre Cool. So sounds like this PR is the desired behavior. I will test it and merge it in today's release.

PR #747, opened 2020-03-11
Parallelize bundle file resolving

### LINKED ISSUE #730: Improve parallel CPU utilization for package/file resolution
Mixer parallelizes package/file resolution, but it only uses a single dnf installroot. This creates a serialized bottleneck because dnf only allows one process to concurrently access the dnf cache. By creating a separate dnf installroot for each goroutine, package and file resolution can better utilize multiple cores.

jwakre:
By creating a separate installroot for each package resolving goroutine,
dnf can run in parallel without contention on the lock file when
accessing the dnf cache. Now, package and file resolution occurs within
the same installroot so that the dnf cache that was fetched when resolving
packages can be re-used when resolving files.

Fixes #730

### DIFF
--- a/builder/bundles.go
+++ b/builder/bundles.go
@@ -190,75 +190,6 @@ func resolveFilesForBundle(bundle *bundle, repoPkgs repoPkgMap, packagerCmd []st
 	return nil
 }
 
-func resolveFiles(numWorkers int, set bundleSet, bundleRepoPkgs *sync.Map, packagerCmd []string) error {
-	var err error
-	var wg sync.WaitGroup
-	fmt.Printf("Resolving files using %d workers\n", numWorkers)
-	wg.Add(numWorkers)
-	bundleCh := make(chan *bundle)
-	// buffer errorCh so it always has space
-	// only one error can be returned to this channel per worker so
-	// buffering to numWorkers will make sure we always have space for the
-	// errors.
-	errorCh := make(chan error, numWorkers)
-	defer close(errorCh)
-
-	fileWorker := func() {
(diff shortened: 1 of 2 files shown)


jwakre:
Timing with 8 workers

Master timing:
* package resolve: 4m24.838s
* file resolve: 3m38.533s
* MCA with all upstream content (local packages): 22 min 

PR timing:
* package and file resolve: 5m33.625s
* MCA with all upstream content (local packages): 15 min

builder/bundles.go
```diff
@@ -452,7 +392,15 @@ func resolvePackages(numWorkers int, set bundleSet, packagerCmd []string, emptyD
 
 			bundleRepoPkgs.Store(bundle.Name, rpm)
```
ashleshaAtrey: Are we are using bundleRepoPkgs anywhere? I think we can skip creating it.

ashleshaAtrey: ohk, we are using it during build validation.

ashleshaAtrey: we can still populate the bundleRepoPkgs only for validation step and save on memory for normal builds.

reaganlo: @jwakre ??

jwakre: I will make a small, quick change to so that bundleRepoPkgs is not populated when calling `mixer build bundles`.

jwakre:
@ashleshaAtrey @reaganlo I updated the PR so that the map is only populated when performing the MCA build check.

reaganlo:
+1 Thanks for making the change..

PR #748, opened 2020-03-19
Handle MCA bundle deletion during format bump

### LINKED ISSUE #743: MCA fails to handle bundle deletion
MCA fails when validating versions with a bundle deletion. Steps to reproduce:
```
./afb.sh
cd afb-test
sudo mixer build validate --from 20 --to 30

ERROR: Failed to load bundle foo from map
```

ashleshaAtrey:
when bundle is deleted from a mix, bundle info file is not generated for that
bundle. When parsing the manifest if bundle info is empty, consider that
bundle to be deleted in the mix and do not resolve packages for that
bundle during build validation process. 

fixes #743

### DIFF
--- a/builder/build_validate.go
+++ b/builder/build_validate.go
@@ -250,7 +250,11 @@ func (b *Builder) mcaManInfo(version int) ([]*swupd.Manifest, error) {
 		if err != nil {
 			return nil, err
 		}
-
+		// skip bundle if bundleInfo is empty for that bundle Manifest, it indicates the bundle is deleted and should
+		// not be added to Manifests list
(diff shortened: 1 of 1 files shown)


builder/build_validate.go
```diff
@@ -250,7 +250,11 @@ func (b *Builder) mcaManInfo(version int) ([]*swupd.Manifest, error) {
 		if err != nil {
 			return nil, err
 		}
-
+		// skip bundle if bundleInfo is empty for that bundle Manifiest, it indicates the bundle is deleted and should
```
reaganlo: Minor comment. Typo.. `Manifiest`

PR #752, opened 2020-04-02
Make sure all DNF errors are printed

phmccarty:
In addition to printing a summary of unresolvable package names whenever
DNF fails, also print error messages for any "unknown" DNF error, and
finally print the full stdout/stderr from failed DNF commands for later
inspection.


phmccarty:
V2 uploaded, with a revision to a code comment.

PR #757, opened 2020-04-06
Remove the mixin feature

reaganlo:
Now mixer has the ability to support 3rd party content, which makes the
need for mixin obsolete.


reaganlo:
@bktan8 FYI

PR #767, opened 2020-05-05
Fix generated warning for delta packs

### LINKED ISSUE #628: Mixer delta pack generation warning
Steps to reproduce:
1. Use the RPMs and bundles from 31290 for a minversion build 10.  Only generate manifests for os-core and os-core-update.
2. Use the RPMs and bundles from 31380 to do an update build 20.  Only generate manifests for os-core and os-core-update.
3. Attempt to generate delta packs between build 10 and 20.
4. See the following warning
```
WARNING: open /srv/mixer/gtkramer-clr-mainline/update/www/20/delta/10-20-9e3f14dcff0e825d3cc30482d6c6ecc28925744e09afd0a26a1a139fabeb7789-9e3f14dcff0e825d3cc30482d6c6ecc28925744e09afd0a26a1a139fabeb7789: no such file or directory
```

In the Manifest.full of 10 (31290), we have:
```
F...    9e3f14dcff0e825d3cc30482d6c6ecc28925744e09afd0a26a1a139fabeb7789        10      /usr/share/gdb/auto-load/usr/lib32/libglib-2.0.so.0.6200.1-gdb.py
F...    9e3f14dcff0e825d3cc30482d6c6ecc28925744e09afd0a26a1a139fabeb7789        10      /usr/share/gdb/auto-load/usr/lib64/libglib-2.0.so.0.6200.1-gdb.py
```

And in the Manifest.full of 20 (31380), we have:
```
F...    9e3f14dcff0e825d3cc30482d6c6ecc28925744e09afd0a26a1a139fabeb7789        20      /usr/share/gdb/auto-load/usr/lib32/libglib-2.0.so.0.6200.2-gdb.py
F...    9e3f14dcff0e825d3cc30482d6c6ecc28925744e09afd0a26a1a139fabeb7789        20      /usr/share/gdb/auto-load/usr/lib64/libglib-2.0.so.0.6200.2-gdb.py
```

After discussing with @phmccarty we suspect this is an edge case in rename logic that is triggered during delta pack creation.  It doesn't make much sense to be generating deltas when the file hashes are the same and a rename just needs performed.

ashleshaAtrey:
While creating delta packs, only possible delta files are found.
If delta file is not found, log the error in debug mode instead
of generating warning, as delta might not have been
created for valid reason.

fixes #628
Signed-off-by: Ashlesha Atrey ashlesha.atrey@intel.com

### DIFF
--- a/swupd/packs.go
+++ b/swupd/packs.go
@@ -294,7 +294,10 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
 
 func copyFromDelta(tw *tar.Writer, delta *Delta) (fallback bool, err error) {
 	f, err := os.Open(delta.Path)
-	if err != nil {
+	if os.IsNotExist(err) {
+		log.Debug(log.Mixer, err.Error())
+		return true, nil
+	} else if err != nil {
 		return true, err
 	}
 	defer func() {
--- a/builder/deltapacks.go
+++ b/builder/deltapacks.go
@@ -84,7 +84,7 @@ func createDeltaPacks(fromMoM *swupd.Manifest, toMoM *swupd.Manifest, printRepor
 
 				if len(info.Warnings) > 0 {
 					for _, w := range info.Warnings {
-						log.Warning(log.Mixer, "    Bundle %s: %s", b.Name, w)
+						log.Warning(log.Mixer, "Bundle %s: %s", b.Name, w)
 					}
 				}
 				report := fmt.Sprintf("  Finished delta pack for bundle %q from %d to %d\n", b.Name, b.FromVersion, b.ToVersion)


swupd/packs.go
```diff
@@ -295,7 +295,8 @@ func WritePack(w io.Writer, fromManifest, toManifest *Manifest, outputDir, chroo
 func copyFromDelta(tw *tar.Writer, delta *Delta) (fallback bool, err error) {
 	f, err := os.Open(delta.Path)
 	if err != nil {
-		return true, err
+		log.Debug(log.Mixer, err.Error())
+		return true, nil
```
reaganlo, line 299: Maybe we should return error as `nil` only if the err is `os.IsNotExist(err)`. For any other error condition it should return the valid error.

PR #773, opened 2021-10-15
builder: Add a bundle's packages as the content to the tracking file

bryteise:
Add the packages found in a bundle to the bundle tracking file. This
will be handy when trying to figure out client side what packages are
installed on the system.

Signed-off-by: William Douglas <william.douglas@intel.com>


bryteise:
@phmccarty if this seems reasonable.

PR #776, opened 2022-08-26
Fix setting of log location with custom config

phmccarty:
When mixer's -c/--config option was used, the referenced config was read
too late to set a default log location. As a result, the log location
was always set to the empty string (disabling logging), that is, unless
--log VALUE was specified, or if mixer's current working directory
contained a builder.conf with LOG defined.

The root cause of this bug is an order-of-operations issue with cobra
integration; fix the issue by setting the log location in
PersistentPreRunE() instead of init(), ensuring that the command line
arguments have been processed, with `configFile` being set to the
-c/--config value if specified.

### DIFF
--- a/mixer/cmd/root.go
+++ b/mixer/cmd/root.go
@@ -49,6 +49,15 @@ var RootCmd = &cobra.Command{
 			}
 		}
 
+		// When the logFile flag is not specified, try to initialize it
+		// from the config when available.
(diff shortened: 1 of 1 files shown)


phmccarty:
Ugh, the commit description needs a slight revision. I'll push that shortly.

bryteise:
Ah that's a tricky one. Thanks!

PR #777, opened 2022-08-30
fix missing files in packs

fenrus75:
we need to not return "nil" if a delta file is missing, because that causes the caller
to not use the fallback option (which would include the full file)...
and we need the fallback option to have a complete pack

Signed-off-by: Arjan van de Ven <arjan@linux.intel.com>


phmccarty (commented):
+1

PR #780, opened 2022-08-31
Skip delta creation for pure rename case

phmccarty:
The change in 73bf833128 resulted in "pure rename" deltas not being
added to packs, but we can go one step further by not queuing these
deltas to be created in the first place, saving some cycles.

### DIFF
--- a/swupd/delta.go
+++ b/swupd/delta.go
@@ -341,6 +341,11 @@ func findDeltas(c *config, oldManifest, newManifest *Manifest) ([]Delta, error) 
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


fenrus75 (approved):

phmccarty:
After some local testing, I see that pure rename deltas are no longer generated (great), but then the fullfiles are added to the packs instead...

I assume that if swupd-client handles the pure rename case okay, then we don't want either the delta or fullfile in the delta pack?

fenrus75:
we don't want either indeed

the same hash check for skipping deltas yesterday was added only to the
error path...
and your flow doesn't get near that

PR #795, opened 2024-01-26
Add apx support

bryteise:


swupd/files.go
```diff
@@ -81,17 +81,38 @@ const (
...
+	SSE_4:    "",
 	AVX2_1:   "/V3",
 	AVX2_3:   "/V3",
+	AVX2_5:   "/V3",
+	AVX2_7:   "/V3",
 	AVX512_2: "/V4",
 	AVX512_3: "/V4",
+	AVX512_6: "/V4",
+	AVX512_7: "/V4",
+	APX_4:    "/V5",
```
fenrus75: should this not be V5 but something like VA in case an actual v5 comes along that's something else

fenrus75 (approved):

fenrus75 (approved):

fenrus75 (approved):

PR #801, opened 2025-01-08
Rename rst2man.py to rst2man

bryteise:
The package has been rst2man for long enough, switch the Makefile to use the "new" name.


bwarden (approved):

bwarden:
Ah yes, this is where I remember we dealt with it before:
https://github.com/clearlinux/swupd-client/commit/abb3f2b4f1d60aaeed0bce2259984b35eeb48820

PR #804, opened 2025-04-30
Improve dir subtraction

bryteise:


swupd/manifest.go
```diff
@@ -755,22 +756,60 @@ func (m *Manifest) addManifestFiles(ui UpdateInfo, c config) error {
...
+		chrootDir := filepath.Join(c.imageBase, fmt.Sprint(ui.version), "full")
 		includes := m.GetRecursiveIncludes()
+		usedDirs := make(map[string]struct{})
+		bundleDirs := []string{}
+		bundleFiles := []string{}
 		for f := range m.BundleInfo.Files {
+			fullPath := filepath.Join(chrootDir, f)
+			if fi, err := os.Lstat(fullPath); err == nil {
+				if !fi.IsDir() {
+					usedDirs[filepath.Dir(f)] = struct{}{}
```
bwarden, line 768: Do we need any special handling for a file `f` in the root dir, e.g. `/foo`? I assume Dir would return `./` in that case, but the only non-dirs in root should be the symlinks `lib`, `lib64`, `bin`, and `sbin`, right?

swupd/manifest.go
```diff
@@ -755,22 +756,60 @@ func (m *Manifest) addManifestFiles(ui UpdateInfo, c config) error {
...
+		slices.Sort(bundleDirs)
+		slices.Reverse(bundleDirs)
+		for _, f := range bundleDirs {
+			if _, ok := usedDirs[f]; ok {
+				usedDirs[filepath.Dir(f)] = struct{}{}
+			}
+		}
+		for _, f := range bundleDirs {
+			isIncluded := false
+			for _, inc := range includes {
```
bwarden, line 805: Do we need to fudge `includes` so `os-core` is always the first element?

swupd/swupd_test.go
```diff
@@ -191,6 +190,82 @@ func TestFullRunDelta(t *testing.T) {
...
+	mustValidateZeroPack(t, ts.path("www/10/Manifest.test-bundle1"), ts.path("www/10/pack-test-bundle1-from-0.tar"))
+	mustValidateZeroPack(t, ts.path("www/10/Manifest.test-bundle2"), ts.path("www/10/pack-test-bundle2-from-0.tar"))
+
+	osCore := ts.parseManifest(10, "os-core")
+	testBundle1 := ts.parseManifest(10, "test-bundle1")
+	testBundle2 := ts.parseManifest(10, "test-bundle2")
+	checkIncludes(t, testBundle1, "os-core", "test-bundle2")
+	checkIncludes(t, testBundle2, "os-core", "test-bundle1")
+	checkFileInManifest(t, osCore, 10, "/a")
+	checkFileInManifest(t, testBundle1, 10, "/a/b")
```
bwarden, line 226: Do we need something like `checkFileNotInManifest(t, testBundle1, 10, "/a")` and for `testBundle2`?

swupd/manifest.go
bryteise: `filepath.Dir("/file")` will return `"/"` which is correct in my mind though won't come up as mixer doesn't put `/` in the manifests so it wouldn't be checked for removal.

swupd/manifest.go
bryteise: In Manifest files I believe os-core is already the first element but mixer shouldn't be impacted by include order. The subtraction should happen regardless if something is in os-core no matter if os-core is looked at first or last from that logic.

swupd/swupd_test.go
bryteise: Yea, seems pretty reasonable to add, though I'll need to create `checkFileNotInManifest`.

bryteise: Looks like `fileNotInManifest` was there and did what I wanted for this.

bwarden (approved):

